Issue #1, opened 2016-12-01
Include tslint in webpack build


artlowel:
Tslint and codelyzer were added to the project, but currently you have to manually run them. 
There's probably a webpack loader to run them every time webpack builds. We should add that

PR #7, opened 2016-12-01
Tslint build: this connects to #1

wwelling:
1. Added tslint-loader
2. Upgraded dependencies.
3. Configured webpack to run tslint on build.

I opted to start with the default recommended tslint.json. This configuration is for latest version of codelyzer. From this we can add other linting options we would prefer.


wwelling:
resolves #1

package.json
```diff
@@ -60,23 +60,24 @@
   },
   "devDependencies": {
     "@ngtools/webpack": "~1.1.7",
```
artlowel: With this new set of packages I get unmet peer dependency warnings if `@ngtools/webpack` v1.1.8 is used, removing the `~` ensures that can't happen.

wwelling:
I will fix the dev dependency warning and resolve merge conflicts as soon as possible.

Issue #3, opened 2016-12-01
i18n: add ng2-translate


artlowel:
- add ng2-translate to the project
- ensure it works on both server and client (that needed a custom provider in the prototype if I'm not mistaken)
- create an english translations file
- move any untranslated labels currently in the UI to this file

PR #9, opened 2016-12-01
I18n translate: this connects to #3

wwelling:
1. Setup ng2-translate
2. Added initial en.json
3. Translate some static content

I included some examples of using data and also manually translating from code.


artlowel:
LGTM, merging

Issue #17, opened 2016-12-12
Test the combination of redux and universal


artlowel:
Before we can commit to using redux (ngrx) we need to test that it works properly in combination with angular universal

PR #18, opened 2016-12-12
Ngrx

artlowel:
This PR connects to #17 

I rewrote the collapsing of the header using ngrx, and everything seems to work fine with ngrx and universal.

### A little bit of an explanation:
- all actions having to do with the header can be found in `header.actions.ts`
  - they are `COLLAPSE`, `EXPAND`, and `TOGGLE`
  - The actions class contains an action type, and a function to create it, for each action. This is my take on https://github.com/erikras/ducks-modular-redux
- `header.reducer.ts` contains an interface to represent the state of the header, an initial state, and the reducer function:
  - The state is simply an object containing the boolean `navReduced`
  - on collapse it's set to true, on expand it's set to false, on toggle it's flipped
  - `Object.assign({}, state, {…})` is used to create a new state object, containing the required change without altering the original state object
- the contents of the collapse, expand and toggle functions in header.component.ts have been replaced with `this.store.dispatch(…)` calls with actions of the corresponding type
- in the `ngOnInit` function of the header component a new observable `isNavbarCollapsed` is initialized.
  - this is done by selecting the correct reducer from the store and mapping it to fit an `Observable<boolean>`.
- in `header.component.html` the `isNavbarCollapsed` observable is used instead of the function that was there before.

When a user clicks the toggle button:
- the header component dispatches the TOGGLE action
- the reducer receives the action, and updates the state
- this triggers a new event for the `isNavbarCollapsed` observable, which updates the UI

### Effects
The navbar also collapsed when the window was resized or when the route chagned. To do these things I added another reducer: the `HostWindowReducer`.  This only has a single action: `RESIZE`. it contains the new width and height of the window as its payload. When angular detects a window resize in `AppComponent` a `RESIZE` action is dispatched.

The response to a `RESIZE` action should be that the header's `COLLAPSE` action is dispatched, after that, the UI will take care of itself. 
The same thing happens in response to an `UPDATE_LOCATION` from the router (provided by the [@ngrx/router-store module](https://github.com/ngrx/router-store)) 
These reactions are specified in header.effects.ts.

Effects (short for _side effects_) listen for actions, can transform their payloads and can dispatch new actions as a result. They can be used in this simple case to respond to the route changing or the window resizing, but they're more commonly used to fetch data from the rest api for example:
- An action is dispatched to initiate a call: e.g. `FIND_ALL_USERS`
- the reducer changes the state accordingly: e.g. sets `isLoading` to true
- the effect picks up on the `FIND_ALL_USERS` action, and does an http call to the REST API
- when the call returns, the effect can use RxJS operators on the result to get the payload: a list of user objects for a `FIND_ALL_SUCCESS` action. 
- and in it's catch block the effect can dispatch a `FIND_ALL_ERROR` action with the error message as its payload
- To either of those actions, the reducer responds by setting `isLoading` to false and either adding the list of users to the state, or adding the error message to the state
- a component subscribed to the list of users will update when it changes, so will a component subscribed to the error message.

### aggregator files
- i've also added the files app.reducers.ts and app.effects.ts to aggregate all reducers and effects and prevent them having to be added to app.module seperately. 
- These aggregator files should be made at the level of each module


artlowel:
A way to verify that everything works properly server-side as well as client side, could be to set `navCollapsed` to false in the initialState, and to dispatch an onResize action in the constructor for app.component, and then opening the app with javascript disabled.

This way the nav bar will start open, but the resize event will fire, causing HeaderEffects to fire a collapse event causing the nav bar to be closed once it reaches the client.

wwelling:
Does not start in production using ```npm start```

```ngc -p tsconfig.aot.json``` fails

wwelling:
Builds and runs in dev mode. However, when the navbar is collapsed the bars icon is not on the right and doubles up above the title.

I will see if I can assist with these issues.

wwelling:
```Error: can't find symbol undefined exported from module```

https://github.com/angular/angular/issues/11402

artlowel:
>when the navbar is collapsed the bars icon is not on the right and doubles up above the title.

I forgot to move the style from AppComponent to HeaderComponent. Fixed now

The AoT compilation errors are fixed as well. Combining the reducers is slightly more complicated when it has to work with ngc.

wellingWilliam:
Awesome. I will test and merge in a few hours.

wwelling:
Works. Merging.

Issue #15, opened 2016-12-08
Create a static mock REST API


artlowel:
To work against before the new REST API is far enough along.

This looks promising:
https://coligo.io/create-mock-rest-api-with-json-server/

PR #19, opened 2016-12-13
Mock rest api

artlowel:
This PR connects to #15 

There was already a mock rest api included in server.ts for the demo data in the seed project I modified it to return basic collection and item objects. 
They likely don't entirely match the json api spec. I'm quite sure the relationship attributes aren't entirely up to scratch. 
But it's a starting point.


wwelling:
Looks good. I assume this will continue to be updated to closer match the new REST API spec. Merging.

Issue #26, opened 2016-12-20
Add postcss and autoprefixer


artlowel:
[postcss](https://github.com/postcss/postcss) allows us to add a number of plugins that perform actions on the css, after it has been converted by sass

The most important such plugin is [autoprefixer](https://github.com/postcss/autoprefixer). It allows you to write plain css without having to worry about browser specific prefixes. You specify in your autoprefixer config how many browser versions you want to support (e.g. `'last 2 versions'`) and autoprefixer does the rest.

[Bootstrap 4 was written with autoprefixer in mind](https://v4-alpha.getbootstrap.com/getting-started/flexbox/), so as we're converting bootstrap's sass to css ourselves we'll need to add autoprefixer as well.

The conversation in [the PR that added bootstrap](https://github.com/DSpace/dspace-angular/pull/10) may be useful

PR #33, opened 2016-12-21
added postcss

artlowel:
This adds postcss with autoprefixer, and switches to nodemon to watch for style changes instead of the built-in node-sass watcher.

This PR connects to #26


wwelling:
@artlowel do you want to resolve the merge conflicts or would you like I handle them? Otherwise this looks straight forward.

Issue #2, opened 2016-12-01
Add and configure test tools


artlowel:
We agreed on using:
- jasmine for unit tests
- protractor for e2e tests
- karma as a test runner

These tools need to be added to the project and configured.

We'll also need some documentation added to the wiki on how to use them (e.g. create a `my.component.spec` file in the same directory as the component for jasmine, put your e2e tests in that folder, that sort of thing…)

sourcedump:
Hello!
I need some hints to decide how to proceed and I'm sure that someone of you has more experience than me with the 'package.json'.
Protractor needs to be installed as global. Should I put it into the 'global' script? Or we should create a 'global-test' or something like that?
Moreover, Protractor has Selenium server as dependency that can be installed with the following shell command:
webdriver-manager update
In this case where I can put that command (must be executed after the Protractor installation)? 
For a more clean install I should check the presence of a Java SDK because it's needed by Selenium. 
Maybe all these operations need a separated shell script?
Thanks

artlowel:
angular-cli uses these exact same tools for testing. If I were you, I'd [set up a new angular-cli project](https://github.com/angular/angular-cli#usage), and take a look at the code it generates: which relevant libraries are added to package.json, what the config files for these tools look like, etc. It should give you a good starting point.

Looking at a local angular-cli project, it would indeed seem that protractor is a global dependency, and should therefore be added to the `global` script. And that they add a `e2e` npm script that runs protractor, and a `pree2e` script (meaning it will trigger automatically before every run of `e2e`) that runs `"webdriver-manager update"`

wwelling:
@artlowel good advice

For a more complete setup for testing I would check out this starter project. It is not using Angular Universal, but that should not play a role on the testing strategy or setup.

https://github.com/AngularClass/angular2-webpack-starter

sourcedump:
Thanks!

PR #34, opened 2016-12-22
Testing

sourcedump:
Hello!
Here the E2E and unit-test integration. All the instructions whitin the README.md.
We have an unresolved issue that we will try to fix ASAP about the 'karma-remap-istanbul' plugin. It's a minor problem you can encounter using the coverage page: the home page is not at the '/' address but inside the folder with the name of the browser you used to run the tests (e.g.: "http://localhost:9875/Chrome%2055.0.2883%20(Linux%200.0.0)/"). You can, however, select the folder with the browser... no need to remember srange urls.


artlowel:
this PR connects to #2

atarix83:
The problem with karma remap it's solved with last commit [11c5e2f](https://github.com/DSpace/dspace-angular/pull/34/commits/11c5e2ff7957b3122d005cf0f01c96d741e5988d)

package.json
```diff
@@ -49,7 +50,14 @@
     "debug:build:prod": "node-nightly --inspect --debug-brk node_modules/webpack/bin/webpack.js  --config webpack.prod.config.ts",
     "docs": "typedoc --options typedoc.json ./src",
     "lint": "tslint \"src/**/*.ts\" || true",
-    "global": "npm install -g angular-cli nodemon npm-check-updates rimraf ts-node typedoc typescript webpack webpack-bundle-size-analyzer rollup marked node-gyp"
+    "global": "npm install -g angular-cli nodemon npm-check-updates rimraf ts-node typedoc typescript webpack webpack-bundle-size-analyzer rollup marked node-gyp protractor",
+    "postglobal": "npm link protractor",
```
artlowel: Why `npm link protractor`? IIRC that should only be necessary if you're building protractor locally instead of downloading it.

artlowel:
I tried modifying the existing tests and adding new ones for both unit and e2e tests, and that all works. I do have a few comments:

[Angular 2's testing guide](https://angular.io/docs/ts/latest/guide/testing.html#!#q-spec-file-location) recommends putting spec files in the same folder as the code they're describing, instead of in a separate `/test` folder. We should consider doing the same.

Would it be possible to have separate `npm run test` and `npm run test:watch` commands? Otherwise people that want to use the watch feature have to override the config file and run the risk of accidentally committing it.

I get the error message `keywords if/then/else require v5 option` when I run a webpack build. This is likely due to [this issue](https://github.com/webpack/webpack/issues/3562) and should be fixed by upgrading to webpack `v2.2.0-rc.2` or higher.

package.json
sourcedump: The link is needed because some libraries use Protractor as a local dependency and they need to find its files.

sourcedump:
Ok, we will fix those things ASAP!

package.json
artlowel: In that case I think the solution should be to add protractor as a local, as well as a global dependency, just like  currently e.g. [webpack is installed as a global dependency](https://github.com/DSpace/dspace-angular/blob/6f3229fb3858847a3ceb6eda38d4d206b8b3a92a/package.json#L52), to be able to use it from the command-line, but [also added as a dependency of the project](https://github.com/DSpace/dspace-angular/blob/6f3229fb3858847a3ceb6eda38d4d206b8b3a92a/package.json#L133) so it can be used by other libs.

[`npm link package-name` is designed to be used during development of an npm package](https://60devs.com/simple-way-to-manage-local-node-module-using-npm-link.html), to be able to use it as a library in a second project it before it's hosted on npm

sourcedump: I surely can do that, but why we should install Protractor two times when a link can resolve all the issues?

artlowel: Because as it is we have no information about the version of protractor that this project is dependent on. 

If later, for a different project you need to install a different version of protractor, globally, then things in this project can start breaking, and since you haven't changed anything to this project, that'll be hard to debug. 

If this project has a local dependency on protractor, we'll know that that version will work, and we always use its executable instead of the global one, by using its path in the `node_modules` folder.

To prevent those kinds of issues we should use as few as possible global dependencies. So if you want to only have a single install of protractor, I'd leave out the global dependency, add it as a local dependency and use `node_modules/protractor/bin/protractor` as the executable in the npm tasks instead

sourcedump: Done!

artlowel:
Thanks.
LGTM

atarix83:
with the last commit [144324e](https://github.com/DSpace/dspace-angular/pull/34/commits/144324e0233dc7ae18bdb4d0195702a396ccb340) : 

- the tests direcory has been deleted and spec files moved in the same folder as the code;
- test:watch added as a script;
- to remove error message `keywords if/then/else require v5 option` webpack has been upgrade to `v2.2.0-rc.3`. The upgrade produce this warnings on building :
`WARNING in asset size limit: The following asset(s) exceed the recommended size limit (250 kB).` 
`This can impact web performance.`
`Assets:` 
  `spec-bundle.js (8.77 MB)`
`WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the` `recommended limit (250 kB). This can impact web performance.`
`Entrypoints:`
  `spec-bundle.js (8.77 MB)`
      `spec-bundle.js`
Now warnings are disabled by adding a change in webpack.config, however to solve the problem it'd be better to follow recommendations [here](https://webpack.js.org/guides/code-splitting/); 
- Now protractor is installed as local dependency.

artlowel:
I just noticed that the upgrade to webpack 2.2.0-rc.3 introduced two new peer dependency warnings:
```
npm WARN string-replace-loader@1.0.5 requires a peer of webpack@1 || 2 || ^2.0.0-beta || ^2.1.0-beta but none was installed.
npm WARN karma-webpack@1.8.0 requires a peer of webpack@^1.1.0 || ^2 || ^2.1.0-beta but none was installed.
```

I checked, and those modules don't have webpack 2.2.0-rc.3 compatible versions yet. But I did look further into that `keywords if/then/else require v5 option` [issue I linked](https://github.com/webpack/webpack/issues/3562). It turns out it's not an issue in webpack itself, but of one of its dependencies: `ajv-keywords`. Both problems can be resolved by rolling back to webpack 2.1.0-beta.27 (and the compatible webpack-dev-server and configuration files) and adding `"ajv-keywords": "1.1.1"` to the devDependencies in package.json

artlowel:
That fixed things, thanks.

wwelling:
Works well. LGTM

PR #39, opened 2017-01-18
Reducer tests

artlowel:
This PR adds tests to the ngrx objects I added in #18.

I also changed the way actions are specified. They used to be a combination of a string to represent the type and a method to create a valid action from a given payload. But in typescript it makes more sense to turn them into classes. That way we can use type checking to determine valid actions for a reducer.

The deep-freeze library that was added as a dev dependency serves to freeze the ngrx state during tests, so we can verify that the a reducer doesn't modify the previous state object when generating the next state. See [the headerReducer spec](https://github.com/artlowel/dspace-angular/blob/reducer-tests/src/app/header/header.reducer.spec.ts#L43) for an example.


wwelling:
LGTM. Unit and e2e tests passing. Production and development running as expected. Watching working as well. Merging.

Issue #27, opened 2016-12-20
Environment variables


artlowel:
Add a mechanism to configure environment variables.

[This is the equivalent in the prototype](https://github.com/DSpace-Labs/angular2-ui-prototype/blob/master/config.ts) however it has the downside of having the values on git. It would be better to have a mechanism where the keys can be on git, but the values remain in a file local to the environment.

atarix83:
I think to create this model for environment variables :

1.  `/src/config.ts` 
    This is the file to import into Components;
2. `/config/environment.dev.json and /config/environment.prod.json`
    This is where we put our env settings for the different environments. We need to have a json file for every enviroment (e.g. `dev`, `prod`);
3. `/config/environment.common.json`
 This is where we put our env settings shared between the environments;
4. All files in config dir are not committed on git but are imported on building into config.ts depending on running enviroment.

Would it be a good scenario?

artlowel:
Sounds good.

Maybe we could also provide a `environment.default.json` file on git that isn't used but serves as a template for the dev and prod files.

PR #42, opened 2017-01-20
Environments

atarix83:
This PR connects to [27](https://github.com/DSpace/dspace-angular/issues/27)

I have created :

- `/src/config.ts`
This is the file to import, It loads environmental and shareable settings and makes them available to app components
- `/config/environment.common.default.json` 
Template file for genereal configuration, where we put our env settings shared between the environments;
- `/config/environment.default.json`
Template file for environment configuration, where we put our env settings for the different environments. We need to have a json file for every enviroment (e.g. dev, prod);

I have changed `.gitignore` file to prevent commit of the configuration file created on git.

I used this [config.ts](https://github.com/DSpace-Labs/angular2-ui-prototype/blob/master/config.ts) as model for environment configuration file.


config/environment.default.json
```diff
@@ -0,0 +1,26 @@
+{
+  "production": false,
+  // Location of REST API
```
artlowel: Unfortunately, comments are not allowed in a json file. They're not part of [the spec](http://www.json.org/)

artlowel:
It works, but I have a few comments:

The proxy section in environment.default.json can be removed. It was necessary for the prototype, not for this project.

Would it be possible to use the default config file by default, and overlay it with the local versions? I realize I was the one that suggested to use the default files as templates only, but seeing it in action the obvious downside is that simply checking out and building the project will result in a build error, because there is no config file. 

It seems to me it would be better to have it the way it works for dspace 6. You have a default file and you can choose to override certain parts of it by putting them in a local file. You can combine two objects in javascript using [Object.assign](https://github.com/DSpace/dspace-angular/blob/master/src/app/header/header.reducer.ts#L16). So something like `Object.assign({}, defaultConfig, localConfig)` would result in a new configuration object that uses all the default config's values, overlayed with the ones that exist in the local config.

To be able to do so, we'd also need proper default values, so `http://localhost:3000` as `baseUrl` for both rest and ui, `/api` as `nameSpace` for rest, `/` as `nameSpace` for ui

config/environment.default.json
atarix83: Yes, I write a note in documentation to specify this, otherwise could we save template as JS file?

artlowel: I'd use a TS file, but yeah, that's what we did in the prototype

atarix83:
I've removed all template files. Now there is an `environment.default.js` with default parameters that could be override by local environmental files.
I have moved JSON file into JS file so we can add comments.

artlowel:
LGTM

sourcedump:
Ok for me also!

Issue #24, opened 2016-12-20
Create a spinner component & service


artlowel:
In order to show when a page is loading

I'm thinking of something like this:
- Create a `router-outlet-with-spinner` component, that replaces the app.component's router-outlet
- Also create a SpinnerService, that stores whether or not the spinner should be active in the ngrx store, has a few methods to change the active state, and to observe it.
- in the `router-outlet-with-spinner` component,  if the service says the spinner should be active, show a spinner component, otherwise show the router-outlet
- Each component that corresponds to a route can set the SpinnerService to active in its constructor, and subscribe to (the combination of) its observables-that-take-time-to-resolve to disable it again.
- Use either a pure css, or svg spinner. Use (a calculation with) bootstrap variables for its color and dimensions. e.g.`$spinner-height: 4*$line-height-base;`

artlowel:
closing in favour of #73

PR #43, opened 2017-01-25
Spinner component & service

LotteHofstede:
This PR connects to #24 

I created a Spinner component that can be displayed when content is loading.
The SpinnerService manages the status of the spinner, which can be activated and deactivated.

A SpinnerWrapper component was created to combine and switch between the SpinnerComponent and whichever content you wish to render. 

For now, the SpinnerWrapper is used in app.component.html and uses the "router-outlet" as content to show when the Spinner is deactivated.
To illustrate the behaviour of the spinner, I added a small timeout delay in the HomeComponent to switch the spinner on and off. 
For real-life use cases, you would activate the spinner before you start loading content and deactivate it when all content is available and ready to be rendered.

The Spinner was created using http://loading.io/ and only uses HTML/CSS. The Spinner template consists of a number of empty divs that are necessary to render the Spinner.

This PR is dependent on PR #39


sourcedump:
Hello!
LGTM but I think we should have something like this example:
[https://codepen.io/MattIn4D/pen/LiKFC](https://codepen.io/MattIn4D/pen/LiKFC)
where we have a page centered spinner and a partial transparent overlay.
Let me know if you agree.

artlowel:
I'm not a big fan of an overlay, and properly centering the spinner vertically may take more effort than we'd like to put in to style at this stage, but I would like to see it centered horizontally, and made smaller, say, half its current size?

artlowel:
Closing, in favour of #73

Issue #30, opened 2016-12-20
Write a JSONAPISerializer


artlowel:
- Write a JSONAPISerializer  based on [yayson](https://github.com/confetti/yayson).
- use generics to specify the type of model it's working with (Collections, Items, …)
- add a deserialize method: this will get the response body from the REST API as input and return a single model of the generic type
	- Don't use yayson's store as a cache, it doesn't have all the features we need and it doesn't use ngrx. Instead create a new store for every new deserialize operation.
- add a deserizalizeArray method: that will do the same thing, but for an array (e.g. a `GET` call to `/api/collections/`)
- add a serialize method: this will turn a single model of the generic type in to a JSON API object to be sent to the REST API.
- add a serializeArray method: this will turn an array of models of the generic type in to a JSON API object to be sent to the REST API.
- instead of making separate methods for a single object and an array, you could also consider using [TypeScript's union types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types) to work with either a single object or an array in the same method.
- To start with every endpoint can use the same serializer, we can extend it (and make e.g. an ItemSerializer) if exceptions need to be made for an endpoint
- The docs for yayson aren't great. It may help to take a look at the [tests on DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/yayson/yayson-tests.ts)
- I also have a [test-branch](https://github.com/artlowel/dspace-angular/tree/yayson-test), that [uses yayson to parse a list of collections](https://github.com/artlowel/dspace-angular/blob/yayson-test/src/app/shared/data/collection-data.effects.ts#L28).

PR #45, opened 2017-01-25
JSON API serializer

artlowel:
This PR connects to #30 

Adds a serializer interface with an implementation for JSON API based on [yayson](https://github.com/confetti/yayson). It will likely have to be adjusted as we implement other things, and as the new rest api starts to take shape, but it's a start.

Also adds a JSON API validator based on [json-schema](https://github.com/tdegrunt/jsonschema). [The schema itself](http://jsonapi.org/schema) was created by the people who wrote the spec


artlowel:
As it seems likely we'll be working with HAL instead of JSON API, I'm closing this PR

Issue #49, opened 2017-01-27
Run all tests at once


artlowel:
It would be useful to be able to run all tests at once using a single command. For example to use in Travis CI. However it should still be possible to run them separately during development.

I would also create a pree2e script that triggers webdriver:update to ensure protractor doesn't fail because webdriver:update hasn't run.

sourcedump:
Hello!
the 'webdriver:update' it's a one shot command and you don't need to run it each time.
Moreover if you have a remote Selenium server you don't need to run it at all.
Are you sure you want to use it in a 'pree2e'?
When you say 'all tests' you mean unit test and e2e together?
Thanks

artlowel:
Yes, when I say 'all tests' I mean both unit tests and e2e tests.

`webdriver:update` downloads its dependencies to the node_modules folder. If you check out a branch with a different package.json, it's best to remove that folder (run `npm run clean`) and run `npm install` again to ensure you're working with the correct set of packages for that branch. Afterwards, if you want to run the e2e tests, you have to run `webdriver:update` again. As you can imagine, when reviewing PR's this happens all the time.

Besides, if you run `webdriver:update` when it's not needed, it doesn't do anything and just exists. Running it every time doesn't do any harm.

PR #51, opened 2017-01-31
'pree2e' and 'ci' scripts added

sourcedump:
This PR connects to [#49](https://github.com/DSpace/dspace-angular/issues/49).

'pree2e' and 'ci' scripts added. 
'pree2e', runned each time 'e2e' is called, keeps protractor drivers up to date while 'ci' runs tslint and all the tests (both Unit tests and E2E tests).


artlowel:
LGTM

sourcedump:
Actually you need to run a 'start:dev' manually before executing 'ci' (for e2e testing).
Do you think it's possible to run 'start:dev' and, when the server is up, auto-run 'ci' and, again, at the end, shutdown the server started with 'start:dev'?

sourcedump:
Ok, the 'ci' script now run the 'server' before the 'e2e'.

artlowel:
It works, but are you using [npm-run-all](https://github.com/mysticatea/npm-run-all) in order to run two npm scripts in parallel? If so, note that we already have a dependency on [concurrently](https://github.com/kimmobrunfeldt/concurrently) that does the same thing.

sourcedump:
Hello Art, I've tryed to use 'concurrently' but I need to use the '-k' option to kill the server at the end of the tests and with 'npm-run-all' all is fine, instead with 'concurrently' I have a false error at the end of the run. All the scripts run correctly but at the end an error is reported. I'll try to figure out why.

artlowel:
If you can't get it to work, I'm also fine with swapping out concurrently for npm-run-all everywhere (it's only used for two scripts), as long as npm-run-all also works on all platforms. It's just that I'd rather not add two dependencies to do the same job.

sourcedump:
I've relplaced 'concurrently' with 'npm-run-all'. Now all is fine.

artlowel:
LGTM. Merging

Issue #47, opened 2017-01-26
Footer component


artlowel:
Create a component for the footer. Take a look at the prototype for [an example](https://github.com/DSpace-Labs/angular2-ui-prototype/blob/master/src/app/app.component.ts). Note that for the prototype it was part of the AppComponent and never refactored in to its own component.

Use the same content as the prototype

Automatically set the end-year in the copyright message to the current year. That way people don't need to remember to update that i18n message every year. Angular's built-in [DatePipe](https://angular.io/docs/ts/latest/api/common/index/DatePipe-pipe.html) could be useful for this

Make it stick to the bottom [using flexbox.](https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/)

As it is a single-use component that is only referenced in AppComponent's template, it belongs in the [CoreModule](https://angular.io/styleguide#!#04-11). Create that module if it doesn't exist yet.

atarix83:
It's possible to fix footer to the bottom with applying a simple CSS styles. Whether to use flexbox is not preferred I think we should avoid the use of it.

Let me know what do you think.

artlowel:
I don't think we should avoid flexbox any longer.
[caniuse.com](http://caniuse.com/#search=flex) says `display:flex` is supported down to IE10 which should be more than enough backwards-compatibility

But if you have a CSS2 solution that works as well the flexbox solution and isn't based on a hack (such as adding 10000px of padding-bottom and -10000px of margin-bottom) that's fine as well.

wwelling:
I suggest using ```display: table```

such as

```
body,
html {
  height: 100%;
  margin: 0;
  padding: 0;
}

app {
  display: table;
  width: 100%;
  height: 100%;
  overflow-x: hidden;
}

main {
  display: table-row;
  table-layout: fixed;
  height: 100%;
  padding-top: 15px;
}

footer {
  display: table-row;
  width: 100%;
  text-align: center;
}
```

wwelling:
IE10 is only partial support of flexbox. As it turns out this partial support does not support display flex for sticky footer.

atarix83:
well, I've applied [this solution](https://codepen.io/pirrera/pen/BzLbC) and It seems to work properly also on microsoft edge emulating IE10

artlowel:
IIRC we've used a flexbox based sticky footer that worked on IE10 in a project recently. I'll see if I can find the details.

I'd also rather avoid `display: table`. It affects how other css rules inside of it work.
[...]

PR #60, opened 2017-02-13
Footer

atarix83:
This PR connects to [47](https://github.com/DSpace/dspace-angular/issues/47)

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -1,13 +1,21 @@
 {
-  "title": "DSpace",
-  "nav": {
-    "home": "Home"
-  },
   "example": {
     "with": {
       "data": "{{greeting}}, {{recipient}}!"
     }
   },
+
+  "footer": {
+    "copyright": "copyright © 2002-{{ year }}",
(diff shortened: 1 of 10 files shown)


artlowel:
Since the wrappers and the main components are part of app.component, I'd move the css you added to `main.scss` to `app.component.scss`

artlowel:
Also, instead of prefixing the footer css with `/* just for demo */` I'd try to take it a step further in the right direction, by using variables as much as possible.
e.g. https://gist.github.com/artlowel/76bc90d2722fbe4ee2e2576bf718ee54
That way it'll be easier to theme later.

artlowel:
LGTM

sourcedump:
Good also for me!

wwelling:
```
+.inner-wrapper {
 +  flex: 1 1 auto;
 +  flex-flow: column nowrap;
 +  display: flex;
 +  min-height: 100vh;
 +  flex-direction: column;
 +}
```
I don't think there should be a fixed min-height.

Issue #32, opened 2016-12-20
Add caching


artlowel:
This task depends on #31 

- Each Data service created in #31  should have a cache. 
- users of the data services (meaning components, other services, …) shouldn't need to know about the cache. They should just go on working the same way (with the same observables) as if the data came from the REST API.
- This cache should store its data in the ngrx store
- It should be implemented in such a way that most code can be shared between data services for different endpoints
- But it should be possible to determine separate time-to-live for each endpoint. For some endpoints it may be necessary to disable the cache altogether
- It should also be possible to combine partial objects:
	- e.g. I first get only a collection's id, name and handle from the REST API, for a certain page
	- next I go to another page and I get that same collection's id and shortDescription from the REST API, but not it's name or handle
	- the store should now contain the collection's name, handle and shortDescription

artlowel:
Initial cache is done.

This ticket was left open because the code still lacks tests and documentation. This was moved to a separate issue: https://github.com/DSpace/dspace-angular/issues/130

Issue #31, opened 2016-12-20
Implement services to interact with the REST API


artlowel:
This task depends on #30 

- For the Collections, Items and Bitstreams endpoints, create a service, a reducer an actions and an effects file
- each service should have
	- a findById method
	- a findAll method, that supports pagination and sorting
	- a findAllInScope method, and supports pagination and sorting
- when a find method is called, dispatch the appropriate FIND_REQUEST action (e.g. `dspace/shared/data/item/FIND_BY_ID_REQUEST`)
- in the corresponding effects file, listen for that FIND_REQUEST action
	- when it arrives, call the appropriate endpoint using ApiService
	- If the call succeeds, use a JSONAPISerializer to deserialize the result, and dispatch the appropriate FIND_SUCCESS action that adds the object(s) to the store
	- If the response is an error, or times out, dispatch the appropriate FIND_ERROR action with an error message.
- The find methods should return observables that select the proper values from the store
- You can find a simplified example of this way of working in [this test branch](https://github.com/DSpace/dspace-angular/compare/master...artlowel:yayson-test?diff=split&name=yayson-test)
- Don't worry about caching at the moment, that's part of a separate task.

artlowel:
The bulk of the code for this issue has been merged with #62
To do: more tests and docs

artlowel:
This ticket was left open because the code still lacks tests and documentation. This was moved to a separate issue: #130

Issue #29, opened 2016-12-20
Create models for Collections, Items and Bitstreams


artlowel:
Also create an interface or class to represent a DSpaceObject. All models should implement it or inherit from it.

Keep in mind that the Collection model should represent both Communities and Collections in the DSpace backend

You can take a look at [the REST API Contract on the wiki](https://wiki.duraspace.org/display/DSPACE/REST+API+Contract#RESTAPIContract-Browsemilestone-Featurerequests), but at the time of writing it far enough advanced yet to be more than a rough guideline

PR #62, opened 2017-02-16
Rest services

artlowel:
This PR connects to #31 and connects to #29

It adds services to interact with the REST API:  CollectionDataService and ItemDataService

These offer a `findById` method, and a `findAll` method with an optional `scope` param, for Items and Collections.

Tests and documentation are still lacking. I'd also like to refactor the data services, and their corresponding ngrx objects, because right now there is too much duplication of code. 

I opened this PR already, as these services are needed to start making UI components such as collection homepages and item view pages.
I suggest that people review this PR in its current state, and that we merge it ASAP so other work can start.
Issue #31 will stay open until docs, tests and refactoring are complete.

I also added an initial version of a CacheService, for the purpose of storing every model in the ngrx store only once. Other places in the ngrx store should only store the ids of the models they need.
That way there is a single source of truth for any one model. It also makes the store less deeply nested, and therefore easier to write a reducer without accidentally mutating part of the state.

The Services can then use RXJS to combine the ids from their section of the store, with the actual objects from the cache to return an observable for the correct set of objects to the user of the service.

You can find an example [here](https://github.com/artlowel/dspace-angular/blob/rest-services/src/app/core/data-services/collection/collection-data.service.ts#L21).

First, get an observable of the ids for the list of collections from the store:

```typescript
this.store.select<Array<string>>('core', 'collectionData', 'findMultiple', 'collectionUUIDs')
```

Then use those ids to fetch an observable from the CacheService with the list of collection objects:

``` typescript
this.cache.getList<Collection>(collectionUUIDs);
```

`flatmap` ensures we return an `Observable<Collection[]>` instead of an `Observable<Observable<Collection[]>>` - it _flattens_ the observables.

There is a separate issue to expand the CacheService to do everything we need: #32


antoine-atmire:
I've tested the item service. I confirm the rest calls are working when calling `findAll` and `findById`.

I confirm the state of the calls are tracked in the store under `core/itemData/findX`. And when it completes succesfully the items are added to the store under `core/cache`.

The state of a succesfull call also contains references to the items that were fetched by this call. The references of those calls are currently used as the source for the observable. This means wrong results will be shown in the UI when mutliples calls override each other.


E.g.
```
this.item1 = this.itemService.findById("8871");
this.item2 = this.itemService.findById("9978");
```
These two calls will be made right after each other, even with a `take(1)`, there's no chance the UI will be able to keep them apart.


I don't think it's necessary to keep those references from the rest calls at all. Instead the service should compose an observable from the store and filtered them according to the find method and parameters that were used.
I don't really see why the state of each rest call is separated by its service function. I would think one list containing all rest calls would be good enough. Maybe adding a random unique ID to keep track of specific rest calls when necessary. But there could be better ways of course.

I still wonder how a component could report on a failed rest call... Its observable will just sit there empty without anything notification.

antoine-atmire:
I've given it a little more thought.
In elm, a generally accepted approach to rest calls is to use the [RemoteData](http://blog.jenkster.com/2016/06/how-elm-slays-a-ui-antipattern.html) type.
This type has 4 possible states: 

- `NotAsked` - We haven't asked for the data yet.
- `Loading` - We've asked, but haven't got an answer yet.
- `Failure e` - We asked, but something went wrong. Here's the error.
- `Success a` - Everything worked, and here's the data.

Now in elm you can't get to `e` or `a` without handling all the cases. This has the merit of forcing the developer to think about what the UI should show in each state. I don't think we can force such a thing in typescript, but we can create a similar type to pass along the information in the same way.

I would do this in angular by giving the `find`-methods the return type `Observable<RemoteData>`.
The `RemoteData` type needs to be filterable by state and be able to give the success/error payloads when appropriate.
The error payload would be a constant object with a static type (response status, error message, etc), and the success payload would be an Observable of the type that we need.

In fact this is very close to what we have. We already keep track of the status of a rest call in the store. It's just a matter of making an observable out of it and mapping the properties into a `RemoteData` type.

 (The points I made in my previous comment about the structure of the store and the source of the success-observable are still valid though)

artlowel:
>The state of a succesfull call also contains references to the items that were fetched by this call. The references of those calls are currently used as the source for the observable. This means wrong results will be shown in the UI when mutliples calls override each other.

I'm aware of this problem. It's because I'm mapping the REST endpoints with redux state, where I should actually be mapping the UI components. Meaning instead of piping every findById request to the same reducer, I should have a separate reducer for every place a findById is necessary in the UI. However in order to do so, we need a way to create all these reducers and their actions without duplicating too much code. 
I'm working on that at the moment by having a generic DataReducer, DataState, and DataActions, but not everything works quite as I want it yet.

>I don't think it's necessary to keep those references from the rest calls at all. Instead the service should compose an observable from the object store and filter them according to the find method and parameters that were used.

I disagree, the logic that determines which set of objects is returned for a certain request should stay on the server. I don't believe recreating it on the client is a good idea. Not only are you doing the same work twice, but there are bound to be cases where either the server and the client return slightly different results for a query, or where you don't have all the information on the client to execute the query properly. So in my opinion we _should_  store the results of every request in the state.

>I don't really see why the state of each rest call is separated by its service function. I would think one list containing all rest calls would be good enough. Maybe adding a random unique ID to keep track of specific rest calls when necessary. But there could be better ways of course.

Because of the problem you mentioned above: if you share the state for different calls, they interfere with one another. A random unique id could work, but it would have to be added to every part of the state (`isLoading`, `errorMsg` too because you'd want to know _which_ request is still loading or had an error ). That basically boils down to having a different state object of the same type for every call which is what I'm trying to achieve now with the generic `Data-` ngrx classes.

>I still wonder how a component could report on a failed rest call... Its observable will just sit there empty without anything notification.

If you're observing a list of Collections, it will return an empty list of Collections until the request is successful. If you want to show something else while it is loading or when there's an error, you can observe the `isLoading` and `errorMsg` values in the state. Granted, the DataServices don't forward those observables as of yet and there are different ways we could implement it. 

I had a look at the elm blog post you linked, and it seems to provide a solid way of implementing that in elm, so I'll play around with it, and see if I can get it to work in TypeScript too.

antoine-atmire:
Oh good, what I meant was also `having a different state object of the same type for every call`. I just thought it would be sufficient to put those in a list instead of a tree, just because it seems simpler.

You make a fair point to keep using the results from the response objects. I can get behind that. But I'm not entirely convinced yet. 
It is a trade off between relying on more state versus relying on duplicated 'server' logic.

So the important questions are:
- Does it make the app more predictable?
- Does it make the app easier to debug when something does not go as expected?
- Does it make it easier to build upon?

I'm leaving the questions open because I would need to reflect on this for longer first, and I still need more experience with angular... Maybe you already know where you stand. 

In any case these are general questions to keep in mind with any development. It's not a bad thing to bring them up once in a while, especially in the early stages.

artlowel:
- The existing cache has been renamed to object cache
- And now there's a second request-cache, that handles all requests to the REST API and their responses.
- If you call a find method on a DataService, the parameters are used to generate a hash, its state is put in the ngrx store using that hash. That way:
	- if the exact same call happens again within a certain amount of time, we don't need to query the server but can use the cached response
	- different calls using the same method (e.g. findById with different IDs) now all have their own space in the ngrx store so they can't conflict or cancel each other, but without having to create separate reducers/actions, etc for every situation in which you need them.
- Responses to the data services' find methods are composed by using the IDs from the request cache to fetch the models from the object cache.
- The state of the ngrx store on the server is now used to initialize the client's store. That way actions that were already performed on the server aren't repeated on the client.

src/app/core/data-services/data.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  findById(id: string): Observable<T> {
+    const key = new ParamHash(this.serviceName, 'findById', id).toString();
+    return this.requestCache.findById(key, this.serviceName, id)
+      .map(entry => entry.resourceUUIDs)
+      .flatMap((resourceUUIDs: Array<string>) => {
+        if(isNotEmpty(resourceUUIDs)) {
+          return this.objectCache.get<T>(resourceUUIDs[0], this.modelType);
+        }
+        else {
+          return Observable.of(undefined);
```
antoine-atmire: I would prefer an empty observable (`Observable.of([])`) over an observable with a single "undefined" item

src/app/core/cache/request-cache.reducer.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    default: {
+      return state;
+    }
+  }
+};
+
+function findAllRequest(state: RequestCacheState, action: RequestCacheFindAllAction): RequestCacheState {
+  console.log('break here', state);
```
antoine-atmire: To be removed

src/app/core/data-services/data.service.ts
artlowel: I'm not sure I agree. This is what `findById` is going to return, and that should be a single object. Either it exists and you return it, or it doesn't, and then returning `undefined` seems like a better idea than returning an empty array. You could argue for not returning anything at all (meaning that the observable doesn't fire a new event when it happens), but that would mean that if you have a UI element that's observing this object, and you clear it (e.g. because the user logged out, and no longer has access rights) the UI wouldn't receive an update, and keep on showing the object

src/app/core/cache/request-cache.reducer.ts
artlowel: will do

src/app/core/data-services/data.service.ts
antoine-atmire: Right, I understand the structure better now. There's indeed a need to fire an event when no data-object is found (anymore)

src/app/core/cache/object-cache.reducer.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    default: {
+      return state;
+    }
+  }
+};
+
+function addToObjectCache(state: ObjectCacheState, action: AddToObjectCacheAction): ObjectCacheState {
+  return Object.assign({}, state, {
+    [action.payload.objectToCache.uuid]: {
```
antoine-atmire: I get this error when asking for an item id that can't be found: `TypeError: Cannot read property 'uuid' of undefined`
Note: I'm still testing using the HomeComponent for now. So I had to comment out the "doRehydrate" in "browser.module.ts" first, because the state from the home component in preloaded by the "universal cache". The errors don't show up on the server side.

artlowel: I had the same issue. The reason is that our mock rest api can't send an error response when it can't find an item. It simply returns a 200 with an empty response. 
But the client should be able to handle unexpected responses as well, so[ I added a check to the data.effects class](https://github.com/DSpace/dspace-angular/pull/62/files#diff-fd2c2e91b0211f4c8e5dd7c329985df5R38) that will throw an error when the returned object is undefined. Eventually this check will be replaced by the [validator](https://github.com/DSpace/dspace-angular/pull/62/files#diff-98b9a305808321337e664dd6c2703c29).

antoine-atmire:
LGTM

Issue #13, opened 2016-12-05
Add typedoc config and comment initial code


wwelling:
We need to get typedocs setup and make sure we keep up on comments. 

Here is a basic typedoc.json
```
{
  "mode": "modules",
  "out": "doc",
  "theme": "default",
  "ignoreCompilerErrors": "true",
  "experimentalDecorators": "true",
  "emitDecoratorMetadata": "true",
  "target": "ES5",
  "moduleResolution": "node",
  "preserveConstEnums": "true",
  "stripInternal": "true",
  "suppressExcessPropertyErrors": "true",
  "suppressImplicitAnyIndexErrors": "true",
  "module": "commonjs"
}
```

We should probably have a npm script to generate typedocs.

wwelling:
http://typedoc.org/guides/doccomments/

artlowel:
Look in to using github pages

PR #71, opened 2017-03-17
Typedoc configuration added

sourcedump:
This PR connects to [#13](https://github.com/DSpace/dspace-angular/issues/13)

Hello! I've configured Typedoc with some exclusions. Let me know if it's ok for you.
FYI: after you run the command to build the docs there is an error that can be ignored (apparently). There is an open issue here with no reply:
https://github.com/TypeStrong/typedoc/issues/438


artlowel:
Looks good. I'd just expand on the part in the readme a bit: add a link to typedoc.org, explain briefly what it does, and maybe also add a link to http://typedoc.org/guides/doccomments/ so people know _how_ to format their comments so they'll work with the generator.

sourcedump:
Ok, done!

PR #78, opened 2017-03-24
Travis CI: this connects to #76

wwelling:


artlowel:
Travis config looks good.

But we agreed not to install yarn as a global npm dependency because [the yarn team recommends that you don't](https://yarnpkg.com/lang/en/docs/install/#alternatives-tab), and if you already have yarn installed using their installer, the used version depends on the order global npm pacakges and yarn are in your $PATH.

It has been added as a prerequisite instead.

wwelling:
I will have yarn installed manually for Travis CI and remove it from the global script.

wwelling:
I am not in favor of yarn. The yarn cli requires write access to global install directory. Even when starting the app with yarn it requires access to the global directory. This must be why they recommend not installing from npm. None of the other cli tools we use require write access to the global directory. On mac I either have to run 'sudo yarn' or give the current user permissions to the global directory.

wwelling:
Correction: The global write access is only when yarn is installed with npm in the global directory. Sorry for the confusion. @artlowel thanks for the clarification.

Issue #79, opened 2017-03-24
Configuration isn't integrated in a useful way


wwelling:
The configuration UI URL is not used within the server.ts and server.aot.ts.  The node express engine binds to the port and address. It should get the values from a configuration file.

The production and development builds and tests are not dictating what environment configuration to use.

Cache time to live isn't setting anything in the server or client side cache. server.ts and server.aot.ts sets the Cache-Control per request but does not reference a configuration.

Whether or not to preboot is a property of the ngApp response. I do not see any reason to ever prevent the browser from rehydrating the universal cache set during pre-render on the server side node. For this to even make since it would have to condition on the configuration everywhere in the code where an asynchronous response is put into the universal cache.

Also, the process to get the environment configuration seems a bit more complex then necessary.

artlowel:
>Cache time to live isn't setting anything in the server or client side cache

The cache time to live is currently used as [the default time for own cache](https://github.com/DSpace/dspace-angular/blob/34d9aece7e27631effd28d9f590a0c09c48da0e1/src/app/core/data-services/data.effects.ts#L41) in the ngrx store, where it is used to determine whether or not a cached request or object is [still valid](https://github.com/DSpace/dspace-angular/blob/34d9aece7e27631effd28d9f590a0c09c48da0e1/src/app/core/cache/object-cache.service.ts#L121). 

>everywhere in the code where an asynchronous response is put into the universal cache.

Nothing but the state of the ngrx store is ever put in to the universal cache. At the moment, I don't really see a reason to. As we're working with redux, we should keep everything in the ngrx store. So async responses are cached in the store, and the state of the store is put in to the universal cache.

> I do not see any reason to ever prevent the browser from rehydrating the universal cache

The option to disabling the rehydrate step was added to make it easier to debug the actions that happen on the ngrx store. If rehydration is enabled all you see in the [redux-devtools](https://github.com/gaearon/redux-devtools) is a single rehydrate action. If you want to see how the server got to that state, you can disable rehydration and then all steps from the initial  state to the present state will be performed on the client, so you can debug them, rollback, change action payloads, etc.

wwelling:
Ok, that configuration may be useful. But still the universal cache should not be populated by the node, server side, if it is not going to be rehydrated by the client. Why not just have the configuration disable pre-render altogether?

artlowel:
I agree, that would be better.

Issue #64, opened 2017-02-16
Use OpaqueToken for injecting global configurations


wwelling:
https://angular.io/docs/ts/latest/api/core/index/OpaqueToken-class.html

The global config should be provided and injected where needed.

example

```
import { OpaqueToken } from '@angular/core';

export interface GlobalConfig {
  API: string
}

export const DEFAULT_GLOBAL_CONFIG: GlobalConfig = {
  API: 'http://localhost/dspace-api'
};

export let GLOBAL_CONFIG = new OpaqueToken('config');
```

Then provide it in the browser and node module

```
 providers: [
    { provide: GLOBAL_CONFIG, useValue: DEFAULT_GLOBAL_CONFIG },
    ...
```

There will have to be adjustments to allow for configurable environment based config. Such as development, production, and test. The configurations could be in separate files and the build can decide which to provide.

PR #82, opened 2017-03-25
Opaque config: this connects to #64 and this connects to #79

wwelling:
Organized src directory. Moved platform specific source files into platform directory.

Provided global configuration as OpaqueToken that can be injected.

Build sets production config value.

Server uses configuration for address, port, and universal configurations preboot and async.

@artlowel Could you please check if setting preboot to false gives the entire action history of the redux state for debugging?


wwelling:
OpaqueToken is now deprecated.

https://angular.io/docs/ts/latest/api/core/index/OpaqueToken-class.html

It will work fine for the current version of Angular we are using.

After upgrade it becomes InjectionToken.

https://blog.thoughtram.io/angular/2016/05/23/opaque-tokens-in-angular-2.html#injectiontoken-since-angular-4x

artlowel:
According to that blogpost, it shouldn't be too difficult to upgrade though: a find and replace, and removing `@Inject(GLOBAL_CONFIG)` everywhere. 
And there are some good changes here, so it seems like a good idea to merge this anyway.

However I wasn't able to get the local `environment.dev.js` or `environment.prod.js` files to work. It keeps saying `Environment file not specified. Using default.`

artlowel:
>@artlowel Could you please check if setting preboot to false gives the entire action history of the redux state for debugging?

It does

wwelling:
@artlowel I am looking into why it is not selecting the dev or prod environment config files.

wwelling:
@artlowel I have addressed the environment config files:

```yarn start```, ```yarn run server```, and ```yarn watch:prod``` will use ```environment.prod.js```
```yarn run start:dev``` and ```yarn watch:dev``` will use ```environment.dev.js```

testing will use ```environment.test.js```

if no environment config files are available, ```environment.default.js``` will be used

artlowel:
LGTM

artlowel:
Oh yeah, apart from one thing: 
I'd cut off the file structure in README.md at the depth of the app directory (as it is currently). The goal of that section was to explain the config files in the root, not to give an overview of _every_ file in the entire project. That would also require us to have to update it way too often.

wwelling:
I will update the depth of the tree in the readme.

artlowel:
+1

atarix83:
I've tested it, environment config files are used properly.
For me it's ok.

Issue #46, opened 2017-01-26
Pagination component


artlowel:
We need a way to handle pagination that can be shared between components.

- It should make use of [the ng-bootstrap pagination component](https://ng-bootstrap.github.io/#/components/pagination) for the pager.
- it should be possible to set the number of results per page - in the style of [discovery](http://demo.dspace.org/xmlui/discover) seems fine - a gear with a dropdown
- the page number, and the number of results per page should be reflected in the URL: if you share the URL to a page with a pageable component it should take you to the exact same place.
- However I can imagine we'll have pages with multiple pageable lists. adding an optional identifier to these URL params is a possible solution
- It should be easy to use as part of another component 
	- e.g. add it to the template, perhaps as a wrapper ([using ng-content tags](https://scotch.io/tutorials/angular-2-transclusion-using-ng-content)) and configure it with @Inputs and @Outputs
	- or extend a PageableComponent class
- In the prototype we had to add a bunch of properties to the models themselves to get information about the pagination to the components using it. It would be great if that wasn't necessary this time.
- An interesting project to take a look at is [ng2-pagination](https://github.com/michaelbromley/ng2-pagination), if not to use it directly (at the time of writing there are open issues about [angular universal](https://github.com/michaelbromley/ng2-pagination/issues/19) and [rollup](https://github.com/michaelbromley/ng2-pagination/issues/111)) then at least to borrow a few ideas from

PR #95, opened 2017-05-22
Pagination

atarix83:
This PR connects to #46 

Pagination component is based on [ng-bootstrap pagination](https://ng-bootstrap.github.io/#/components/pagination) and it can be used with [ng2-pagination pipe](https://github.com/michaelbromley/ngx-pagination#paginatepipe)

To configure the pager you have to use [pagination-options.model](https://github.com/4Science/dspace-angular/blob/pagination/src/app/core/cache/models/pagination-options.model.ts) that extends [NgbPaginationConfig](https://ng-bootstrap.github.io/#/components/pagination#NgbPaginationConfig), so options that are available are the same of NgbPagination documentation.

According to bootstrap responsive breakpoints, I've introduced new selector in hostWindowReducer that stores size of device window. By using new selector Pagination component is responsive and automatically resizes itself.

This is a [component](https://github.com/4Science/dspace-angular/blob/pagination-test/src/app/home/home.component.ts) and its [template](https://github.com/4Science/dspace-angular/blob/pagination-test/src/app/home/home.component.html) that show how to use it.

I've to update ng-bootstrap and bootstrap packages so I've changed header component to make it work.


artlowel:
Looks good, a few comments:
- I wouldn't add the bootstrap breakpoint to the state, as you can deduce it from the width, and we should avoid duplicating data in the state if possible. What I meant when I said on slack to "add a selector" was something like [this](https://github.com/artlowel/dspace-angular/blob/bootstrap-breakpoints/src/app/shared/host-window.service.ts)
- I can't get the links in a pagination component to work when javascript is disabled. Ideally this should also be supported server-side.
- There are new dependencies, but yarn.lock isn't updated. It should contain the version numbers of all dependencies. You can fix this by runnning `yarn` in the root of the project. In the future, you're better off not touching `package.json` manually when you want to add dependencies, but running `yarn add ${depname}`. See [README.md](https://github.com/DSpace/dspace-angular#3rd-party-library-installation) for more info.
- When building I got a warning that ng2-pagination has been renamed to ngx-pagination. If it's compatible with angular 2 we might as well update immediately

atarix83:
Hi @artlowel 

> I wouldn't add the bootstrap breakpoint to the state, as you can deduce it from the width, and we should avoid duplicating data in the state if possible. What I meant when I said on slack to "add a selector" was something like this

Thanks for the suggestion, I'm going to implement it.

> There are new dependencies, but yarn.lock isn't updated. It should contain the version numbers of all dependencies. You can fix this by runnning yarn in the root of the project. In the future, you're better off not touching package.json manually when you want to add dependencies, but running yarn add ${depname}. See README.md for more info.

I'll pay more attention in the future.

> When building I got a warning that ng2-pagination has been renamed to ngx-pagination. If it's compatible with angular 2 we might as well update immediately

I get this error with ngx-pagination
`Unexpected value 'NgxPaginationModule' imported by the module 'SharedModule'`
Do you have any suggestion?

> I can't get the links in a pagination component to work when javascript is disabled. Ideally this should also be supported server-side.

Without javascript It does not work completely, I didn't know it was a requirement to work without javascript, if so I need to get better in depth. Meanwhile, to accelerate merging, could we create a separate issue for this?

artlowel:
>I get this error with ngx-pagination
Unexpected value 'NgxPaginationModule' imported by the module 'SharedModule'
Do you have any suggestion?

Yeah I had a similar issue with the upgrade to ngx-translate. The solution was to add the prefix of its package name to `webpack.config.ts`. See [this commit](https://github.com/DSpace/dspace-angular/pull/94/files#diff-b0ba29b245f5327bca2d51ac8bbb0821L115)

I added `|@ngx-`, in this case you'll probably have to add `|ngx-` since ngx-pagination's package name doesn't start with an `@`

>Without javascript It does not work completely, I didn't know it was a requirement to work without javascript, if so I need to get better in depth. Meanwhile, to accelerate merging, could we create a separate issue for this?

A separate issue is fine.

atarix83:
@artlowel 

I review PR based on your comments.

- I've deleted breakpoint data from the state and added a service instead;
- I've added an OpaqueToken to inject 'window' object in AppComponent to allow initialization of the state (I've missed this part in the first PR commit);
- I've tried to update ngx-pagination, following your hint I don't get the error but unfortunately package works only with Angular >= 2.3.1 and I get mismatch versione error on AoT compilation;
- Adding new host-window.service I've tried to import variables from sass unsuccessful. I've tried [this](https://github.com/buildo/sass-variables-loader)  and [this](https://github.com/nordnet/sass-variable-loader) and in both cases I can't load the scss file because Angular didn't find loader module, Have you ever had any experience on this, or do you have another way to import sass variables?

artlowel:
>Adding new host-window.service I've tried to import variables from sass unsuccessful. I've tried this and this and in both cases I can't load the scss file because Angular didn't find loader module, Have you ever had any experience on this, or do you have another way to import sass variables?

No I also didn't know how to do that, I just added the TODO comment as a reminder that we should fix it some day. 

The `yarn.lock` update had as a side-effect that `@ngrx/effects` was updated to 2.0.3, and that depends on a version of rxjs that's too new, and conflicts with the version in our project. So please run `yarn add @ngrx/effects@2.0.2` to fix that problem

I also get a peer-dependencies warning about `ng-bootstrap@1.0.0-alpha.24`. It probably also needs to be rolled back a few versions

atarix83:
@artlowel 

done!

Issue #75, opened 2017-03-23
Full Item View Page


artlowel:
Add a component for the full Item view page
Its route should be `/items/:id/full`

It should contain a table that lists the key, value and language for all the Item's metadata fields, sorted alphabetically by key
Use [bootstrap's responsive table styles](https://v4-alpha.getbootstrap.com/content/tables/#responsive-tables) to make the table responsive for now, we may replace it with something a little more useable on smaller screens later

It should also contain a file section that shows the following fields for each bitstream in the ORIGINAL and LICENSE bundles
- thumbnail: reuse the component from the simple item view page
- name
- size: reuse the pipe from the simple item view page
- format
- description
- a download link

It should also contain a list of the collections the item is a direct descendant of, with a link and their short description.

PR #100, opened 2017-06-01
Full item page

LotteHofstede:
This PR connects to #75 

I designed the full item page to contain the following information:

- The title of the item
- A button to the simple item page (and vice versa: on the simple page a link to the full page)
- A table with all available metadata for this item
- A list of files (from both the ORIGINAL and LICENCE bundle) with their matching thumbnails, description, size, mimetype, name and download link
- A list of collections that are direct parents of this item

While working on the full item page, Art made some changes to the model classes that make relationships easier to handle. I.E. Bundles are now stored in an Item as a RemoteData object of an Array of Bitstreams (RemoteData<Bitstream[]>, instead of an array of remote data objects of bitstreams (Array<RemoteData<Bitstream>>).
Thanks to those changes, we were able to simplify most of the Item model functions.

An example full item page can be viewed by visiting [base-url]/items/8871/full


tdonohue:
@LotteHofstede and @artlowel: I gave this a test today. It looks good overall. However, I noticed that on the Simple Item View, the "No Thumbnail" image has now turned into a broken image. Here's a screenshot:
screenshot

LotteHofstede:
@tdonohue This happens because we added an actual thumbnail image for this item. However, since the bitstream retrieve URLs do not have an implementation yet, there is no image to show here. If you inspect the element, you'll see the image source is filled out with the right retrieve URL.

tdonohue:
@LotteHofstede : Ok, that makes sense.  However, I still wonder how we plan to correct this behavior (eventually)?  Shouldn't we have a better way to avoid showing a broken image if we cannot retrieve/display the referenced thumbnail? @artlowel any thoughts here?

All in all, this PR works. I just found it slightly odd that the thumbnail is now broken. If we have a plan to resolve this broken image in the near future, then I'm OK with merging this as-is.

artlowel:
We can work around it using the error event of the img element: https://stackoverflow.com/questions/36429903/angular-2-check-if-image-url-is-valid-or-broken

However I'd merge this as is, and then create a new issue for that problem

tdonohue:
Sounds good to me. Merging as-is. I'll log a separate ticket for broken images (as this also seems to affect Collection homepage).

Issue #109, opened 2017-06-12
Google Analytics issue with matrix URL notation


atarix83:
PaginationComponent  added in PR #95 adopt matrix URL notation for adding route params, eg. `/foo;id=15;foo=foo`

As It's reported in [this issue](https://github.com/angular/angular/issues/11740), Google Analytics doesn't seem to like semicolons in the URL and cuts off everything after them.

To avoid this behavior It's better to use standard notation with `?` and `&`

PR #110, opened 2017-06-12
remove Matrix URL notation

atarix83:
This PR connects to #109 

Remove matrix URL notation and add the possibility to keep any other parameters, in the URL, which do not belong to PaginationComponent


artlowel:
LGTM

PR #113, opened 2017-06-16
Fix Collection Homepage Errors

artlowel:
Fixed an issue where incomplete item mock data caused an error on collection homepages
Also added an unsubscribe for the logo's subscription on collection homepages

This PR connects to #112


artlowel:
This PR connects to #112

Issue #86, opened 2017-04-27
Switch to ngrx platform


artlowel:
The next version of ngrx will combine their most commonly used separate modules in to a single module: [platform](https://github.com/ngrx/platform). It will bring their versions in sync with Angular's. It hasn't been released yet.

There will be a few breaking changes. Nothing problematic for us at first sight. Here is the [migration guide](https://github.com/ngrx/platform/blob/master/MIGRATION.md)

artlowel:
I'll have to put this on hold. Everything is ported, but I get an exception when I try to JSON.stringify the router's state, in order to transfer it from the server to the client. It is a known issue with ngrx platform, see https://github.com/ngrx/platform/issues/150, which links to https://github.com/ngrx/platform/issues/97

I'll pick this up again when https://github.com/ngrx/platform/issues/97 is fixed

Issue #119, opened 2017-06-23
Pagination Issues


artlowel:
Switching to changeDetection.OnPush in [this commit](https://github.com/DSpace/dspace-angular/pull/117/commits/f57c0a4fac240e69601b4b54f6c2b1b5cfbaecef) has the effect that changing the page size no longer works. This points to an underlying problem that not all state of the pagination component is stored in the ngrx store, otherwise onPush would work.

Another likely related issues is that if you go to the second page of a paginated list, click on an item and use the browser's back button, you see the results of the first page, but the second page is still highlighted

atarix83:
I should have solved the resize issue by adding local change detection checks as [here](https://github.com/4Science/dspace-angular/commit/d46b66f87f83230b80b1d9215cd75c0ca26a834f).

About the second issue I think that the right behavior should be to get back to the last page visited, that in the example is the second page. In this case I've seen that the PaginationComponent works properly, should it be an issue related to cache or REST call? I've tested a simple paginated list page and the same use case works well.

wwelling:
Pagination has a glitch with universal. When the pre-render finishes and swaps the state on the client the pagination switches to the first page and back to the desired page quickly. It is not noticible in development mode and more so in replay redux prerender strategy. In production mode with rehydrate prerender strategy it happens but so quick it cannot be noticed. If the console is opened it slows everything down enough to notice it even in production mode.

This may be an issue with ngx-pagination library. Does it fully support Angular Universal. 

The desired behavior would be to never fetch and render any page but the desired.

This is required for smooth page refresh and deep linking.

PR #123, opened 2017-07-12
Angular upgrade: this connects to #83 and this connects to #86

wwelling:
- upgraded dependencies
- updated server and platform modules
- updated build process
- enhanced lint config
- fixed lint errors
- applied style guide for separating vendor imports
- updated readme

and for some niceties 
- added favicon generated from new DSpace logo
- added new DSpace logo on the home page


config/environment.test.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
+  // Angular2 UI server settings.
```
AlexanderS: Angular**2**?

src/app/shared/pagination/pagination.component.ts
```diff
@@ -147,209 +151,209 @@ export class PaginationComponent implements OnDestroy, OnInit {
...
+  /**
+   * Validate query params
+   *
+   * @param page
+   *    The page number to validate
+   * @param pageSize
+   *    The page size to validate
+   */
+  private validateParams(page: any, pageSize: any, sortDirection: any, sortField: any) {
+    let filteredPageSize = this.pageSizeOptions.find((x) => x === pageSize);
```
atarix83: Pagination doesn't work properly because method 'find()' doesn't work.
My editor (WebStorm) report 'Unresolved function or method find()' error.
Same error for 'Object.assign()' but this method seems to work.

Maybe some issue with new version of Typescript?

config/environment.test.js
wwelling: I will update.

src/app/shared/pagination/pagination.component.ts
wwelling: I will look into it. Object.assign is a JavaScript method. I don't think typescript should complain about it.

package.json
```diff
@@ -1,187 +1,197 @@
...
-    "morgan": "1.7.0",
-    "ng2-pagination": "^2.0.0",
-    "preboot": "4.5.2",
-    "reflect-metadata": "^0.1.10",
-    "rxjs": "5.0.0-beta.12",
-    "ts-md5": "^1.2.0",
-    "webfontloader": "1.6.27",
-    "zone.js": "0.6.26"
+    "morgan": "1.8.2",
+    "ng2-pagination": "2.0.2",
```
atarix83: Now we can update with 'ngx-pagination' package.
It should work properly, the only thing to do is to remove [these lines](https://github.com/wwelling/dspace-angular/blob/c08f5c672b1f848d1cf62f275269cd8e2b76616e/src/app/shared/pagination/pagination.component.ts#L21,L22)

wwelling: Just got into the office. Good point. I will upgrade it.

src/app/shared/pagination/pagination.component.ts
atarix83: To test the correct behavior just navigate to [this url](http://localhost:3000/collections/1c11f3f1-ba1f-4f36-908a-3f1ea9a557eb?pageId=collection-browse&page=2&pageSize=4&sortDirection=0&sortField=name) and you should see the results of the second page

wwelling: I upgraded to ngx-pagination and began debugging this issue. However, I can not recreate the problem. Also, I use two different editors; VSCode and Atom, both of which can resolve the find method and Object.assign method.

Can you try cleaning npm modules?

```
yarn run clean
yarn run global
yarn install
```

wwelling: Pull latest commit first of course. Thanks.

atarix83: No I still get the error on editor.
the real problem is on find() function that at the line I've indicated return always undefined, and this cause the problem described in the [issue](https://github.com/DSpace/dspace-angular/issues/124) you've opened

wwelling: I see. I will see if I can resolve it. I should of thought more about opened the issue. Didn't realize it was related to this.

wwelling: Ok, it turns out the problem was an exact comparison of a number with a string. 

x is a number
pageSize is a string

```
x === pageSize
```

wwelling: Pre-rendering does not work for this. It pre-renders the first page and then jumps to the page specified in the query parameters.

wwelling:
Ready for review.

src/app/shared/pagination/pagination.component.ts
atarix83: Hi,

I've looked at it.
For different types comparisation do you think that should be better to cast variable pageSize to number like these :
```
let filteredPageSize = this.pageSizeOptions.find((x) => x === +pageSize);
```

A same issue it's present [here](https://github.com/wwelling/dspace-angular/blob/c08f5c672b1f848d1cf62f275269cd8e2b76616e/src/app/shared/pagination/pagination.component.ts#L181,L182,L183) that should became : 
```
this.paginationOptions.currentPage !== +queryParams.page
|| this.paginationOptions.pageSize !== +queryParams.pageSize
|| this.sortOptions.direction != +queryParams['sortDirection']
```

About other issue, unfortunately I've noted the same problem is present with both pre-render strategy, with 'rehydrate' strategy it seems that the jump to the page specified is quicker than with 'replay' strategy.
Before upgrade this wrong behavior was not present

wwelling: I will look into it.

wwelling: Casting variables to identical type would probably be better. Would allow exact comparison. For now I will disable the lint rule for these cases. Beyond the scope of the upgrade.

You are right the pre-prender does revert to the first page. Rehydrate is just faster in navigating to the correct page. I will attempt to fix this quickly.

wwelling:
@atarix83 I just checked the behavior of the master branch and the demo. The same thing is happening, when the browser loads it quickly switch to the first page and back. Open the console and it will slow down where you can see it. Actually, the upgrade branch in production mode is slightly quicker.

https://dspace7-demo.atmire.com/home

atarix83:
@wwelling ok, so please add this wrong behavior to [this issue](https://github.com/DSpace/dspace-angular/issues/119)

wwelling:
Done. #119

atarix83:
Just only last comment.
If i don't have NODE_ENV environment variable setted and don't have neither `environment.dev.js` and `environment.prod.js` files, when run `yarn watch` i have these warnings : 
```
WARNING in ./src/config.ts
    Module not found: Error: Can't resolve '../config/environment.prod.js' in 'C:\Users\Giuseppe\Documents\Development\Javascript-Angular-Typescrypt projects\Dspace\dspace-angular-review\src'
     @ ./src/config.ts 17:51-91
     @ ./src/main.server.ts

WARNING in ./src/config.ts
    Module not found: Error: Can't resolve '../config/environment.dev.js' in 'C:\Users\Giuseppe\Documents\Development\Javascript-Angular-Typescrypt projects\Dspace\dspace-angular-review\src'
     @ ./src/config.ts 33:51-90
     @ ./src/main.server.ts
``` 

Considering that it should be in `development` environment, I'd expect only warning for `environment.dev.js`, is not that so?

wwelling:
@atarix83 yes. I will fix this immediately.

wwelling:
@atarix83 done. Upgraded Angular again in the process.

atarix83:
@wwelling  

* only few warnings when run `yarn watch` :
```
 WARNING in ./src/config.ts
    [9, 5]: Identifier 'configContext' is never reassigned; use 'const' instead of 'let'.
    [9, 37]: " should be '

     @ ./src/main.server.ts 16:0-38

WARNING in ./~/express/lib/view.js
    80:29-41 Critical dependency: the request of a dependency is an expression
```

* when run `yarn test`  test run successfully but i get an error at the end of the process: 
```
ERROR [coverage]: TypeError: Cannot read property 'split' of undefined
```

wwelling:
The first one is easy. Will commit in a second. The second existed previously.

wwelling:
@atarix83 

fixed the following:

```
 WARNING in ./src/config.ts
    [9, 5]: Identifier 'configContext' is never reassigned; use 'const' instead of 'let'.
    [9, 37]: " should be '

     @ ./src/main.server.ts 16:0-38
```

the following error existed previously and I do not think I will be able to resolve

```
ERROR [coverage]: TypeError: Cannot read property 'split' of undefined
```

here is a link to an potentially related issue: https://github.com/karma-runner/karma-coverage/issues/278

I will continue to look into the following error:

```
WARNING in ./~/express/lib/view.js
    80:29-41 Critical dependency: the request of a dependency is an expression
```

wwelling:
Only remaining warning:

```
WARNING in ./~/express/lib/view.js
    80:29-41 Critical dependency: the request of a dependency is an expression
```

I will continue to debug this. Not sure if it should block further review of this pull request.

wwelling:
@atarix83 fixed.

atarix83:
@wwelling I still get express dependency warning, other are fixed

wwelling:
@atarix83 ok, should be gone now

wwelling:
@atarix83 should be gone now

atarix83:
@wwelling perfect thanks, for me review it's ok!!

artlowel:
First of all: thanks a lot for taking on this job, it can't have been easy.
Most things seem to work just fine, but I still found a few issues:

It no longer seems to be possible to have a partial `environment.dev.js` file. It used to be overlayed on the default config using `Object.assign`. That's no longer the case, so now if you leave out options that don't need to change from the default you get an exception.

There's an issue with the logic that adds the port to `EnvConfig.baseUrl`, it looks like it's supposed leave out the port if it's 80 or 443, but that doesn't seem to work. As a consequence the cache doesn't work optimally: the HAL links in REST API responses don't contain a port, so when the client compares URLs of objects in the cache (that do contain the port, because of the baseUrl) to those, it will conclude an object isn't cached yet, and fetch it a second time.

On the angular 2 version, I changed it so that if preboot (as the option was called at the time) was disabled, SSR would be disabled completely. Without that option, if an error occurs that prevents the first page from being completed (the server doesn't even return a response) you couldn't debug it in the browser. Having the option to disable SSR entirely ensured that you could. Is that still possible?

Because it takes a while for the watch task to pick up on changes, I also added a timestamp to the "Listening at http://localhost:3000" line. That way you could tell whether the server has already restarted after your latest change. That timestamp seems to have disappeared, could you add it again please?

wwelling:
@artlowel I will find time to address these issues. None of which seem will be a problem resolving.

wwelling:
@artlowel I have addressed the issues you described. The only thing I do not like is that for CSR fallback and ability to disable SSR an additional index file is required. With Angular 4 Universal the client JavaScript bundle is now injected automatically and does not require a script tag. For CSR the client bundle script tag is required. Other than, the fixes went smoothly.

On another note, I am going to make a new issue to upgrade ngrx. The are at version 4 now. 

Below is the build errors with the latest version of ngrx. Also, ngrx-store-freeze is not yet upgraded and is incompatible with ngrx 4.

```
 WARNING in ./src/app/app.module.ts
    47:16-33 "export 'RouterStoreModule' was not found in '@ngrx/router-store'
    
    WARNING in ./src/app/header/header.effects.ts
    22:20-33 "export 'routerActions' was not found in '@ngrx/router-store'
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/app.reducer.ts (2,25): Module '"/Users/wwelling/workspace/dspace-angular/node_modules/@ngrx/router-store/router-store"' has no exported member 'RouterState'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/app.reducer.ts (40,5): Type '{ core: any; router: RouterReducerState; hostWindow: HostWindowState; header: HeaderState; }' is not assignable to type 'ActionReducer<any, Action>'.
      Type '{ core: any; router: RouterReducerState; hostWindow: HostWindowState; header: HeaderState; }' provides no match for the signature '(state: any, action: Action): any'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/shared/host-window.service.ts (26,12): Expected 1-6 arguments, but got 2.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/object-cache.service.ts (68,12): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/object-cache.service.ts (73,12): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/object-cache.service.ts (85,12): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/object-cache.service.ts (126,5): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/object-cache.service.ts (145,5): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/response-cache.service.ts (37,12): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/response-cache.service.ts (54,5): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/data/request.service.ts (31,5): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/data/request.service.ts (41,12): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/builders/remote-data-build.service.ts (37,7): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/builders/remote-data-build.service.ts (39,9): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/cache/builders/remote-data-build.service.ts (114,24): Expected 1-6 arguments, but got 4.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/header/header.effects.ts (3,10): Module '"/Users/wwelling/workspace/dspace-angular/node_modules/@ngrx/router-store/router-store"' has no exported member 'routerActions'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/core.effects.ts (9,17): Property 'run' does not exist on type 'typeof EffectsModule'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/core.effects.ts (10,17): Property 'run' does not exist on type 'typeof EffectsModule'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/core/core.effects.ts (11,17): Property 'run' does not exist on type 'typeof EffectsModule'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/app.effects.ts (9,17): Property 'run' does not exist on type 'typeof EffectsModule'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/app.effects.ts (10,17): Property 'run' does not exist on type 'typeof EffectsModule'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/header/header.component.ts (22,48): Argument of type '"header"' is not assignable to parameter of type '"navCollapsed"'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/app.module.ts (6,10): Module '"/Users/wwelling/workspace/dspace-angular/node_modules/@ngrx/router-store/router-store"' has no exported member 'RouterStoreModule'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/app.module.ts (46,17): Property 'provideStore' does not exist on type 'typeof StoreModule'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/app.module.ts (48,25): Property 'instrumentOnlyWithExtension' does not exist on type 'typeof StoreDevtoolsModule'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/browser-app.module.ts (59,19): Property 'run' does not exist on type 'typeof EffectsModule'.
    
    ERROR in /Users/wwelling/workspace/dspace-angular/src/app/server-app.module.ts (77,19): Property 'run' does not exist on type 'typeof EffectsModule'.
    
    ERROR in Error encountered resolving symbol values statically. Calling function 'EffectsModule', function calls are not supported. Consider replacing the function or lambda with a reference to an exported function, resolving symbol ServerAppModule in /Users/wwelling/workspace/dspace-angular/src/app/server-app.module.ts, resolving symbol ServerAppModule in /Users/wwelling/workspace/dspace-angular/src/app/server-app.module.ts
```

artlowel:
Thanks.

During today's meeting we decided to merge this as is, and re-open the ngrx-platform issue, to upgrade it in a separate PR.

Issue #128, opened 2017-07-27
Old views stay visible when the route changes


artlowel:
If you go to the "DSpace related records in PubMed Europe" collection, and click on a Recent Submission, the item view page opens, but the collection homepage doesn't disappear. The Item homepage is rendered above it.

If you keep clicking links afterwards this keeps happening, the new page opens above the old page.

This only seems to happen for that specific collection, and the bug was introduced with [the upgrade to Angular 4](https://github.com/DSpace/dspace-angular/pull/123).

![screen shot 2017-07-27 at 15 44 59](https://user-images.githubusercontent.com/1567693/28673353-91ec7e74-72e2-11e7-98ee-4f7c8e481961.png)

wwelling:
Seems to only occur on /items/6212604b-c778-42b3-88e4-cc3b1262ad28

Solution structure of apamin determined by nuclear magnetic resonance and distance geometry

wwelling:
https://dspace7-demo.atmire.com/collections/8e0928a0-047a-4369-8883-12669f32dd64

try clicking on 

Solution structure of apamin determined by nuclear magnetic resonance and distance geometry

artlowel:
I found the issue, it was a preexisting bug. The symptoms have just changed due to the update. I'll open a PR soon

wwelling:
Good to know. Symptoms seem less severe after the upgrade.

PR #129, opened 2017-07-27
fixed an issue where if there is only one bitstream it would be retur…

artlowel:
…ned as a single object and not an array in the item.bitstreams property.

This bug would throw an exception and as a consequence the router wouldn't remove the previous view:

This PR connects to #128


wwelling:
LGTM

Issue #91, opened 2017-05-11
404 page status code


artlowel:
The status code of the current 404 page is 200. It should be 404.

rivaldi8:
I've searched for a way to set the status code, but I haven't found any reference or example of how to do so. All examples I've seen redirect to a component like our `PageNotFoundComponent`, but don't do anything about the status code.

Is there some well-known way to do this that I've missed?

artlowel:
Maybe this is of some help?:
https://github.com/angular/universal/issues/667

We can only set the status code on the server side for the initial page request, afterwards, in the client-side code you can no longer change the status code, because the request that has already arrived at the browser. That is probably why examples that don't work with universal don't set the status code

rivaldi8:
Ok, I hadn't thought about Universal. Anyway, I see it's a bit tricky. I'd rather go after easier stuff for now :)

patrickmichalina:
@rivaldi8 this way is easier:

https://github.com/patrickmichalina/fusebox-angular-universal-starter/blob/master/src/client/app/shared/services/server-response.service.ts

rivaldi8:
@patrickmichalina Thanks for the tip! I'll try again.

PR #142, opened 2017-09-01
Bugfix 91:  404 page status code

rivaldi8:
This connects to #91.

I've kept `ServerResponseService` to the minimum, but I can add more methods if you think it might be useful in the future.

### DIFF
--- a/src/app/pagenotfound/pagenotfound.component.ts
+++ b/src/app/pagenotfound/pagenotfound.component.ts
@@ -9,8 +9,6 @@ import { Component, ChangeDetectionStrategy } from '@angular/core';
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class PageNotFoundComponent {
-
-  data: any = {};
   constructor(responseService: ServerResponseService) {
     responseService.setNotFound();
   }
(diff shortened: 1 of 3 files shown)


src/app/pagenotfound/pagenotfound.component.ts
```diff
@@ -1,12 +1,15 @@
-import { Component } from '@angular/core';
+import { ServerResponseService } from '../shared/server-response.service';
+import { Component, ChangeDetectionStrategy } from '@angular/core';
 
 @Component({
   selector: 'ds-pagenotfound',
   styleUrls: ['./pagenotfound.component.scss'],
-  templateUrl: './pagenotfound.component.html'
+  templateUrl: './pagenotfound.component.html',
+  providers: [ServerResponseService],
```
artlowel: Is there a particular reason for specifying the provider at the component instead of the module. Doing so will create a separate instance of the provider for each instance of the component. If you specify it at the module level it's a singleton. 

https://angular.io/guide/ngmodule-faq#should-i-add-other-providers-to-a-module-or-a-component

rivaldi8: No reason, just ignorance :)

The proper place should be in `SharedModule`, isn't it?

artlowel: indeed

rivaldi8:
I've implemented the suggested changes.

artlowel:
LGTM

PR #143, opened 2017-09-07
Switch to ngrx platform

artlowel:
This PR connects to #86


artlowel:
This merge closes #86

Issue #121, opened 2017-07-07
Detail page for an item without fulltext is broken


abollini:
If an item doesn't have bitstreams in the original bundle, a blank page is shown.

I have reinstated an item on the public REST demo so to have a test item that can be used to work on this bug
http://dspace7.4science.it/dspace-spring-rest/api/core/items/6212604b-c778-42b3-88e4-cc3b1262ad28/bitstreams

rivaldi8:
I'm no longer able to reproduce this issue. Has it been fixed?

artlowel:
It seems like it. Most likely due to the angular 4 upgrade.
One thing that should still be fixed however, is that the "Files" header is still shown if there are no bitstreams

PR #146, opened 2017-09-14
Bugfix 121: Detail page for an item without fulltext is broken

rivaldi8:
This connects to #121


artlowel:
LGTM

src/app/item-page/simple/field-components/file-section/file-section.component.html
```diff
@@ -1,4 +1,4 @@
-<ds-metadata-field-wrapper [label]="label | translate">
+<ds-metadata-field-wrapper *ngIf="(files | async).length > 0" [label]="label | translate">
```
artlowel: It's probably safer to add a question mark after the right bracket, to ensure it doesn't break if `(files | async)` returns null or undefined for any reason.
```typescript
(files | async)?.length
```

rivaldi8: Done!

artlowel:
Looks good, merging

Issue #147, opened 2017-09-15
Remove hardcoded REST Endpoints


artlowel:
- remove the hardcoded endpoints from the data services, and have each service store its corresponding linkName in the api endpoint
	- https://dspace7.4science.it/dspace-spring-rest/#/dspace-spring-rest/api
- When the UI is initialized, the api endpoint should get fetched and the mapping of link names to endpoints should be put in the ngrx store
- the dataservices should use their linkName field to fetch the actual endpoint url form the store with every call to the rest api
- They also should get a method to check if the rest backend exists for their linkName
- Currently, none of the other (non-root) endpoints should be hardcoded, but please verify that this is the case

PR #149, opened 2017-09-20
Improve support for hal links

artlowel:
This PR connects to #147 

It removes the need for hardcoded resource endpoints, and adds a method to check whether a dataservice's endpoint is enabled on the rest api

It doesn't yet remove the need for hardcoded browse endpoints. That will be fixed in a separate PR.


wwelling:
@artlowel I would recommend merging this before #152

Only one file conflicts now. It could be many more if #152 is merged first.

artlowel:
@wellingWilliam as you wish

Issue #145, opened 2017-09-14
Base Path support


paulo-graca:
As you build your app, using default settings, your project will work under:
http://localhost:3000/

But if you need to use a different base url like http://localhost:3000/angular/
You can use the simplest solution wich is to set the HTML5 `<base>` tag in the index.html file:
`<base href="/angular/">`

changes here can be a problem regarding the config settings in different environments. In a production environment you might not want to use the "/angular" base path.

Other solution is to use the APP_BASE_HREF, 
`<base ng_href="{{APP_BASE_HREF}}">`

and in the app module, include:
```
import { APP_BASE_HREF } from '@angular/common';

@NgModule({
providers: [
    { provide: GLOBAL_CONFIG, useFactory: (getConfig) },
    { provide: APP_BASE_HREF, useValue: '/angular/'}
  ],
```
I think this base path should be a supported enviornment setting (ServerConfig already includes the **baseUrl** property).

artlowel:
There is already a property for that: [ui.nameSpace](https://github.com/DSpace/dspace-angular/blob/master/config/environment.default.js#L8) but a quick test shows that it doesn't work with any value other than `/` so that should be fixed.

wwelling:
This has been resolved, but not closed. Closing now.

atarix83:
I reopen this issue because not all works properly. If you add a property for that: ui.nameSpace, the images and the fonts that are not loaded.
Indeed `assets/images` and `assets/fonts` paths does not consider the namespace. 
Also the client.js is not found because we're using an absolute path here : 
https://github.com/DSpace/dspace-angular/blob/605c3524c813fd019005387eb643e3d00fa0ba6e/src/index.csr.html#L17

Issue #139, opened 2017-08-17
Complex config objects don't merge with the default config


artlowel:
If you override a part of the config that's a complex object (e.g. the ui or rest properties), you currently need to copy all children of that complex object from the default config, instead of only adding the ones that differ, otherwise the options you haven't overridden will be missing.

This differs from the behavior at the root level, where the default properties that aren't overridden are merged in.

wwelling:
This has been resolved.

Issue #126, opened 2017-07-19
Upgrade @ngrx


wwelling:
This will have to wait until https://github.com/codewareio/ngrx-store-freeze/issues/17 is resolved.

artlowel:
I'd rather upgrade to @ngrx/platform as soon as possible. The ngrx team has been closing a quite a few issues I was watching with the notification that they're fixed in platform. 

Ngrx-store-freeze is just a tool that ensures objects stay immutable during development.
if you write a reducer, you should write tests for that immutablility anyway
ngrx-store-freeze just makes it easier to detect those issues early.

I say we leave store-freeze out now, and create an issue to put it back in as soon as it's up to date

StevenLiekens:
Hey @artlowel the store-freeze tool is up to date with ngrx v4 now.

artlowel:
Thanks for the heads up.

adding ngrx-store-freeze back in is trivial. But I can't seem to figure out an AoT compatible way to determine whether we're in production mode, in order to decide whether or not to add ngrx-store-freeze to AppModule's imports. 

I've tried all sorts of things, using it directly, using the getConfig() function, wrapping the imports in another function, wrapping the entire initialization of ENV_CONFIG in config.ts in a function … I keep running in to the `Error encountered resolving symbol values statically` error

@wellingWilliam can you have a look? You can find my current, broken branch [here](https://github.com/artlowel/dspace-angular/tree/ngrx-store-freeze-update)

wwelling:
@artlowel I will try to make some time to look at this later today or tomorrow.

wwelling:
This upgrade is complete. Closing now.

PR #152, opened 2017-09-27
Upgrade: this connects to #145, this connects to #139, this connects to #126 

wwelling:
1. dependency upgrades
2. dependency injection of meta resolvers
3. conditionally included development modules
4. recursive configuration merging
5. functional base from namespace UI configuration
6. corrected use of core and shared modules
7. lazy routes
8. simplified build, single main server file


wwelling:
Shared module should not contain app wide singletons. 

https://angular.io/guide/ngmodule#shared-modules

Core module should only be imported to the app module.

https://angular.io/guide/ngmodule#the-core-module

src/config.ts
```diff
@@ -41,7 +42,7 @@ try {
 // if envConfigFile set try to get configs
 if (envConfigFile) {
   try {
-    ENV_CONFIG = Object.assign(ENV_CONFIG, configContext(envConfigFile)) as GlobalConfig;
+    merge(configContext(envConfigFile) as GlobalConfig);
   } catch (e) {
     console.warn('Cannot find file ' + envConfigFile.substring(2, envConfigFile.length), 'Using default environment.');
```
artlowel: If this happens (e.g. when you're in dev mode, but don't have a custom environment.dev.js) , buildBaseUrls() is never executed, and the methods that rely on baseUrls break.

src/app/app.metareducers.ts
```diff
@@ -33,6 +32,6 @@ export function universalMetaReducer(reducer) {
 }
 
 export const appMetaReducers = [
-  // debugMetaReducer,
+  debugMetaReducer,
```
artlowel: Any particular reason for re-enableing this? 
I made that metareducer to have some idea what happened to the store when @ngrx/store-devtools had an issue with the angular 4 router. But that issue is fixed now. I left the debug metareducer there, commented out, in case we should need it again.

src/config.ts
wwelling: Good catch. I will fix this. I meant to separate the merge and configContext from the try, but didn't catch the base url construction. Thanks.

src/app/app.metareducers.ts
wwelling: No reason. Maybe the debug metareducer should be conditionally added from the debug configuration property.

artlowel: That's fine

wwelling:
This pull request also begins to address issue #48, but not to completeness. There are some naming refactoring that should be done, some better encapsulation of providers rather than bundle them in core module, separate shared better, and of course go through remaining style guide.

wwelling:
@artlowel let me know if you choose not to merge this pull request first. Also, if you wish I do some merging I do not mind. FYI, I have node 8 engine installed on all my machines and cannot build with the upgrade of this pull request or reverting to earlier version node again.

PR #157, opened 2017-09-29
Coveralls, this connects to #155

wwelling:


wwelling:
this connects to #155

Issue #161, opened 2017-10-04
Merge issues with Simple Search


LotteHofstede:
After merging 3 large pull requests into the master branch, the Simple Search page broke.
This needs to be fixed before any other changes to the simple search page can be made.

PR #162, opened 2017-10-04
Simple Search fixes for merge related issues

LotteHofstede:
This PR connects to #161 
Added some routing fixes so the search page can be used again.
Updated the search-page component to also use lazy loading.
Some other minor (merge related) fixes/updates to the simple search feature.


wwelling:
LGTM

PR #180, opened 2017-10-12
Remove hardcoded browse endpoints

artlowel:
This PR connects to #147 

It is still a work in progress. 
The code should be done, but I need to add more tests.


artlowel:
This PR is now ready for review

The performance issues I mentioned during last week's meeting turned out to be caused by inadvertently triggering Angular's change detection in the RemoteDataBuildService (sometimes dozens of times a second). 

Fixing the problem made a different issue come to light: Angular's change detection doesn't work automatically on RemoteData objects used as `@Input()`, they're too complex. We've been having this issue all along but weren't able to notice it because change detection was triggered by the RemoteDataBuildService all the time.

The solution was to turn all observable properties from RemoteData in to regular properties, and instead having the RemoteDataBuildService return an `Observable<RemoteData>`. So now, every time something about the RemoteData object's state changes, a complete RemoteData object is emitted, not just the property that changed.

Because the observable is now on the outside, we _can_ rely on Angular to automatically detect changes. It also makes working (and testing) with remote data objects easier.

The change also allowed me to switch to `ChangeDetection.OnPush` everywhere, which further reduces the number of times change detection needs to run.

artlowel:
As we discussed yesterday, I refactored all the `async` usages in the templates using a [custom `ngVar` directive](https://github.com/DSpace/dspace-angular/pull/180/files#diff-eee395f6cbdbf381695bbd5b1295a87c), that allows us to [define a variable outside an `ngIf`](https://github.com/DSpace/dspace-angular/pull/180/files#diff-f0d2adfbbfa37d7bc151da2039fcf896R1)

src/app/object-list/object-list.component.html
```diff
@@ -10,11 +10,9 @@
   (sortDirectionChange)="onSortDirectionChange($event)"
   (sortFieldChange)="onSortDirectionChange($event)"
   (paginationChange)="onPaginationChange($event)">
-  <ul *ngIf="objects.hasSucceeded | async" @fadeIn> <!--class="list-unstyled"-->
-      <li *ngFor="let object of (objects.payload | async) | paginate: { itemsPerPage: (pageInfo | async)?.elementsPerPage, currentPage: (pageInfo | async)?.currentPage, totalItems: (pageInfo | async)?.totalElements }">
+  <ul *ngIf="objects?.hasSucceeded"> <!--class="list-unstyled"-->
+      <li *ngFor="let object of objects?.payload | paginate: { itemsPerPage: pageInfo?.elementsPerPage, currentPage: pageInfo?.currentPage, totalItems: pageInfo?.totalElements }">
```
atarix83: Pipe "paginate" can be remove. It should be used when we're not doing server-side paging and we have the entire object to paginate

artlowel: Makes sense, thanks

PR #181, opened 2017-10-13
Meta Service

### LINKED ISSUE #104: Title and Meta Tags
When changing the route, the page title should update.

Also, in order to be indexable by Google Scholar, all Item View pages should generate certain meta tags.

You can find google scholar's specs [here](https://scholar.google.com/intl/us/scholar/inclusion.html#indexing).

You can take a look at [the prototype's GoogleScholarMetaService](https://github.com/DSpace-Labs/angular2-ui-prototype/blob/master/src/app/utilities/services/google-scholar-metadata.service.ts) for inspiration, but back then we made our own MetaService because nothing existed yet. 

Now I'd look in to something like [@ngx-meta](https://github.com/ngx-meta/core) to do it for you.

wwelling:
this resolves #104 

- meta service
  - google scholar citations meta tags for item
  - route based meta tags
  - test coverage
- updated model to include serialized bitstream format
- upgraded dependencies
- refactored testing files for consistency *not fixed to naming convention


src/app/core/metadata/metadata.service.ts
```diff
@@ -0,0 +1,390 @@
...
+        return { params: route.params, data: route.data };
+      }).subscribe((routeInfo: any) => {
+        this.processRouteChange(routeInfo);
+      });
+  }
+
+  private processRouteChange(routeInfo: any): void {
+    this.clearMetaTags();
+    if (routeInfo.params.value.id && routeInfo.data.value.type) {
+      this.objectCacheService.getByUUID(routeInfo.params.value.id, routeInfo.data.value.type)
```
artlowel: I'm not sure reconstructing the object again from route params in this service is the way to go.

This way we're duplicating the code to fetch the main object the route is about which is quite complicated and not guaranteed to work the same way for every future route. 

To do so you also had to add things like normalized objects to the route data. In my opinion we should try to avoid accessing normalized objects and the ObjectCacheService directly everywhere but in the data services.

It makes more sense to me to use the object that's already fetched by the page components (e.g. [in the CommunityPageComponent](https://github.com/DSpace/dspace-angular/blob/master/src/app/+community-page/community-page.component.ts#L34)), and to have those page components set the currentObject in the MetadataService, or if you don't want to couple the page components with the MetadataService, perhaps create an intermediate service that holds the object(s) associated with the current route so any other service that might need it can fetch it there.

wwelling: I wanted to avoid a switch statement conditioning against the route in order to fetch the DSpaceObject.

wwelling: https://angular.io/api/platform-browser/Title

Setting the title is easy.

wwelling: I am ok with not using ObjectCacheService and RemoteDataBuildService if there was a way to get a serialized DSpaceObject from the DataService. I think it is to static if we had to condition to determine which service to use to retrieve the DSpaceObject. 

CommunityDataService, CollectionDataService, ItemDataService, and any future services. 

Also, if the ObjectCacheService and RemoteDataBuildService are not meant to be used by other services, they probably should not be injectable.

artlowel: The reason there's no general DSpaceObject dataservice is because there's currently no way to fetch a resource from the rest api without specifying its type. 

My idea for the dataservices was that they represent the dso endpoints on the rest api. The components that use them shouldn't need to know whether the resource they need is cached or not, they always call the find method in the same way, as if they were directly querying the rest api. 

That's why, even though we could write a service that fetches any old DSO from the cache regardless of its type, I haven't done so. Because you can't do that on the rest api and I want the cache to be invisible for parts of the code that don't directly need to interact with it.

The reason the ObjectCacheService etc are injectable is because they _do_ need to be injected in to some backend services, but ideally not everywhere. Perhaps that can be enforced by better use of modules.

wwelling: Do you suggest injecting CommunityDataService, CollectionDataService, and ItemDataService into the MetadataService and conditioning on the route to get the DSpaceObject from its corresponding service?

wwelling: Or is there an alternative? 

The reason I feel safe about using the ObjectCacheService is that the intended route will have already fetched its DSpaceObject. The route than defined the type and it knows how to retrieve what the component already requested.

wwelling: I didn't investigate the code, but I assume if two requests where made right after each other, the second would just piggy back on the first and wait for its response.

artlowel: >Do you suggest injecting CommunityDataService, CollectionDataService, and ItemDataService into the MetadataService and conditioning on the route to get the DSpaceObject from its corresponding service?

No, I suggest keeping them all out of MetadataService, and rather injecting MetadataService in to the page components, and setting the current object on MetadataService from the page component. Or if you want to avoid injecting MetadataService directly in to the page components, creating an intermediary service that knows which objects correspond to the current route. Have the page components set the objects on that service, and have MetadataService fetch them from that service.

wwelling: So be it.

wwelling:
@tdonohue I am all for a single source of mock data. In addition it would be nice for a suite of utilities to offer up the mock data in various formats that can be used within specs. And it makes since the mock data not be the same as the mock data currently coming from the REST API in order to distinguish the two.

With that said, I am unsure the status of the backend data. It is most likely out of date and will need to be updated. I do know that the backend service is not functional any longer. That may need to be repaired as well. I will try to find some time work a bit on the mock data.

wwelling:
@artlowel I have made the changes you had requested. MetadataService is now injected into the CommunityPageComponent, CollectionPageComponent, and ItemPageComponent.  The components pass the corresponding RemoteData to the MetadataService to process its payload and add the meta tags. I decided against the proxy service to manage the object/route mapping since there is an eminent change on the REST API that will embed the meta tag properties and values for a given schema.

@tdonohue I will commit any mock data improvements I make while this is under review. But, please don't defer the merge on that account. I may not get to them anytime soon.

artlowel:
Thanks. LGTM

PR #182, opened 2017-10-17
Bug 171:  List grid view switcher

### LINKED ISSUE #171: List grid view switcher
Create a component to switch between the list and the grid views for sets of items, as depicted in [the mockups on the wiki](https://wiki.duraspace.org/display/DSPACE/Search+Mockup#SearchMockup-Facetsareshowninasidebar)

The icons should be links that redirect to the current page (with any set of request params it may already have) and set the view param to `list` or `grid` accordingly, overwriting it if it already exists.

The reasons to make them links are
	- to ensure they also work with js disabled.
	- for usability sake: since they redirect to a different page they should be `a` tags

Use font-awesome icons: [list](http://fontawesome.io/icon/list/) and [grid](http://fontawesome.io/icon/th-large/). Font awesome should already be included in the project

If the search sidebar component is already finished, put this component on the sidebar as shown in the mockup. If it isn't, just add it to the top of the search page somewhere. It will be moved to its correct position later.

rivaldi8:
Fixes #171

@artlowel finally I didn't need to return the URL from `SearchService` to make it work without JavaScript. I hadn't tested it yet when we discussed, but it turns out the non-JavaScript case was already working through the attributes in the `a` tags.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -81,6 +81,10 @@
     },
     "results": {
       "title": "Search Results"
+    },
+    "view-switch": {
+      "show-list": "Show as list",
+      "show-grid": "Show as grid"
     }
   },
   "loading": {
--- a/src/app/+search-page/search-options.model.ts
+++ b/src/app/+search-page/search-options.model.ts
@@ -1,7 +1,13 @@
 import { SortOptions } from '../core/cache/models/sort-options.model';
 import { PaginationComponentOptions } from '../shared/pagination/pagination-component-options.model';
 
+export enum ViewMode {
+  List = 'list',
+  Grid = 'grid'
+}
+
 export class SearchOptions {
   pagination?: PaginationComponentOptions;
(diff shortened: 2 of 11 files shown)


rivaldi8:
The build has some errors. I'm working on it.

rivaldi8:
Fixed!

artlowel:
>it turns out the non-JavaScript case was already working through the attributes in the a tags.

That's good to know, it makes things more convenient.

It all looks good, and works, but would you mind writing tests for the new bits of code you added? [We recently decided to put more emphasis on testing](https://github.com/DSpace/dspace-angular/pull/151#pullrequestreview-65985173). That's also why coveralls was added.

rivaldi8:
Sure! I'll work on it.

rivaldi8:
Done!

artlowel:
LGTM

LotteHofstede:
LGTM

Issue #174, opened 2017-10-05
Content width


artlowel:
The goal of this task is to limit the max-width of most of the content in dspace, but give us the option to keep certain things (background-colors most likely) full width.

Wrap the contents of all components with `-page` in their name in a div with the class `container`. 

In the home-component, wrap the search form and top-community list components in such a div, but leave out the news, it's background needs to retain its full width

In the home-news component change the container-fluid div to container.

Ensure you don't get double margins anywhere, by looking at all changed pages in varying window sizes, if you do, you likely need to add a div with class `row` inside the container, as it contains a negative margin that will offset the container's padding

courtneypattison:
Hey, could I please get an invite to contribute? I'd like to submit a PR for this issue, but I don't want to mess with your workflow!

artlowel:
Sure! I've sent you an invite.

PR #184, opened 2017-10-25
Limit content width for pages

courtneypattison:
This PR connects to #174 

Apologies for any mistakes I make: I'm a noob.

In the bootstrap docs it says that although containers can be nested, most layouts don't need them, so I removed the app level `container-fluid` class and wrapped each page in a `container` to make sure there wasn't any nesting. For rows without columns, I added the `no-gutters` class, but maybe the content in those rows should be in columns, then `no-gutters` could be removed. Also, I changed the shared logo's class from `img-responsive` to `img-fluid` because it was messing up the content width for the sub-collection-list for mobile sizes.

It looks like I changed way more than I did because of indenting!

Please let me know what I need to fix :)

### DIFF
--- a/src/app/+collection-page/collection-page.component.html
+++ b/src/app/+collection-page/collection-page.component.html
@@ -1,50 +1,52 @@
-<div class="collection-page">
-  <div *ngIf="collectionData.hasSucceeded | async" @fadeInOut>
-    <div *ngIf="collectionData.payload | async; let collectionPayload">
-      <!-- Collection Name -->
-      <ds-comcol-page-header
-        [name]="collectionPayload.name">
-      </ds-comcol-page-header>
-      <!-- Collection logo -->
-      <ds-comcol-page-logo *ngIf="logoData"
-        [logo]="logoData.payload | async"
-        [alternateText]="'Collection Logo'">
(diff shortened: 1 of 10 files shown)


LotteHofstede:
Hey Courtney!

Thanks for your PR, it looks really good.

I have one small remark about the bootstrap classes you added in the news `jumbotron` on the home page.
As you mentioned, something is a bit off with using the `no-gutter` class to undo the negative margin that's created by the `row` classes here.

I personally think you don't need the `row` class at all here. Rows are reserved for when you need the grid system, which you don't in this case, because there is no defined width for each child object.

You probably noticed that when you remove the `row` class, the header jumps below the DSpace logo. This happens because the row class adds `display: flex` (in bootstrap 4).
You can simply add the `d-flex` class here instead.

artlowel:
>I tried to resolve the search page's conflict with 273da5e. Sorry if that was the wrong thing to do

That was the right thing to do, but you may have seen that github still said the PR had merge conflicts afterwards. It turns out that was an issue in github itself. I hadn't seen it before, but I couldn't find anything amiss with the PR, so I contacted github support, and they fixed it.

The PR looks good. I'll merge it.

Issue #185, opened 2017-10-25
Pagination dropdown menu placement


courtneypattison:
Currently the pagination dropdown menu, seen on the search page after clicking the gear icon, is placed to the bottom left of the button, showing the menu off screen for smaller screen sizes. Maybe placing it to the bottom right would be better.

PR #186, opened 2017-10-25
Moved dropdown menu placement to the bottom right

courtneypattison:
This PR connects to #185

Used NgbDropdown's input `placement` to specify where the menu should go rather than using a bootstrap class.


artlowel:
Obvious fix, thank you.

PR #198, opened 2017-11-09
fixed an issue where pagination events would emit after every click

### LINKED ISSUE #197: Fade-in's after every click
On paginated lists, on any page but the first one, every click anywhere on the page will cause the list to fade in again.

artlowel:
This PR closes #197

However, I'd like to keep this PR open for now, because there are other places where fade-in's happen too often: e.g. on a collection home page, the collection part at the top fades-in once, when it has loaded, and again when the recent items finish loading. I'd like to look in to them first


artlowel:
closed in favour of #220

PR #205, opened 2017-11-24
Increase test coverage

LotteHofstede:
This PR connects to #130 
I added a couple of tests to increase our general test coverage. 
Will add some more in the upcoming week.


LotteHofstede:
@wwelling Do you have any idea why the test coverage has been cut in half by **adding** 1 new .spec file? I know that we previously had this issue because not all files were imported, but this was fixed in #179. As far as I can see, my new .spec file does not perform any strange new imports...

Issue #194, opened 2017-11-06
Update to Angular 5


artlowel:
[Angular 5 was released](https://blog.angular.io/version-5-0-0-of-angular-now-available-37e414935ced)

At first glance, the only things that need to change to be compatible are:
- a few OpaqueTokens that need to be replaced by InjectionTokens
- we need to replace Http with HttpClient in the DSpaceRESTv2Service

The following dependencies have no compatible version yet:
- `@angularclass/idle-preload`: 
  - Symptom: Breaks build (uses OpaqueToken)
  - Issue: https://github.com/AngularClass/angular2-idle-preload/issues/6
- `@ngrx/platform`: 
  - Symptom: peer-dependencies warning
  - Issue: https://github.com/ngrx/platform/issues/538
- `ng-bootstrap`: 
  - Symptom: peer-dependencies warning
  - Issue: https://github.com/ng-bootstrap/ng-bootstrap/issues/1938

@wellingWilliam do we still need idle-preload? it hasn't been updated in almost a year…

artlowel:
After the update we can remove our custom transfer-state module, and [switch to the new built-in one](https://medium.com/@evertonrobertoauler/angular-5-universal-with-transfer-state-using-angular-cli-19fe1e1d352c)

born2net:
sorry if repeating myself, but we are using previous version of ngrx and wondering if we can have it also be maintained so we can upg to ng5 (opaque token removed version)

regards

Sean

artlowel:
@born2net You're asking in the wrong place. This project makes use of ngrx. Try https://github.com/ngrx/platform/issues

born2net:
Oops ... sorry. tx!

artlowel:
Angular 5 requires a different AoT webpack plugin: https://github.com/angular/angular-cli/tree/master/packages/%40ngtools/webpack

artlowel:
ng-bootstrap should be compatible now. They closed https://github.com/ng-bootstrap/ng-bootstrap/issues/1938 and [released a new version](https://github.com/ng-bootstrap/ng-bootstrap/releases/tag/1.0.0-beta.6)

tech-no-logical:
I installed beta 6, but still have the warning : 

`npm WARN @ng-bootstrap/ng-bootstrap@1.0.0-beta.6 requires a peer of @angular/forms@^4.0.3 but none was installed.`

PR #210, opened 2017-12-19
Replaced Http with HttpClient

courtneypattison:
Relates to #194 

I tried to replace Http with HttpClient in the DSpaceRESTv2Service and ApiService, removing unused imports.

It doesn't seem like ApiService is used, so I didn't add tests, but I did add a test suite for the DSpaceRESTv2Service. It is pretty much a boilerplate HTTP test suite though because I'm not sure how to create a mock DSpaceRESTv2Response (it's just an empty object for now).

Yet again, I don't really know what I'm doing, so I'd appreciate any suggestions on how to improve this PR! Thank you :)

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -1,6 +1,6 @@
 import { NgModule } from '@angular/core';
 import { CommonModule, APP_BASE_HREF } from '@angular/common';
-import { HttpModule } from '@angular/http';
+import { HttpClientModule } from '@angular/common/http';
 
 import { EffectsModule } from '@ngrx/effects';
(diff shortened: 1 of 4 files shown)


courtneypattison:
Ahh, that makes sense! Thank you for the explanation :)

If I've understood correctly, the `get()` method should be properly returning a DSpaceRESTV2Response now.

For the tests, I used the HttpTestingController to mock the httpClient and added some error testing.

Please let me know if there is anything else I should change!

artlowel:
Looks good now, thanks for making the changes

Issue #217, opened 2018-01-29
Post and patch response cache 


atarix83:
When executing more post or patch request, the response retrieved by `ResponseCacheService` is not the latest received from the server but the first that has been cached.

To test this behavior I've created a branch here (change the  REST API server to https://dspace7.dev01.4science.it) : 
https://github.com/4Science/dspace-angular/tree/post-patch-cache-issue

In the home page are executed three patch operations on the same workspaceitem.
The responses returned by the service and logged to console are the same, while could be different.
In the network panel of the browser's console the you can check that the responses from the server are correct.

To temporarily avoid this problem I've modified `RequestService` so below (to reproduce the issue change is commented out) : 
https://github.com/4Science/dspace-angular/blob/b105f47a743c256ef2a8d99f8593b0bba80dacdb/src/app/core/data/request.service.ts#L106

Could this be a fix or other corrections are necessary?

PR #230, opened 2018-03-01
Authentication by credentials

atarix83:
This PR implements authentication by credentials.

this module was development following this blog spot http://brianflove.com/2017/04/10/angular-reactive-authentication/
Unit tests are missing at the moment, but I've created PR anyway so it can be review.

There is a known issue with SSR : 
indeed to allow to retrieve an authenticated session JWT is saved in cookie browser. Every time the application is bootstrapped there is a check to verify the existence of the JWT and its status. Since JWT token validation uses the client ip to also validate the token, if universal is enabled validation fails, because ip is different.

This problem doesn't occur when running application on local environment

While waiting for a solution, to try it client ip validation can be disable by config https://github.com/DSpace/DSpace/blob/master/dspace/config/modules/authentication.cfg#L81


src/app/+home-page/home-page.component.html
```diff
@@ -1,5 +1,6 @@
 <ds-home-news></ds-home-news>
 <div class="container">
+  <p *ngIf="(isAuthenticated | async)">Loggato</p>
```
artlowel: Is this line left here as a test? It breaks the AoT build, as isAuthenticated isn't defined in HomePageComponent

src/app/shared/log-in/log-in.component.scss
```diff
@@ -0,0 +1,32 @@
...
+  margin-bottom: -1px;
+  border-bottom-right-radius: 0;
+  border-bottom-left-radius: 0;
+}
+.form-login input[type="password"] {
+  margin-bottom: 10px;
+  border-top-left-radius: 0;
+  border-top-right-radius: 0;
+}
+
```
artlowel: I'd try avoid writing all this css. You can accomplish most of what `log-in.component.scss` does with a few choice boostrap classes and the end-result will more robust on different screen sizes. Since bootstrap 4 has classes for things like [spacing](https://getbootstrap.com/docs/4.0/utilities/spacing/) now, there are even fewer reasons to write your own css than before.

 e.g. if you change `login-page.component.html` to the following, you're 90% there:
```html
<div class="container">
  <div class="text-center mt-5 row justify-content-md-center">
    <div class="col-sm-6 col-md-4">
      <img class="mb-4" src="assets/images/dspace-logo.png" alt="" width="72" height="72"> 
      <h1 class="h3 mb-0 font-weight-normal">{{"login.form.header" | translate}}</h1>
      <ds-log-in></ds-log-in>
    </div>
  </div>
</div>
```
All this does is add basic grid classes and a few extra divs to put them in.

If you do end up having to write css, try to avoid hardcoded pixel values when possible. Instead derive what you need from existing bootstrap variables. That makes the end result easier to theme, and pre-existing bootstrap themes will automatically work with your components as well. 

The one exception perhaps in the example above are the dimensions of the dspace logo. I'd move those to css, and I wouldn't mind using pixel values for it, though I'd perhaps put them in variables of our own, so that if we need a similar logo later, we can reuse them.

src/app/shared/services/platform.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Inject, Injectable, PLATFORM_ID } from '@angular/core'
+import { isPlatformBrowser, isPlatformServer } from '@angular/common'
+
+export interface IPlatformService {
+  readonly isBrowser: boolean
+  readonly isServer: boolean
+}
+
+@Injectable()
+export class PlatformService implements IPlatformService {
```
artlowel: Are you sure we need this service? I'd prefer not to use isBrowser/isServer checks whenever possible. 

There are separate main modules for browsers and server, in most cases it should be possible to use a different implementation of a service for server or browser, instead of explicitly checking whether you're running on the server or on the browser, which kind of defeats the purpose of Angular Universal 

For an example: see the TransferStateService for [browser](https://github.com/DSpace/dspace-angular/blob/4fbceab1b941f491c1d88e148a72f5608def00dc/src/modules/transfer-state/dspace-browser-transfer-state.service.ts) and [server](https://github.com/DSpace/dspace-angular/blob/4fbceab1b941f491c1d88e148a72f5608def00dc/src/modules/transfer-state/dspace-server-transfer-state.service.ts).

artlowel:
I've also tried it on our internal rest server now for which I do have the credentials. but I still get an issue when logging in. Likely related to this: https://github.com/DSpace/DSpace/pull/1959

Also, would it be possible to have the login link in the header to redirect to the login page if you have javascript disabled? currently the link does nothing without js

src/app/+home-page/home-page.component.html
atarix83: yes, I am going to remove it

src/app/core/dspace-rest-v2/dspace-rest-v2.service.ts
```diff
@@ -45,9 +56,18 @@ export class DSpaceRESTv2Service {
...
-  request(method: RestRequestMethod, url: string, body?: any): Observable<DSpaceRESTV2Response> {
-    return this.http.request(method, url, { body, observe: 'response' })
-      .map((res) => ({ payload: res.body, statusCode: res.statusText }))
+  request(method: RestRequestMethod, url: string, body?: any, options?: HttpOptions): Observable<DSpaceRESTV2Response> {
+    const requestOptions: HttpOptions = {};
+    requestOptions.body = body;
+    requestOptions.observe = 'response';
+    if (options && options.headers) {
+      requestOptions.headers = Object.assign(new HttpHeaders(),  options.headers);
+    }
```
tomdesair: Regarding the "disable ip check" problem. Based on this StackOverflow answer, it should be possible in the Angular app to check if you're running in a browser or in server-side rendering mode: https://stackoverflow.com/questions/44369945/get-remote-address-in-angular-universal#answer-45117793. Using an custom provider, you should also be able to retrieve the original HTTP request received by the express.js server (see the same StackOverflow answer). 

So can't we add some logic here like `if (isPlatformServer(this.platformId))` is true, then add an `X-Forwarded-For` header with the original client ip address from the Express request?

tomdesair: (Note that I still need to learn a lot about Angular so I don't know if this is the best approach)

artlowel: That should be doable, but the same comment I made for platform.service.ts applies here as well: It would be better if we have a separate implementation for server and browser and load the correct version in the platform's main module file, instead of writing if server -> do this, if browser -> do that

atarix83:
@artlowel @tomdesair 
PR can be reviewed again

here what I've changed : 
* bugs fixes
* adjusted use of custom CSS
* removed use of platform service and created different service for `browser` and `server`
* added the possibility on SSR to send the original client IP to avoid "IP check problem"

atarix83:
After adding tests, covarage is ok.
There an error during build with node v8 but I can't understand where is the problem, because I don't have it when run build locally. [Here](https://travis-ci.org/DSpace/dspace-angular/jobs/380108624) the build report

src/app/core/auth/auth.actions.ts
```diff
@@ -0,0 +1,347 @@
...
+  }
+}
+
+/* tslint:enable:max-classes-per-file */
+
+/**
+ * Actions type.
+ * @type {AuthActions}
+ */
+export type AuthActions
```
artlowel: `Should be `AuthAction` instead of `AuthActions`

src/app/core/data/request.service.ts
```diff
@@ -66,9 +67,10 @@ export class RequestService {
       .flatMap((uuid: string) => this.getByUUID(uuid));
   }
 
-  configure<T extends CacheableObject>(request: RestRequest): void {
-    if (request.method !== RestRequestMethod.Get || !this.isCachedOrPending(request)) {
-      this.dispatchRequest(request);
+  // TODO to review "overrideRequest" param when https://github.com/DSpace/dspace-angular/issues/217 will be fixed
```
artlowel: I'd pull the `trackRequestsOnTheirWayToTheStore` if out of `dispatchRequest`, rename `overrideRequest` to `forceBypassCache` to make it clearer what it does, and add the `ResponseCacheRemoveAction` you proposed in #217:
 
```typescript 
configure<T extends CacheableObject>(request: RestRequest, forceBypassCache: boolean = false): void {
  const isGetRequest = request.method === RestRequestMethod.Get;
  if (forceBypassCache || !isGetRequest || !this.isCachedOrPending(request)) {
    this.dispatchRequest(request);
    if (isGetRequest && !forceBypassCache) {
      this.trackRequestsOnTheirWayToTheStore(request);
    }
  }
}
private dispatchRequest(request: RestRequest) {
  this.store.dispatch(new ResponseCacheRemoveAction(request.href));
  this.store.dispatch(new RequestConfigureAction(request));
  this.store.dispatch(new RequestExecuteAction(request.uuid));
}
```

The `ResponseCacheRemoveAction` will likely make it unnecessary for you to use `forceBypassCache` in this PR, but I still think it's a good option to have. We may need it in the future, so I'd leave it in.

src/app/shared/host-window.service.ts
```diff
@@ -92,4 +92,10 @@ export class HostWindowService {
       distinctUntilChanged()
     );
   }
+
+  isMobileView(): Observable<boolean> {
```
artlowel: The fact that `width < MD_MIN` doesn't really say anything about it being mobile or not. All it does it check wether isXs() or isSm() is true. So I'd call it `isXsOrSm()`, and if it's only needed in `auth-nav-menu.component` I'd move it there

src/app/shared/log-in/log-in.component.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../../styles/variables.scss';
+
+.form-login .form-control:focus {
+  z-index: 2;
+}
+.form-login input[type="email"] {
+  margin-bottom: -1px;
+  border-bottom-right-radius: 0;
```
artlowel: Can be achieved using [the `rounded-bottom` bootstrap class](https://getbootstrap.com/docs/4.0/utilities/borders/#border-radius)

src/app/shared/log-in/log-in.component.scss
```diff
@@ -0,0 +1,15 @@
...
+  border-bottom-left-radius: 0;
+}
+.form-login input[type="password"] {
+  border-top-left-radius: 0;
```
artlowel: Can be achieved using [the `rounded-top` bootstrap class](https://getbootstrap.com/docs/4.0/utilities/borders/#border-radius)

src/app/shared/services/platform.service.ts
artlowel: It doesn't look like this is used anymore (which is good, thank you 🙂). Can you remove the service? Otherwise others may be tempted to use it.

src/app/core/data/request.service.ts
atarix83: unluckily this cause some problem with the MyDSpace, because the same rest call are made several times. I've found a solution that we don't need to include in this PR

src/app/shared/host-window.service.ts
atarix83: something similar it's used [here](https://github.com/4Science/dspace-angular/blob/43380732def7a92845fa36a0b4b4d95e16a8caa5/src/app/%2Bsearch-page/search-page.component.ts#L55)  and [here](https://github.com/4Science/dspace-angular/blob/43380732def7a92845fa36a0b4b4d95e16a8caa5/src/app/%2Bsearch-page/search-sidebar/search-sidebar.service.ts#L18). Should we use only this method in the service after renaming it?

artlowel: That'd be great, yes

atarix83:
@artlowel 
now travis can build, but can't run test, and these errors occur only with travis. Can you reproduce them on your environment?

other minor comments should be fixed now

tomdesair:
Does this PR still requires you to "disable ip check" in the REST backend? I cannot really find the part where you set the `X-Forwarded-For` header for server-side rendered responses.

atarix83:
@tomdesair the part that add `X-Forwarded-For` header should be implemented, but i left a comment on how to retrieve client IP here : 
https://github.com/4Science/dspace-angular/blob/66295d10ed62fc2cab3ecf35d1ab5a37917d28eb/src/app/core/auth/server-auth.service.ts#L32

atarix83:
@artlowel  @tdonohue 

I've tried to run travis-ci locally and build succeed.
than I've activated travis-ci on 4science repository and tried to build `authentication` branch and it's finished successfully as you can see here https://travis-ci.org/4Science/dspace-angular/builds/389181417

At this point I think that a new PR should be resolve the problem, what do you think?

artlowel:
Yes, if the exact same commit succeeds, with the exact same travis config in the 4science repo, I really don't see what else we can try apart from closing this PR and recreating it.

PR #241, opened 2018-03-23
Added support for Google Analytics

### LINKED ISSUE #240: Add support for google analytics
Add support for google analytics using [angulartics2](https://angulartics.github.io/angulartics2/)

[Add it to the project](https://github.com/DSpace/dspace-angular#managing-dependencies-via-yarn), and configure it to work with google analytics. Each route change should send a pageview event to GA.

It should only be used by browsers, not by the server. So add it to BrowserAppModule 

Put the google analytics tracking id in the config. If that id isn't configured, no tracking should happen.

Atmire-Kristof:
This PR closes #240 

The tracking ID is stored in the config under the name 'gaTrackingId'.
When this field is empty or doesn't exist, there will be no tracking.
Because the tracking id has to be loaded dynamically, the script tag is added in main.browser.ts.
Because ServerAppModule uses AppComponent and AppComponent uses the Angulartics2GoogleAnalytics service, I created a mock class (named AngularticsMock) for this service to provide in ServerAppModule.
This mock is also used in the AppComponent tests.

A useful tool for testing Google Analytics (chrome extension):
https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna

This tool will print everything google analytics is trying to send in your developer console.
Leaving the tracking id empty, should show nothing with this tool active.
With the tracking id filled in, it should show a pageview event being sent on every route change.

### DIFF
--- a/config/environment.default.js
+++ b/config/environment.default.js
@@ -27,6 +27,8 @@ module.exports = {
     async: true,
     time: false
   },
+  // Google Analytics tracking id
+  gaTrackingId: '',
   // Log directory
   logDirectory: '.',
   // NOTE: will log all redux actions and transfers in console
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "@ngx-translate/core": "9.1.1",
     "@ngx-translate/http-loader": "2.0.1",
     "angular-idle-preload": "2.0.4",
+    "angulartics2": "^5.2.0",
     "body-parser": "1.18.2",
     "bootstrap": "^4.0.0",
     "cerialize": "0.1.18",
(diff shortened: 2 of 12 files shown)


artlowel:
>However, I'll note that this seems to just be the first step to Google Analytics. Currently, this works well for any pageview tasks (i.e. route changes). But, it doesn't work for a file (bitstream) download (as that's a request to the REST API and not a new route).

You're right. I was under the impression that download events were still a custom option, but [according to the wiki](https://wiki.duraspace.org/display/DSDOC6x/DSpace+Google+Analytics+Statistics) it has been part of the codebase since 5.0. We'll discuss on whether to send them from the client or the server side in today's meeting.

As for the docs, I'm not opposed to creating a documentation folder with a few markdown files, but we could also consider putting them on the wiki. On github the docs are automatically tied to the codebase at a certain commit (while on the wiki it may be harder to discern for which version of the code the docs were written), and it allows us ask for a documentation update in the context of a PR, but changing docs on the wiki has less overhead (no PR necessary) and the docs will likely need to end up on the wiki eventually. Which do you prefer?

Issue #160, opened 2017-10-03
Notifications


wwelling:
Create an injectable service that can push a notification to a component. 

- have bootstrap’s statusses: info, warning, error, etc.
- have the option to be closable, either manually or on a timer (an ngrx effect I’d say)
- The creator of a notification should also be able to close it (e.g. when the reason for an error no longer exists)
- As came up in last meeting: there should be a way to add other interactive elements in a message (e.g. a button that opens something) so we’d have the most freedom if notifications could contain any other type of component, but you’d have to look at the feasibility.
- It would be nice if they animated, nothing too flashy, just to make it clear to the user that something changed. The way atom does it for example: https://github.com/atom/notifications

- use redux for the state of notifications

wwelling:
@artlowel I copied the bulleted list you messaged me in slack for this. Also, added that the notifications state should be managed by redux.

PR #243, opened 2018-04-12
Notifications

atarix83:
This PR implements #160 

This implementation provide a notification service that allow to launch four different type of notification : 
-   Success
-   Error
-   Info
-   Warning

Styles are according to the bootstrap `.alert-*` classes

The notification's message can have both title and content or just one of these, and the relative parameters should be string or Observable of string.

By default a notification disappears after 5 seconds, but it's possible to change this behavior and other parameters, through the [`NotificationOptions`](https://github.com/4Science/dspace-angular/blob/notifications/src/app/shared/notifications/models/notification-options.model.ts) object.

This is a simple example of how to launch a new notification : 
```
this.notificationsService.success( this.translate.get('path.to.title'), this.translate.get('path.to.message'));
```


artlowel:
Looks good, thanks.

A few comments:

- It would be nice if the default notification options would be in the environment config files. I can imagine that some institutions would like to increase the default time-out for example
- Please try to use or derive as many scss px values and color codes from the bootstrap variables. E.g. things like this [padding](https://github.com/DSpace/dspace-angular/pull/243/files#diff-d03b5a580f14b87f86e39eb3647f5bcaR16) could use [bootstrap's variable](https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L96), [these color codes](https://github.com/DSpace/dspace-angular/pull/243/files#diff-d03b5a580f14b87f86e39eb3647f5bcaR37) could likely straight up use or be derived using the darken and adjust-hue functions from [bootstrap's state variables](https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L504), I'd rather not [introduce custom media break points](https://github.com/DSpace/dspace-angular/pull/243/files#diff-b106075a246d8676b3b9e26bdc1233d7R37), but stick to [bootstrap's predefined ones](https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L292), etc. If we consistently do this, not only will the end result look more consistent (because not every dev is choosing their own similar, but not quite the same px sizes for everything) but it will also be much easier to theme.

artlowel:
Thanks

One more comment though: wouldn't it make more sense to have [the `content` field](https://github.com/DSpace/dspace-angular/pull/243/files#diff-83f8cc447dc2c3532f68d2313020fd4bR10) always contain the content, and to turn [the `html` field](https://github.com/DSpace/dspace-angular/pull/243/files#diff-83f8cc447dc2c3532f68d2313020fd4bR33) in to a boolean you set to true when `content` contains html?

atarix83:
yes, it is reasonable, I try to modify it this way

atarix83:
@artlowel  

done

artlowel:
Thanks. 

The  state of the `html` boolean doesn't seem change anything though:

If I specify `'<b>test</b>'` in the content field I would expect it to be rendered as `<b>test</b>` when HTML is enabled and as **`test`** when it is disabled. But in both cases it is rendered as **`test`**.

In other words, when `html` is `true` I'd expect it to be put as 
`<div [innerHtml]="(content | async)">` in the template. When it is `false` I would expect `<div>{{content | async}}</div>`. 

This is important for security as well: as you'd only use the HTML option if you're absolutely certain of the source and quality of the content. When you're not (and I can imagine that these notifications will in many cases be used to simply display an error message that comes from a third party), you'd want to escape anything to be sure it doesn't break the formatting of the page, or run unauthorized scripts.

atarix83:
@artlowel 

Corrected it

PR #264, opened 2018-05-10
Dynamic forms Components

atarix83:
This PR includes components and services that provide to create an input form UI dynamically.

To achieve this it's been used and extended this [third party library](https://github.com/udos86/ng-dynamic-forms). 
The way in which the library has been extended can be significantly improved, after making an upgrade of the library, that allow to easily create [custom form controls](https://github.com/udos86/ng-dynamic-forms#custom-form-controls). This refactoring can be done after an upgrade of our angular version (at least v6.0.0)

Here a gist that show how the new `FormComponent` can be used (this should help to review the PR) : 
https://gist.github.com/atarix83/fba1be54b075f2e2d424cc79e5a0aaad


src/app/shared/uploader/uploader.component.scss
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.ds-document-drop-zone {
+  top: 0;
+  left: 0;
+  z-index: -1;
+}
+
+.ds-document-drop-zone-active {
+  z-index: 1025 !important;
```
artlowel: If these z-indexes can't use one of bootstrap's standard values, best move them to `_custom_variables.scss` so we can get an overview of all custom z-indexes in dspace, and prevent unexpected conflicts.

src/app/shared/form/form.reducers.ts
```diff
@@ -0,0 +1,211 @@
+import {
```
artlowel: Please call reducer files `.reducer.ts` for consistency

src/app/core/integration/integration-data.ts
```diff
@@ -0,0 +1,12 @@
+import { PageInfo } from '../shared/page-info.model';
+import { IntegrationModel } from './models/integration.model';
+
+/**
+ * A class to represent the data retrieved by a Eperson service
```
artlowel: An EPerson service? 
Could you give another example, besides an authority value, of something that could be an integration (in the future perhaps)? Because it isn't entirely clear to me what they're supposed to represent.

src/app/shared/chips/chips.component.ts
```diff
@@ -0,0 +1,89 @@
...
+import { ChipsItem } from './models/chips-item.model';
+import { UploaderService } from '../uploader/uploader.service';
+
+@Component({
+  selector: 'ds-chips',
+  styleUrls: ['./chips.component.scss'],
+  templateUrl: './chips.component.html',
+})
+
+export class ChipsComponent implements OnChanges {
```
artlowel: It's also not clear to me what Chips are in this context.

src/app/core/data/config-response-parsing.service.ts
```diff
@@ -27,7 +27,7 @@ export class ConfigResponseParsingService extends BaseResponseParsingService imp
   }
 
   parse(request: RestRequest, data: DSpaceRESTV2Response): RestResponse {
-    if (isNotEmpty(data.payload) && isNotEmpty(data.payload._links) && data.statusCode === '200') {
```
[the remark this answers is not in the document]
atarix83: statuscode  is `OK` only on SSR that's why I added it

src/app/core/integration/integration-data.ts
atarix83: I supposed to create an abstract service like the `ConfigService`.
If we think that there will not be other integration endpoints I remove it and use only `AuthorityService`

src/app/shared/chips/chips.component.ts
atarix83: `ChipsComponent` is used to render and handle (add/remove/sort) chip elements and can be used in a form :
chips

src/app/shared/chips/models/chpis-item.model.spec.ts
```diff
@@ -0,0 +1,78 @@
+import { ChipsItem, ChipsItemIcon } from './chips-item.model';
```
tdonohue: This Spec file has a misspelling in the file name `chpis-item.model.spec.ts`

src/app/shared/chips/models/chpis.model.spec.ts
```diff
@@ -0,0 +1,126 @@
+import { Chips } from './chips.model';
```
tdonohue: This Spec file has a misspelling in the file name `chpis.model.spec.ts`

src/app/shared/chips/models/chpis-item.model.spec.ts
atarix83: fixed

src/app/shared/chips/models/chpis.model.spec.ts
atarix83: fixed

src/app/shared/chips/chips.component.ts
tdonohue: @atarix83 : this is something that comments in the component could help with.  Perhaps we need to describe what a "Chip" element is, possibly even with links for examples.  I admit, I didn't understand the work "chip" here either. But, I see it's a common term: https://www.w3schools.com/howto/howto_css_contact_chips.asp

PR #277, opened 2018-07-10
Fix truncated text being difficult to select

MatMercer:
Related to Google Chrome browser and the issue #234, this minor fix works since the `::after` pseudo element blocks elements behind it. Adding `pointer-events: none` to the `::after` pseudo element fixed the issue.


artlowel:
This PR closes #234

PR #304, opened 2018-10-11
Working language dropdown

bram-atmire:
First version of a working language switch component. Styling is not 100% ok yet, but I understand there are still changes in progress to the nav styling anyway?

Let me know what you think!

### DIFF
--- a/src/app/header/header.component.ts
+++ b/src/app/header/header.component.ts
@@ -7,6 +7,7 @@ import { HeaderState } from './header.reducer';
 import { HeaderToggleAction } from './header.actions';
 import { AppState } from '../app.reducer';
 import { HostWindowService } from '../shared/host-window.service';
+import {TranslateService} from "@ngx-translate/core";
 
 const headerStateSelector = (state: AppState) => state.header;
 const navCollapsedSelector = createSelector(headerStateSelector, (header: HeaderState) => header.navCollapsed);
@@ -27,7 +28,8 @@ export class HeaderComponent implements OnInit {
 
   constructor(
     private store: Store<AppState>,
-    private windowService: HostWindowService
+    private windowService: HostWindowService,
+    public translate: TranslateService
   ) {
   }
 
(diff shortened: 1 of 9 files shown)


src/app/header/header.component.ts
```diff
@@ -7,6 +7,7 @@ import { HeaderState } from './header.reducer';
 import { HeaderToggleAction } from './header.actions';
 import { AppState } from '../app.reducer';
 import { HostWindowService } from '../shared/host-window.service';
+import {TranslateService} from "@ngx-translate/core";
```
bram-atmire: don't think this is needed here anymore

src/app/header/header.component.ts
```diff
@@ -27,7 +28,8 @@ export class HeaderComponent implements OnInit {
 
   constructor(
     private store: Store<AppState>,
-    private windowService: HostWindowService
+    private windowService: HostWindowService,
+    public translate: TranslateService
```
bram-atmire: don't think this is needed here anymore

cjuergen: In the list of languages to display I would use the labels in their respective language. There is no mechanism in ngx-translate for this kind of having a locale and translate a certain part with this locale rather than the current one, so I've done:
https://github.com/cjuergen/dspace-angular/blob/language-switch/resources/i18n/en.json#L277
Looks a bit odd, but there is no way to put comments in json.

cjuergen:
In the list of languages to display I would use the labels in their respective language. There is no mechanism in ngx-translate for this kind of having a locale and translate a certain part with this locale rather than the current one, so I've done:
https://github.com/cjuergen/dspace-angular/blob/language-switch/resources/i18n/en.json#L277
Looks a bit odd, but there is no way to put comments in json.

cjuergen:
The default languages and supported languages should be configurable and the switch only be displayed, if there is more than one supported language, something like
https://github.com/cjuergen/dspace-angular/blob/language-switch/src/app/shared/language-switch/language-switch.component.html#L1
https://github.com/cjuergen/dspace-angular/blob/language-switch/src/app/shared/language-switch/language-switch.component.ts#L13

bram-atmire:
awesome suggestions, thank you very much @cjuergen

bram-atmire:
closing this again, way to many things that still need to be fixed before this is ready. jumped the gun a bit to quickly.

PR #307, opened 2018-10-12
#293 Added link with the query quoted for search results if empty

### LINKED ISSUE #293: Show hints when a query returns no results
When a search returns no results, or the server returns an error, it's possible the reason is that the query contained one or more characters or words that were interpreted by solr as query syntax. We want to advise users to put their query between quotes and try again. 

When a search query returns an error with statuscode 400 (bad request), or no results, display the following text instead of the results section:

>Your search returned no results.
>
>Having trouble finding what you're looking for? 
>Try putting [quotes around your query.](#)

And turn the text "quotes around your query" in to a link to a search for their query with quotes around it. Any active filters and sorting options should stay active.

christian-scheible:
If there is an error 400 or a search returns an empty result this commit will show a link with the query quoted. 
Also improved testing of search results component as the component has changed.

This PR closes #293.

### DIFF
--- a/src/app/+search-page/search-results/search-results.component.spec.ts
+++ b/src/app/+search-page/search-results/search-results.component.spec.ts
@@ -1,11 +1,12 @@
 import { ComponentFixture, TestBed, async, tick, fakeAsync } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
-import { DebugElement, NO_ERRORS_SCHEMA } from '@angular/core';
+import { NO_ERRORS_SCHEMA } from '@angular/core';
 import { ResourceType } from '../../core/shared/resource-type';
 import { Community } from '../../core/shared/community.model';
 import { TranslateModule } from '@ngx-translate/core';
 import { SearchResultsComponent } from './search-results.component';
+import { QueryParamsDirectiveStub } from '../../shared/testing/query-params-directive-stub';
 
 describe('SearchResultsComponent', () => {
   let comp: SearchResultsComponent;
@@ -14,7 +15,9 @@ describe('SearchResultsComponent', () => {
   beforeEach(async(() => {
     TestBed.configureTestingModule({
       imports: [TranslateModule.forRoot(), NoopAnimationsModule],
-      declarations: [SearchResultsComponent],
+      declarations: [
+        SearchResultsComponent,
+        QueryParamsDirectiveStub],
       schemas: [NO_ERRORS_SCHEMA]
     }).compileComponents();
   }));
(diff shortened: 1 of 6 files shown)


src/app/+search-page/search-results/search-results.component.spec.ts
```diff
@@ -9,32 +10,50 @@ import { SearchResultsComponent } from './search-results.component';
...
+  it('should display error message if error is != 400', () => {
+    (comp as any).searchResults = { hasFailed: true, error: { statusCode: 500 } };
+    fixture.detectChanges();
+    expect(fixture.debugElement.query(By.css('ds-error'))).not.toBeNull();
+  });
+
+  it('should display link with new search where query is quoted if search return a error 400', () => {
+    (comp as any).searchResults = { hasFailed: true, error: { statusCode: 400 } };
+    fixture.detectChanges();
+    expect(fixture.debugElement.query(By.css('a'))).not.toBeNull();
```
artlowel: This doesn't test what the link does, only that there is a link. 

Ideally you'd test: 
- that the link you navigate to contains the same query but in quotes
- that any other params are still there, and haven't been modified.

[This section of the angular.io testing guide](https://angular.io/guide/testing#components-with-routerlink) can help with that

src/app/+search-page/search-results/search-results.component.spec.ts
```diff
@@ -9,32 +10,50 @@ import { SearchResultsComponent } from './search-results.component';
...
+  });
 
-  it('should not display heading when results is empty', () => {
-    expect(heading).toBeNull();
+  it('should display link with new search where query is quoted if search result is empty', () => {
+    (comp as any).searchResults = { payload: { page: { length: 0 } } };
+    fixture.detectChanges();
+    expect(fixture.debugElement.query(By.css('a'))).not.toBeNull();
```
artlowel: Same comment as above

src/app/+search-page/search-results/search-results.component.ts
```diff
@@ -35,4 +36,16 @@ export class SearchResultsComponent {
    */
   @Input() viewMode: ViewMode;
 
+  /**
+   * Method to change the given string by surrounding it by quotes if not already present.
+   */
+  surroundStringWithQuotes(input: string): string {
```
artlowel: I'd also write a test for this method on its own.

christian-scheible:
Hi @artlowel, sorry for the late response. I was ill but I will look into this.

src/app/+search-page/search-results/search-results.component.spec.ts
christian-scheible: I solved the first part of your suggestion. But I am not sure about how to test

> that any other params are still there, and haven't been modified.

I tried providing a mocked ActivatedRoute, with some params set, to TestBed. But I am not sure how I can test how the resulting link will look like as the ComponentFixture contains only the router link before merging the parameters. 

And I am not sure if this test is really necessary as it is just testing if Angulars queryParamsHandling is working as expected. Or am I missing something here?

artlowel:
Thanks.

Your tests look good. But travis still fails, due to this issue: https://github.com/angular/angular/issues/13590.

TL;DR: The AoT compiler tries to compile your QueryParamsDirectiveStub even though it is only used for testing, and throws an error when it sees it isn't used in a module.

To remedy this I suggest you add a bogus module to the `/app/shared/testing` folder:
```typescript
import { NgModule } from '@angular/core';
import { QueryParamsDirectiveStub } from './query-params-directive-stub';

/**
 * This module isn't used. It serves to prevent the AoT compiler
 * complaining about components/pipes/directives that were 
 * created only for use in tests.
 * See https://github.com/angular/angular/issues/13590
 */
@NgModule({
  declarations: [
    QueryParamsDirectiveStub
  ]
})
export class TestModule {}
```

christian-scheible:
Sorry but I missed the failing Travis build. I added the bogus model as suggested. The Travis checks do now pass.

PR #344, opened 2018-12-19
Made home news responsive

courtneypattison:
- Set width and height of logo to prevent page reflow
- Resized logo to match width and height specified
- Used bootstrap for margins
- Set flex-wrap so h1 text stays on screen for small devices


src/app/+home-page/home-news/home-news.component.html
```diff
@@ -1,9 +1,7 @@
 <div class="jumbotron jumbotron-fluid">
   <div class="container">
-    <div class="d-flex">
-      <div class="dspace-logo-container">
-        <img src="assets/images/dspace-logo.png" />
-      </div>
+    <div class="d-flex flex-wrap">
+      <img class="mr-4 mt-3"  src="assets/images/dspace-logo.png" width="110px" height="110px" />
```
artlowel: Is there a reason to put these dimensions in html? It seems cleaner to me to keep them in scss.

courtneypattison: I added the svg and I forgot to put an alt in last time, so added that as well. The dimensions are in the html to stop the page from reflowing when the image loads. I changed to just specifying the width because the only reflow that happens is the text to the right of the logo.

artlowel: It will reflow in the same way if you give the image a css class, and put the width in the scss file. I'd rather have all style together in the scss files, so that if you want to change something later on you know exactly where to find it. If we keep it in the html and someone were to try and set a width from css, they'd notice it wouldn't work, as the in-html width would get priority.

courtneypattison: Sorry, my goof! For some reason I thought it had to be specified in the html. Hopefully it's better now.

PR #352, opened 2019-01-16
[Configurable Entities] Rename 'entity' to 'item'

artlowel:
This PR renames all instances of the word "entity" to "item" as discussed in #315


resources/i18n/en.json
```diff
@@ -207,7 +207,7 @@
         "f.dateIssued.max": "End date",
         "f.subject": "Subject",
         "f.has_content_in_original_bundle": "Has files",
-        "f.entityType": "Entity Type"
+        "f.entityType": "Item Type"
```
artlowel: Note that this `"f.entityType"` key comes from the server, so it can't change until there's a REST PR that fixes it.

artlowel: I created #353 to track those

PR #364, opened 2019-02-21
Browse by date and Starts-With component

### LINKED ISSUE #248: StartsWith Component
Create a component that allows a user to traverse a list by selecting a letter of the alphabet, or entering a few letters the results should start with.

See [the current browse by title page](http://demo.dspace.org/xmlui/browse?rpp=20&sort_by=1&type=title&etal=-1&starts_with=A&order=ASC) for an example

If the user clicks a link the component should update the URL to reflect the change. 

If the user enters something in the "first few letters" box, and clicks "go" the component should update the URL to reflect the change. 

Both clicking a link and entering something in the box and clicking "go" should work with javascript disabled. So use routerlinks, and a form submit

It would be nice if after a debounce of a few hundred ms, the URL would update automatically when the user enters a character in the "first few letters" input for people that do have JS enabled.

The parent component using this startwith component is responsible for watching the url change, and updating the rest call accordingly.

Add a `startsWith: string` property to the PageInfo model, and take one of those models as an `@Input`for your StartsWith component. If the input `pageInfo` has a `startsWith` value, show it in the "first few letters" box.

Atmire-Kristof:
This PR closes #248 and #254

The purpose of this PR is to add a Browse-By page for date fields as well as adding a Starts-With component to all Browse-By pages.

### Browse-By-Date-Page
Currently, the only date field using this component is "dc.date.issued" a.k.a. "dateissued". The route to this page can be found at `[base-url]/browse/dateissued`.
This is a generic component extending from BrowseByMetadataPage and can be re-used for other date fields in the future.

### Starts-With
Two new components have been added: StartsWithText and StartsWithDate, both extending from an abstract component StartsWithAbstract.
- StartsWithText renders the alphabetic starts-with as we know it from the browse-by pages in DSpace6
- StartsWithDate renders the starts-with containing a dropdown with years and input field as we know it from the browse-by pages in DSpace6

These components come with a decorator `starts-with-decorator.ts` to switch components depending on their "StartsWithType", currently being either Text or Date.
This decorator is then used in `browse-by.component.ts` to switch to the correct StartsWith component depending on the type provided.

The StartsWith component adds a `startsWith` parameter to the url to limit your search on the Browse-By pages.


src/app/core/browse/browse.service.ts
```diff
@@ -103,42 +109,26 @@ export class BrowseService {
...
    * @param {string} filterValue      metadata value to filter by (e.g. author's name)
-   * @param options                   Options to narrow down your search:
-   *                                  { pagination: PaginationComponentOptions,
-   *                                    sort: SortOptions }
+   * @param options                   Options to narrow down your search
    * @returns {Observable<RemoteData<PaginatedList<Item>>>}
    */
   getBrowseItemsFor(filterValue: string, options: BrowseEntrySearchOptions): Observable<RemoteData<PaginatedList<Item>>> {
-    const request$ = this.getBrowseDefinitions().pipe(
+    return this.getBrowseDefinitions().pipe(
```
paulo-graca: I'm getting an `IllegalArgumentException` on the rest side because scope isn't defined. Scope has the value "undefined". I think the problem is related with this part of the code:
```
        if (isNotEmpty(options.sort)) {
          args.push(`scope=${options.scope}`);
        }
```

The requested URL: 
`http://localhost/spring-rest/api/discover/browses/dateissued/items?scope=undefined&sort=default,ASC&page=NaN&size=undefined&startsWith=2017`

The logged error on DSpace (java side):
```
2019-02-22 14:07:32,766 WARN  org.dspace.app.rest.utils.ScopeResolver @ The given scope string undefined is not a UUID
java.lang.IllegalArgumentException: Invalid UUID string: undefined
```

src/app/core/browse/browse.service.ts
```diff
@@ -155,6 +145,9 @@ export class BrowseService {
           args.push(`page=${options.pagination.currentPage - 1}`);
           args.push(`size=${options.pagination.pageSize}`);
         }
+        if (isNotEmpty(options.startsWith)) {
+          args.push(`startsWith=${options.startsWith}`);
```
paulo-graca: I'm getting this strange behavior of having different rows vs results in facets. You can reproduce it by following this steps and use the Atmire's database - https://www.dropbox.com/s/ovqp394y3vofnwa/entities7-test-db.sql.gz?dl=1:

Access the browse by subject page:
`/browse/subject`
Filter the results by letter F:
`/browse/subject?startsWith=F`
Choose: Financial Therapy that has 8 results
`/browse/subject?startsWith=F&value=financial%20therapy`
You will only see 4 records but financial therapy has 8. This is because the parameter "startsWith=F" is assumed and it's used along with the filter subject. I think the parameter shouldn't be set on this request.
I excepted to see the same results as:
`/browse/subject?value=financial%20therapy`

src/app/+browse-by/+browse-by-metadata-page/browse-by-metadata-page.component.ts
```diff
@@ -104,6 +123,11 @@ export class BrowseByMetadataPageComponent implements OnInit {
           }
           this.updateParent(params.scope);
         }));
+    this.updateStartsWithTextOptions();
+  }
+
+  updateStartsWithTextOptions() {
```
paulo-graca: Please add comments to describe this function

src/app/shared/starts-with/date/starts-with-date.component.html
```diff
@@ -0,0 +1,25 @@
+<form class="w-100" [formGroup]="formData" (ngSubmit)="submitForm(formData.value)">
+  <div class="row">
+    <span class="col-12 font-weight-bold mb-1">
+      {{'browse.startsWith.jump' | translate}}
+    </span>
+    <div class="col-5 col-md-3">
+      <select class="form-control" (change)="setStartsWithEvent($event)">
+        <option [value]="-1" [selected]="!startsWith">
+          {{'browse.startsWith.choose_year' | translate}}
+        </option>
```
paulo-graca: Please also consider the support for the month as in JSPUI (http://demo.dspace.org/jspui/browse?type=dateissued&sort_by=2&order=ASC&rpp=20&etal=-1&year=2019&month=1&starts_with=).

src/app/core/browse/browse.service.ts
Atmire-Kristof: The issue should be fixed now. I simply forgot to change the "sort" property in the check to "scope". Nice catch!

src/app/core/browse/browse.service.ts
Atmire-Kristof: The startsWith parameter is not included in the links anymore when navigating browse entries.

src/app/+browse-by/+browse-by-metadata-page/browse-by-metadata-page.component.ts
Atmire-Kristof: Added comments.

src/app/shared/starts-with/date/starts-with-date.component.html
Atmire-Kristof: Added a dropdown to support adding a month to the startsWith parameter.

src/app/+browse-by/+browse-by-metadata-page/browse-by-metadata-page.component.ts
paulo-graca: 👍

src/config/browse-by-config.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { Config } from './config.interface';
+
+export interface BrowseByConfig extends Config {
```
tdonohue: Could we have some TypeDocs added here, please.  Especially since the `oneYearLimit`, `fiveYearLimit` fields below are not self-explanatory, so it'd be good to describe the purpose of this interface

tdonohue:
@Atmire-Kristof and @artlowel : I've tested this today. While the StartsWith & Browse By Date functionality seems to work fine overall, I've got some concerns/questions about the design approach here.

* First, I've noticed that with the addition of the StartsWith components, all of the Browse By pages no longer return a *count* of total results.  I'm used to seeing the "Now showing items _ of _" and that is no longer displayed.
* Because the "Now showing" is missing, the behavior of the StartsWith components is very odd initially, until you understand what it is doing.  It actually took me a moment to realize that saying "StartsWithDate = 2011" means I'll jump to seeing everything that has an Issue Date >= 2011.  I think that's the correct *behavior*, however it's not at all clear to the user what is going on (i.e. this is likely a usability issue, and I think the "Now showing" (or similar text) would help, as it'd be clear you jumped to a new "page").
* Browse by Author / Subject behave *differently* from Browse by Title / Date.  The former seem to be acting as a filter (e.g. if you enter "S" in Browse By Author or Subject, you only see Authors/Subjects starting with "S").  The latter jump you to a page in the list.  This may actually be an issue with the underlying REST API (or Solr index) -- I don't recall offhand.

artlowel:
All of the points you noted are consequences of how the rest side works I'm afraid:

>- First, I've noticed that with the addition of the StartsWith components, all of the Browse By pages no longer return a *count* of total results.  I'm used to seeing the "Now showing items _ of _" and that is no longer displayed.
>- Because the "Now showing" is missing, the behavior of the StartsWith components is very odd initially, until you understand what it is doing.  It actually took me a moment to realize that saying "StartsWithDate = 2011" means I'll jump to seeing everything that has an Issue Date >= 2011.  I think that's the correct *behavior*, however it's not at all clear to the user what is going on (i.e. this is likely a usability issue, and I think the "Now showing" (or similar text) would help, as it'd be clear you jumped to a new "page").

When using the startsWith param, the rest api will return the exact same pagination info for any item that would be on the same page if startsWith wasn't used. So if normally the first Title starting with 'S' would be the fourth item on the 15th page, then the pagination object the rest api returns when you ask for `startsWith=S` will contain `number: 15`, but no indication of the offset that item has within the page. So we don't have, and can't calculate the "now showing" info. This is also the reason why we can't show the more extended pagination control, because all we have to go on are the previous and next hal links.

>- Browse by Author / Subject behave *differently* from Browse by Title / Date.  The former seem to be acting as a filter (e.g. if you enter "S" in Browse By Author or Subject, you **only** see Authors/Subjects starting with "S").  The latter jumps you to a page in the list (so if you enter "S" in Browse by Title, you see Titles that start with S, T, U, V, etc).  This may actually be an issue with the underlying REST API (or Solr index) -- I don't recall offhand. 

Again that's determined by what the rest api returns. I agree that it would be nice to make this more consistent, but this seems to be exactly the way it works in dspace 6 as well:
- [Browse by date](http://demo.dspace.org/xmlui/browse?year=2011) is an offset
- [Browse by author](http://demo.dspace.org/xmlui/browse?rpp=20&etal=-1&sort_by=-1&type=author&starts_with=D&order=ASC) is a filter
- [Browse by title](http://demo.dspace.org/xmlui/browse?rpp=20&etal=-1&sort_by=1&type=title&starts_with=S&order=ASC) is an offset

tdonohue:
@artlowel : thanks for the clarifications here.  I agree that it sounds like the third point is current behavior (so I'll withdraw that). 

However, my first point is a change in behavior from DSpace 6.x.  In DSpace 6.x we are able to display a "Now showing..." option even after jumping to a specific letter.  For example, here's Browse by Title starting with "D" in the XMLUI: http://demo.dspace.org/xmlui/browse?rpp=20&etal=-1&sort_by=1&type=title&starts_with=D&order=ASC

I definitely understand that this PR isn't at fault for that change in behavior.  But, I wonder if we can log this as a known bug and figure out a way to fix this in the REST API (and a future Angular UI PR) -- perhaps you could brainstorm with @benbosman on this?  If we feel it's nearly impossible (or very hard) to retain that "Now showing..." text, then I feel we need some other sort of contextual information (in the UI) that helps the user figure out what this component is doing.  

Currently, as I noted, the behaviour was even confusing to me (a DSpace expert).  For example, in testing this PR, I went to Browse by Title and clicked on letter "F". As I had no titles starting with the letter "F", the first title I saw began with "H".  I initially thought it wasn't working correctly, until I realized that what was going on.  However, realizing what was going on was difficult, cause there was no way to page backwards. I literally had to clear out my search and double check if I had any titles starting with "F".  So, I'm worried this UI currently has some significant usability issues (again, not really a fault of this PR alone).

So, in summary, I'm OK with accepting this PR as-is (since it seems to "work") if we can develop a plan to improve it in the future. I'd like that plan to be documented (as best we can) in tickets.  We could also discuss it further in next week's meeting (as a discussion topic) if necessary.

(As a sidenote, this PR now has minor merge conflicts related to the merger of the Administrative Edit PR.)

tdonohue:
@Atmire-Kristof and @artlowel : After today's meeting we decided to move ahead with this PR as-is, and I'll log a bug about this behavior.

However, I just tried to retest this PR, and I'm now finding that Browse by Title & Author are no longer working.  I'm seeing empty results for both that look like this (those are empty parentheses displayed for each item):
empty-results

I'm guessing that some recent code on `master` must now be conflicting with this PR.  Both Browse by Title & Author seem to work fine on `master`.

artlowel:
@tdonohue You were right, this was an issue due to a merge in master, that didn't show up as a merge conflict in github. Thanks for testing it again before you merged 👍

PR #374, opened 2019-03-21
Search performance optimisations

LotteHofstede:
This PR aims to improve the performance of the search page.
Especially the sidebar facets tended to be rerendered more often than necessary, which caused the whole page to be very slow.


LotteHofstede:
Thanks for your review @atarix83 ! 
I tried reproducing the error you came across, and while I was able to recreate the UI issue, I can't reproduce the error you reported.
However, I did fix the UI so it has the same behaviour as it did before this PR.

Could you check if this now works for you or that you still get this error in your console?

atarix83:
@LotteHofstede 

now the issue with the empty box is fixed.

The error is already present, I suggest this fix that I've applied for mydspace : 
https://github.com/4Science/dspace-angular/blob/c1ccf43d512b516256af37faf998ec057b2a9652/src/app/%2Bsearch-page/search-service/search.service.ts#L137

LotteHofstede:
@atarix83 Could you elaborate on what REST server you used to test this? Because neither on our internal nor on the public 4Science demo server I am able to reproduce the error.

When looking at the MyDSpace code you shared, it seems like you are expecting search results without dspaceObjects. This does not seem like a valid response from the REST server - all search results should contain a DSpace Object - and the solution you implemented will most likely break pagination. If for example you requested 10 items, but you filter out 2, the user will only be presented with 8 results instead of 10.

atarix83:
@LotteHofstede 
it happens because I'm testing over mydspace REST branch https://github.com/4Science/DSpace/tree/DS-4166_mydspace and there REST return a `resultObject` instead of `dspaceObject` as response of `search/objects` endpoint.
However it could be a fallback strategy in case for some reason REST does not respond as expected

atarix83:
@LotteHofstede 

After a deep analysis I found a problem that cannot be found on the "search" but occurred with the "mydspace".
Based on the current implementation the facet values for a filter are retrieve `OnInit` of the `SearchFacetFilterComponent` and updated every time `this.searchConfigService.searchOptions` emit a new new value ([I refer to this](https://github.com/atmire/dspace-angular/blob/Search-optimizations/src/app/%2Bsearch-page/search-filters/search-filter/search-facet-filter/search-facet-filter.component.ts#L100)).
This doesn't create any problems, if the number of filters is always the same.

In the mydspace implementation I've added the `configuration` param to `SearchOptions` model as you can see [here](https://github.com/4Science/dspace-angular/blob/c1ccf43d512b516256af37faf998ec057b2a9652/src/app/%2Bsearch-page/search-options.model.ts#L11). This means that when `configuration` param change a new value for `this.searchConfigService.searchOptions`  is emitted.

For the mydspace the configuration param is changed by a form select

Schermata del 2019-03-26 15-54-59

and could affect the number of the filters retrieved from `api/discover/facets` endpoint.
In mydspace specific case : 
this is the response from `api/discover/facets?configuration=workspace` call 
```
{
    "id": null,
    "scope": null,
    "configuration": "workspace",
    "type": "discover",
    "_links": {
        "self": {
            "href": "http://localhost:8080/spring-rest/api/discover/facets?configuration=workspace"
        }
    },
    "_embedded": {
        "facets": [
            {
                "name": "namedresourcetype",
                "facetType": "text",
                "facetLimit": 10,
                "_links": {
                    "self": {
                        "href": "http://localhost:8080/spring-rest/api/discover/facets/namedresourcetype"
                    }
                }
            },
            {
                "name": "itemtype",
                "facetType": "text",
                "facetLimit": 10,
                "_links": {
                    "self": {
                        "href": "http://localhost:8080/spring-rest/api/discover/facets/itemtype"
                    }
                }
            },
            {
                "name": "dateIssued",
                "facetType": "date",
                "facetLimit": 5,
                "_links": {
                    "self": {
                        "href": "http://localhost:8080/spring-rest/api/discover/facets/dateIssued"
                    }
                }
            }
        ]
    }
}
```
while this is the response from `api/discover/facets?configuration=workflow` call : 
```
{
    "id": null,
    "scope": null,
    "configuration": "workflow",
    "type": "discover",
    "_links": {
        "self": {
            "href": "http://localhost:8080/spring-rest/api/discover/facets?configuration=workflow"
        }
    },
    "_embedded": {
        "facets": [
            {
                "name": "namedresourcetype",
                "facetType": "text",
                "facetLimit": 10,
                "_links": {
                    "self": {
                        "href": "http://localhost:8080/spring-rest/api/discover/facets/namedresourcetype"
                    }
                }
            },
            {
                "name": "itemtype",
                "facetType": "text",
                "facetLimit": 10,
                "_links": {
                    "self": {
                        "href": "http://localhost:8080/spring-rest/api/discover/facets/itemtype"
                    }
                }
            },
            {
                "name": "dateIssued",
                "facetType": "date",
                "facetLimit": 5,
                "_links": {
                    "self": {
                        "href": "http://localhost:8080/spring-rest/api/discover/facets/dateIssued"
                    }
                }
            },
            {
                "name": "submitter",
                "facetType": "text",
                "facetLimit": 10,
                "_links": {
                    "self": {
                        "href": "http://localhost:8080/spring-rest/api/discover/facets/submitter"
                    }
                }
            }
        ]
    }
}
```
As you can see for the `configuration=workflow` there is an extra filter `submitter`.

So when turn from `workflow` to `workspace` configuration, the `SearchFacetFilterComponent` related to `submitter` filter try to update facet values by a call to `/api/discover/facets/submitter?configuration=workspace` that obviously fails.

This happens because `SearchFacetFilterComponent` makes the REST call before `SearchFiltersComponent` has updated its filters [here](https://github.com/atmire/dspace-angular/blob/281a4cd1d7b77ff8c9f51f5cb3e254f497821521/src/app/%2Bsearch-page/search-filters/search-filters.component.ts#L40) and `SearchFacetFilterComponent` related to that filter has destroyed in the view.

Do you think you could provide a solution for this issue in this PR?

LotteHofstede:
Hey @atarix83 
I understand this is inconvenient for you, but I can't fix any issues with PR's that have not even been created yet. 
If you renamed the dspaceObject, you should probably refactor it in the Angular PR that matches the REST PR where it was changed. I cannot possibly create a fall back for every invalid rest response, that's why we have a contract.
The only thing I can do is make sure errors like this are displayed in a notification instead of the console, although I don't think this would help the user in any way.

atarix83:
@LotteHofstede I agree that renamed dspaceObject is not the goal of this PR.

The fallback solution suggested needs only to avoid the search hanging for results, I think that a notification on unexpected error is a better solution

LotteHofstede:
@tdonohue Thanks for the review, I solved the merge conflict

PR #383, opened 2019-04-04
Admin Edit Metadata - bugfix

LotteHofstede:
@Atmire-Kristof discovered a small issue with my previous Edit Metadata PR #362
When a user discards all changes a notification appears and the Discard button turns into an Undo button.

However, when clicking - and therefore closing - the notification, the undo button would stay in place and only turn back into a Discard button after initial the time-out has finished. This PR solves this issue.

After fixing this specific issue, @artlowel also discovered a caching issue with the metadata fields. When the user would go directly to the edit metadata page, already logged in, the code would not be able to correctly retrieve the list of available metadata fields. When the user subsequently tried to edit the field, every metadata field would be rendered as invalid and the auto complete feature would break. I fixed this by making sure the metadata fields are always interpreted as MetadataField objects.


paulo-graca:
The code looks good to me. I will try it out.

PR #389, opened 2019-04-25
Preview release theme, extended for Configurable Entities

LotteHofstede:
This PR is an extension on #376 (and *includes* #376) and is based on the configurable-entities branch.
It aims to add a custom theme for all currently existing types of entity detail pages; such as journal, person and publication pages.

### Example page of a journal:
image

### Example page of a person:
image


artlowel:
https://github.com/DSpace/dspace-angular/pull/390 is now also included in this PR, but #390 was turned in to a separate PR as well because the changes aren't specific to the theme

PR #402, opened 2019-05-20
Item relations edit delete

Atmire-Kristof:
This PR adds the possibility to remove relationships from an item.

### UI
A new tab labeled “Item Relationships” has been added to the item edit page (under sub-route `/relationships`)
The content of this tab functions much like the “Item Metadata” tab, except it is limited to deleting relationships (no adding or editing yet).

Relationships are also displayed as lists of items instead of a table, separated by relationship type.

Selecting items by clicking the trash bin icon will mark them with a red background.

Clicking “save” will remove the item relationships, display a notification stating the removal was successful (or an error notification for each failed delete request) and eventually reload the lists of relationships.

The other options (such as undo or discard) function the exact same way as it does on the edit item metadata page.

### Code changes
A notable code change that was made in this PR is a new abstract item-update component, which contains all the shared code between item-metadata and item-relationships. This is to avoid code duplication between the two components (and open for expansion).


src/app/+item-page/edit-item-page/item-relationships/edit-relationship/edit-relationship.component.ts
```diff
@@ -0,0 +1,74 @@
...
+   */
+  ngOnChanges(): void {
+    this.item = cloneDeep(this.fieldUpdate.field) as Item;
+  }
+
+  /**
+   * Sends a new remove update for this field to the object updates service
+   */
+  remove(): void {
+    this.objectUpdatesService.saveRemoveFieldUpdate(this.url, this.item);
```
paulo-graca: just after removing the relationship, the related item still appears in the "Items relationships" tab. And you can even try to repeat the same action (you will get an error - 
2019-06-18 12_21_29-Item Edit - Relationships
).
You need to switch to any other tab and then go back to Items relationships tab in order to be updated.

benbosman: Hi @paulo-graca
This was apparently caused by some updates in master from REST
Kristof has solved these errors now, and I was able to use delete relationships

src/app/+item-page/edit-item-page/item-relationships/edit-relationship-list/edit-relationship-list.component.html
```diff
@@ -0,0 +1,15 @@
+<ng-container *ngVar="(updates$ | async) as updates">
+  <div *ngIf="updates">
+    <h5>{{getRelationshipMessageKey(relationshipLabel) | translate}}</h5>
+    <ng-container *ngVar="(updates | dsObjectValues) as updateValues">
```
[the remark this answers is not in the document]
benbosman: Hi @paulo-graca
I would like to handle that use case in a separate PR as it will be dependent on https://github.com/DSpace/DSpace/pull/2450

Once that's in master, we can use paging per relationship type, which would be important for this functionality. This can also be used for item display pages once merged

src/app/+item-page/edit-item-page/item-relationships/edit-relationship/edit-relationship.component.ts
paulo-graca: Ok, thank you @benbosman! I will perform some more tests.

paulo-graca: Thank you @benbosman and @Atmire-Kristof , that fix that issue.
But just after removing that relationship, that item is still showed as a part of Works associated with that Person.
2019-07-01 17_00_40-dev5 rcaap pt_items_3f685bbd-07d9-403e-9de2-b8f0fabe27a7

and you click on that entry, you will see that the item is no longer related:
2019-07-01 17_00_56-dev5 rcaap pt_items_8e4314ca-2f60-4b0d-aa1f-48acb9aee6ed

It seems to me that is a caching issue.

benbosman: Thanks @paulo-graca
I've asked Kristof to verify it in detail, it seems that no new REST discovery is created for displaying the item page. The discovery cache would need to be invalidated here.

Do you know how this is handled for other situations where the discovery cache has to be invalidated (e.g. a new item is submitted, item metadata has been edited, …)?

Atmire-Kristof: It was indeed a caching issue. This should be fixed now.

paulo-graca: Ok, thank you @Atmire-Kristof and @benbosman! I will perform another test.

paulo-graca:
I'm sorry but I still see the related items in the list just after delete the relationship between them. 
Also, and perhaps this is another issue (not specifically related with this PR), my CPU reach top levels of 90% when I'm loading the page with relations.

paulo-graca:
> my CPU reach top levels of 90% when I'm loading the page with relations

I think this: https://github.com/DSpace/dspace-angular/issues/458 might be related with what I'm experiencing.

benbosman:
@paulo-graca Thanks for the analysis on the performance, we'll also verify these tickets

PR #407, opened 2019-05-23
Fixed filtered search page

LotteHofstede:
This PR resolves issues with the fixed filter search.
When a filtered search page (Eg. /search/person) was loaded using preboot, the SSR would return correct results, but CSR would request all search results (not filtered on entity type).

It also makes sure the loading icons are shown when the user changes pages or other parameters on the search page and new results are loading.

Edit: I found a small remaining issue on the MyDSpace page and fixed it.

Also, travis wasn't able to build anymore because the latest angular-cli version asks whether you want to send usage analytics back to google, and waits for a user prompt. This was causing a time out.

This question started appearing because Travis ran the `global` script before every build, causing the latest version of angular-cli to be installed, not the version specified in `package.json`. The `global` script hasn't been necessary for a while now.

I removed `yarn run global` from the travis config and now the build works again.

### DIFF
--- a/src/app/+search-page/search-filters/search-filter/search-filter.component.ts
+++ b/src/app/+search-page/search-filters/search-filter/search-filter.component.ts
@@ -129,22 +129,23 @@ export class SearchFilterComponent implements OnInit {
    * @returns {Observable<boolean>} Emits true whenever a given filter config should be shown
    */
   private isActive(): Observable<boolean> {
-    return this.selectedValues$.pipe(
-      switchMap((isActive) => {
-        if (isNotEmpty(isActive)) {
-          return observableOf(true);
-        } else {
-          return this.searchConfigService.searchOptions.pipe(
-            switchMap((options) => {
-                return this.searchService.getFacetValuesFor(this.filter, 1, options).pipe(
-                  filter((RD) => !RD.isLoading),
-                  map((valuesRD) => {
-                    return valuesRD.payload.totalElements > 0
-                  }),)
-              }
-            ))
-        }
-      }),
-      startWith(true));
+    return observableOf(true);
+    // return this.selectedValues$.pipe(
+    //   switchMap((isActive) => {
+    //     if (isNotEmpty(isActive)) {
+    //       return observableOf(true);
+    //     } else {
+    //       return this.searchConfigService.searchOptions.pipe(
+    //         switchMap((options) => {
+    //             return this.searchService.getFacetValuesFor(this.filter, 1, options).pipe(
+    //               filter((RD) => !RD.isLoading),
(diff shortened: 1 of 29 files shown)


src/app/+search-page/search-filters/search-filter/search-filter.component.ts
```diff
@@ -129,22 +129,23 @@ export class SearchFilterComponent implements OnInit {
...
+    //             return this.searchService.getFacetValuesFor(this.filter, 1, options).pipe(
+    //               filter((RD) => !RD.isLoading),
+    //               map((valuesRD) => {
+    //                 return valuesRD.payload.totalElements > 0
+    //               }),)
+    //           }
+    //         ))
+    //     }
+    //   }),
+    //   startWith(true));
```
paulo-graca: Do you expect to re-use this code again? Probably it would be better to remove this.

src/app/shared/services/route.reducer.ts
```diff
@@ -0,0 +1,49 @@
+import { Params } from '@angular/router';
```
paulo-graca: Could you add comments?

src/app/shared/services/route.service.ts
```diff
@@ -8,24 +8,46 @@ import {
...
+);
+export const queryParametersSelector = createSelector(
+  coreSelector,
+  (state: CoreState) => state.route.queryParams
+);
+
+export const routeParameterSelector = (key: string) => parameterSelector(key, routeParametersSelector);
+export const queryParameterSelector = (key: string) => parameterSelector(key, queryParametersSelector);
+
+export function parameterSelector(key: string, paramsSelector: (state: CoreState) => Params): MemoizedSelector<CoreState, string> {
```
paulo-graca: could you add comments in this function?

src/app/+search-page/search-filters/search-filter/search-filter.component.ts
LotteHofstede: I enabled this piece of code again, because it was accidentally commented out during testing something else.

src/app/shared/services/route.service.ts
LotteHofstede: Added TypeDoc for all selectors

src/app/shared/services/route.reducer.ts
LotteHofstede: Added TypeDoc for reducer

LotteHofstede:
@paulo-graca Thanks for your feedback! The requested changes have been applied.
I also fixed an issue I came across when looking at the code again: the route data is now cleared whenever the route changes.

PR #420, opened 2019-06-07
DS-4223 Metadata Schemas for configurable entities

benbosman:
One of the priorities identified in https://wiki.duraspace.org/display/DSPACE/2019-05-23+DSpace+7+Entities+WG+Meeting was the migration of the metadata fields for new entity types to schema.org
This PR removes the temporary metadata fields from phase 1 of the configurable entities which is also part of https://jira.duraspace.org/browse/DS-4223 and includes the new metadata fields and search facets on the item pages

Of course any other flat metadata schema can also be configured, DSpace does support the metadata registry to easily configure the desired metadata schema

This is related to https://github.com/DSpace/DSpace/pull/2443 and would preferably be merged simultaneously with that PR

### DIFF
--- a/src/app/entity-groups/journal-entities/item-list-elements/journal/journal-list-element.component.html
+++ b/src/app/entity-groups/journal-entities/item-list-elements/journal/journal-list-element.component.html
@@ -4,9 +4,9 @@
     [innerHTML]="firstMetadataValue('dc.title')"></a>
   <span class="text-muted">
     <ds-truncatable-part [id]="item.id" [minLines]="1">
-            <span *ngIf="item.allMetadata(['journal.identifier.issn']).length > 0"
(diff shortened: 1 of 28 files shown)


src/app/entity-groups/journal-entities/item-list-elements/journal/journal-list-element.component.html
```diff
@@ -4,9 +4,9 @@
     [innerHTML]="firstMetadataValue('dc.title')"></a>
   <span class="text-muted">
     <ds-truncatable-part [id]="item.id" [minLines]="1">
-            <span *ngIf="item.allMetadata(['journal.identifier.issn']).length > 0"
+            <span *ngIf="item.allMetadata(['creativeworkseries.issn']).length > 0"
```
paulo-graca: shouldn't we use the existing: `dc.identifier.issn` ?

PR #430, opened 2019-06-27
One-sided relationship filtering and refactoring

Atmire-Kristof:
This PR allows for filtering a list of relationships on one side, depending on the item you're fetching relationships for. This can be useful in cases where two entities of the same type have a relationship with each other, for example child-parent relationships. Without one-sided filtering, you wouldn’t know when to display the parent or child relationship on item pages.

To achieve this, `filterRelationsByTypeLabel` now has an optional parameter for the item's ID. Using the item's ID, we can filter out irrelevant relationships where the item is on the same side as the label requested.

To avoid having to call both operators every time you're filtering relations, I’ve also added a new operator called `getRelatedItemsByTypeLabel`, essentially combining the other two operators into one.

All entity pages have been refactored to use this new operator.

Some example code of how you would fetch parent and child related items for Organisational Units:

```
  this.parentOrgUnits$ = this.resolvedRelsAndTypes$.pipe(
    getRelatedItemsByTypeLabel(this.item.id, 'isParentOrgUnitOf')
  );

  this.childOrgUnits$ = this.resolvedRelsAndTypes$.pipe(
    getRelatedItemsByTypeLabel(this.item.id, 'isChildOrgUnitOf')
  );
```

### DIFF
--- a/src/app/+item-page/simple/item-types/shared/item-relationships-utils.ts
+++ b/src/app/+item-page/simple/item-types/shared/item-relationships-utils.ts
@@ -3,7 +3,7 @@ import { MetadataRepresentation } from '../../../../core/shared/metadata-represe
 import { MetadatumRepresentation } from '../../../../core/shared/metadata-representation/metadatum/metadatum-representation.model';
 import { MetadataValue } from '../../../../core/shared/metadata.models';
 import { getSucceededRemoteData } from '../../../../core/shared/operators';
-import { hasValue } from '../../../../shared/empty.util';
+import { hasNoValue, hasValue } from '../../../../shared/empty.util';
 import { Observable } from 'rxjs/internal/Observable';
 import { Relationship } from '../../../../core/shared/item-relationships/relationship.model';
 import { RelationshipType } from '../../../../core/shared/item-relationships/relationship-type.model';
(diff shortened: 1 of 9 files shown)


src/app/+item-page/simple/item-types/shared/item-relationships-utils.ts
```diff
@@ -43,26 +51,32 @@ export const compareArraysUsingIds = <T extends { id: string }>() =>
...
+export const filterRelationsByTypeLabel = (label: string, side?: RelationshipSide) =>
   (source: Observable<[Relationship[], RelationshipType[]]>): Observable<Relationship[]> =>
     source.pipe(
       map(([relsCurrentPage, relTypesCurrentPage]) =>
         relsCurrentPage.filter((rel: Relationship, idx: number) =>
-          hasValue(relTypesCurrentPage[idx]) && (relTypesCurrentPage[idx].leftLabel === label ||
-          relTypesCurrentPage[idx].rightLabel === label)
+          hasValue(relTypesCurrentPage[idx]) && (
+            (hasNoValue(side) && (relTypesCurrentPage[idx].leftLabel === label ||
+              relTypesCurrentPage[idx].rightLabel === label)) ||
```
benbosman: It's best not to work with a side parameter, but rather to check the label vs the item ID:
* If the current item occurs on the left, the label should match the left one OR
* if the current item occurs on the right, the label should match the right one

It would also be useful to add some comments explaining this behavior

Atmire-Kristof: As requested, I've removed the RelationshipSide enum and replaced it with the item's ID. Using the item's ID, it now filters out irrelevant relationships where the item is on the same side as the label requested.
I've also updated the PR's description to match this change.

PR #441, opened 2019-07-31
Temporary fix to Travis webdriver chrome version

Atmire-Kristof:
This PR adds `--versions.chrome 75.1.0` to the webdriver-manager start and update as a temporary fix to failing Travis builds.
Example PR that currently has this issue: https://github.com/DSpace/dspace-angular/pull/440


artlowel:
We tried a few things, but nothing worked so far. It seems it's due to this webdriver-manager issue: https://github.com/angular/webdriver-manager/issues/404. We'll try again when there's a fix

PR #444, opened 2019-08-01
Patch protractor with the latest webdriver-manager

artlowel:
Travis has switched to chrome 76, which is out of the range of supported versions used by the webdriver-manager that ships with the latest version of protractor. This PR adds the latest version of webdriver-mangager as a direct dependency, and patches protractor after `yarn install`. This patch and the direct dependency on webdriver-manager and ncp (needed to do the patch) can be removed after a protractor update that includes webdriver-manager >= 12.1.6


artlowel:
The protractor issue is tracked here: https://github.com/angular/protractor/issues/5289

PR #450, opened 2019-08-02
WIP:  Increasing contrast on two colors

lhenze:
For WCAG 2.0 compliance, text needs to have "at least 4.5:1" contrast.

### DIFF
--- a/src/styles/_custom_variables.scss
+++ b/src/styles/_custom_variables.scss
@@ -36,3 +36,5 @@ $edit-item-metadata-field-width: 190px;
 $edit-item-language-field-width: 43px;
 
 $thumbnail-max-width: 175px;
+
+$text-muted: #757575;
--- a/src/styles/_custom_variables.scss
+++ b/src/styles/_custom_variables.scss
@@ -36,5 +36,3 @@ $edit-item-metadata-field-width: 190px;
 $edit-item-language-field-width: 43px;
 
 $thumbnail-max-width: 175px;
-
-$text-muted: #757575;
--- a/src/styles/_bootstrap_variables.scss
+++ b/src/styles/_bootstrap_variables.scss
@@ -24,7 +24,7 @@ $gray-100: lighten($gray-base, 93.5%) !default; // #eee
 /* Reassign color vars to semantic color scheme */
 $blue:      #2B4E72 !default;
 $green: #94BA65 !default;
-$cyan: #2790B0 !default;
+$cyan: #007C88 !default;
 $yellow: #ec9433 !default;
 $red: #CF4444 !default;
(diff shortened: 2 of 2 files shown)


src/styles/_custom_variables.scss
```diff
@@ -36,3 +36,5 @@ $edit-item-metadata-field-width: 190px;
 $edit-item-language-field-width: 43px;
 
 $thumbnail-max-width: 175px;
+
+$text-muted: #757575;
```
artlowel: Since this overrides a variable that comes with bootstrap, it should go in _bootstrap_variables.scss as well.

lhenze: @artlowel. I agree, it would be good to use another classname.   However,  it does seem that that class is used widely through the codebase, and it would be better to never have text that is illegible.  So, up to you.   Good to know about the variable placement, I can make another pull request for that. 

Are you ok with overwriting $cyan ?   Since the original is specified as !default, we could re-assign elsewhere too.

artlowel: >Are you ok with overwriting $cyan ? Since the original is specified as !default, we could re-assign elsewhere too.

`$cyan` is in the right file, I would expect it there

>Good to know about the variable placement, I can make another pull request for that.

No need to make a separate pull request, just check out this branch ([449-textcolors](https://github.com/lhenze/dspace-angular/tree/449-textcolors)) move `$text-muted` to `_bootstrap_variables.scss`, commit, and then push, and it will show up in this PR.

LotteHofstede: @lhenze IMHO, the value of `$text-muted` should be derived from the `$body-color` variable. Otherwise, you could have instances where the muted text is actually brighter than the rest of the text. 
If you're set on using `#757575`, I would prefer to use another variable/CSS-class.

lhenze:
New idea.
1) Create a new gray color that *does* pass a11y check
2) Assign some of the failing text color variables to it, in case they are used in the future. 
3) Not yet done -- use a different class name when the text does not actually need to be muted.

src/styles/_custom_variables.scss
lhenze: @LotteHofstede I see what you mean, it should work within the whole system.  But if the muted text is brighter than the rest of the text, then the rest of the text will be too low contrast and will fail the check. (assuming they're both variants of gray and have the same background.)  

I just updated the PR with a new color $gray-710 (rather than changing $gray-700) with the intent of not unintentionally affecting anything else (such as UI element borders.)  But they're all percentages of $gray-base.  (It looks like $body-color is $gray-900 by default.)

Does this make sense or should we just look at each text-color related variable individually?

PR #461, opened 2019-08-22
Rename relationship type properties

benbosman:
This PR includes the Angular changes for https://github.com/DSpace/DSpace/pull/2488
Since the property is renamed in https://github.com/DSpace/DSpace/pull/2488, Angular has to be updated as well


benbosman:
Merging since https://github.com/DSpace/DSpace/pull/2488 has been merged and this has 3 approvals but was just waiting for https://github.com/DSpace/DSpace/pull/2488

PR #463, opened 2019-08-24
Pagination messaging:  Hiding "Now Showing"

lhenze:
Here is an idea to fix this small bug.
Moving the ngIf directive to wrap both parts of the statement.


lhenze:
Yes! Thanks @artlowel

PR #465, opened 2019-08-26
Collection pages, recent submissions: Added "empty" message

lhenze:
Modeled on browse-by.component.html


src/app/+collection-page/collection-page.component.html
```diff
@@ -52,6 +52,9 @@ <h2>{{'collection.page.browse.recent.head' | translate}}</h2>
                           message="{{'error.recent-submissions' | translate}}"></ds-error>
                 <ds-loading *ngIf="!itemRD || itemRD.isLoading"
                             message="{{'loading.recent-submissions' | translate}}"></ds-loading>
+                <div *ngIf="!itemRD?.isLoading && itemRD?.payload?.page.length === 0" class="alert alert-info w-100" role="alert">
+                  {{'browse.empty' | translate}}
```
artlowel: In these situations I prefer to duplicate the key. Because even though in english the message may be exactly the same, in other languages it might not be, and if they don't use different keys, translators don't have the option to change the message.

lhenze: @artlowel, understood.  I see there is a "collection.page.browse.recent.head" key.   Fine for the new one to be "collection.page.browse.recent.empty"  ?
Ok to put it in all four language json files, copying the value from "browse.empty"?

artlowel: `collection.page.browse.recent.empty` sounds fine.

We ask devs to always update `en.json` so that it contains all the keys. If you happen to be fluid in another supported language, it'd be great if you can add it, but it's not a requirement.

Issue #394, opened 2019-05-10
forceBypassCache should be removed from the RequestService


artlowel:
When the submission and MyDspace code was originally written, all requests were cached and for the same amount of time. This was an issue because a POST requests would get cached as well so subsequent POSTs wouldn't be executed etc. 

To work around this problem a forceBypassCache param was added to the RequestService that would ensure the request would be sent without first looking for a cached response.

This problem has been fixed. Only GET requests are cached now, and [when creating a request you can specify how long its response should be cached](https://github.com/DSpace/dspace-angular/blob/f54a3a4d6b3319af83857b7af02e076f7de6037e/src/app/core/data/request.models.ts#L25), so all instances where forceBypassCache is used can be refactored, and the param can be removed from the RequestService.

PR #468, opened 2019-08-29
forceBypassCache param removed

mspalti:
All references are (or should be) gone and the requestService is modified.

Fixes https://github.com/DSpace/dspace-angular/issues/394

Also, I had an unrelated problem with the unit test for date lists.  Although it had passed previously, the 1950-01-01 full year was being translated to 1949 in my environment.  I changed the method call from getFullYear() to getUTCFullYear() and the problem was solved.  I left the change here in the code with a TODO note.


mspalti:
@artlowel , when the forceBypassCache had been `true` (in dataService, authRequest) I set the `responseMsToLive` property to 0 as you suggest.

For subclasses (e.g. WorkflowItemDataService) that had previously set `forceBypassCache` on the dataService, I introduced a new property on dataService and then set the value of that new property in the subclasses.  The value is set to 0 in call cases. Let me know if you'd prefer to just set `responseMsToLive` to 0 in dataService for now since it's clear that something like this property will be needed.

src/app/core/data/data.service.ts
```diff
@@ -134,7 +134,7 @@ export abstract class DataService<T extends CacheableObject> {
       first((href: string) => hasValue(href)))
       .subscribe((href: string) => {
         const request = new FindAllRequest(this.requestService.generateRequestId(), href, options);
-        if (this.resetMsToLive) {
+        if (this.resetMsToLive !== undefined) {
```
artlowel: Because this is a use case that occurs very often (and indeed because of truthy/falsy values you're better off avoiding `if(variable)` in JS), we have a method for this: [`hasValue()`](https://github.com/DSpace/dspace-angular/blob/2cbe6a6d91d2ff2fbb2da2543174426e8a120f0a/src/app/shared/empty.util.ts#L84)

src/app/core/data/data.service.ts
```diff
@@ -45,11 +45,14 @@ export abstract class DataService<T extends CacheableObject> {
...
   protected abstract halService: HALEndpointService;
-  protected abstract forceBypassCache = false;
   protected abstract objectCache: ObjectCacheService;
   protected abstract notificationsService: NotificationsService;
   protected abstract http: HttpClient;
   protected abstract comparator: ChangeAnalyzer<T>;
+  /**
+   * Allows subclasses to reset the response cache time.
+   */
+  protected resetMsToLive: number;
```
artlowel: I think it's a useful property to have on all dataservices, but I'd also call it `responseMsToLive` or perhaps `defaultResponseMsToLive` because that's what it is: the time in ms responses for this dataservice should stay in the cache. 
With the name `resetMsToLive` I would expect a boolean that would set `responseMsToLive` to 0 when `true`.

mspalti: You're right. I originally added a boolean property and then decided that number is more useful here. The name should have changed, too.

src/app/core/data/data.service.ts
mspalti: Nice.  I'll remember to use it.

artlowel:
It shouldn't be necessary to ever manually clear those requests. 

The reason `requestsOnTheirWayToTheStore` exists, is because a request doesn't immediately arrive in the store as soon as you dispatch the action. The store processes these actions asynchronously. 

So when the requestService dispatches a new `RequestConfigureAction`, the `trackRequestsOnTheirWayToTheStore` [adds that request to a local array](https://github.com/DSpace/dspace-angular/blob/0b878af92a30ff31bc67cc956863f8ec22b9e3b6/src/app/core/data/request.service.ts#L251) and immediately after [selects the place in the store where that request will be put when the store processes the action](https://github.com/DSpace/dspace-angular/blob/0b878af92a30ff31bc67cc956863f8ec22b9e3b6/src/app/core/data/request.service.ts#L252). It subscribes to that observable to [remove the request from the local array when it turns up in the store](https://github.com/DSpace/dspace-angular/blob/0b878af92a30ff31bc67cc956863f8ec22b9e3b6/src/app/core/data/request.service.ts#L256). 

I don't see any reason to ever manually clear that array. If two submission requests to the exact same URL happen in such quick succession that the first request hasn't even reached the store when the second arrives, we should just ignore the second request and await the response to the first.

src/app/core/data/request.service.ts
```diff
@@ -145,14 +145,11 @@ export class RequestService {
...
-   * @param {boolean} forceBypassCache When true, a new request is always dispatched
    */
-  configure<T extends CacheableObject>(request: RestRequest, forceBypassCache: boolean = false): void {
+  configure<T extends CacheableObject>(request: RestRequest): void {
     const isGetRequest = request.method === RestRequestMethod.GET;
-    if (forceBypassCache) {
-      this.clearRequestsOnTheirWayToTheStore(request);
-    }
-    if (!isGetRequest || (forceBypassCache && !this.isPending(request)) || !this.isCachedOrPending(request)) {
+    const isSubmission =  request instanceof SubmissionRequest;
```
mspalti: This might not be the right solution, but it does appear to fix the problem noticed by @atarix83 

I think previously the combination of forceBypassCache and removing pending requests from `requestsOnTheirWayToTheStore` had the effect of forcing a new dispatch request. Without this, a lookup by href is attempted and fails. Checking for the request type is a different way to get the same result but may have unintended consequences.  

Also, I noticed that `SubmissionService` and `RemoteDataBuildService` are both importing methods from the `shared/testing/utils` It seems like those methods would be better located in a different utility.

mspalti:
Now that see what's causing the problem I'm wondering about this suggestion by @artlowel 

> first request hasn't even reached the store when the second arrives, we should just ignore the second request and await the response to the first.

That's sort of what I did here but I think at the cost  of dispatching a second and (maybe!) unnecessary`RequestExecuteAction.` I'll look at this tomorrow.  If either of you see this before feel free to make suggestions.

mspalti:
I looked at this more.  From what I saw, I don't think the issue is related to pending requests in `requestsOnTheirWayToTheStore` but there do seem to be issues with the cache. 

I still don't fully understand the data cache, but it's clear that simply calling RequestService.getByHref and dispatching an AddToIndexAction fails because there's no href workspace entry in the `get-request/href-to-uuid` cache (even when data for the workspace item does exist in `cache/object`).

mspalti:
Also RequestService.isCachedOrPending correctly returns true because the entry exists in the `object/cache`. But currently this should be ignored for submissions.

Presumably the cached data will be used if we "forceBypassCache" by other means, but I'm still not clear on why this is necessary in the case of workspace submission requests.

mspalti:
OK. I think I see the reason why bypassing the cache is necessary (for example by adding something like `isSubmission` to `RequestService.configure`). In the current state there's no entry in `data/request`. But an entry is required for the lookup in `SubmissionRestService.fetchRequest`. This entry is created when we bypass cache and dispatch new configure and execute actions. That forces a new REST api request.

It's a "new" request because data from prior workspaceitem requests does exist in the object cache and is ignored. I'm certain that is by design. It's probably a good idea to always fetch workspace item data from the server, in case something has changed.

atarix83:
So the situation is that the workspaceitem is present in `cache/object` because it's retrieved previously while loading the mydspace page. When requesting the workspaceitem by id there is no response because there is no entry in `data/request`.
So regardless if a new request is dispatched or not, we should get a response, I think this is a bug to fix. 

maybe @artlowel could have suggestions in this regard

artlowel:
I took a look in the debugger, and the reason for the issue @atarix83 reported seems to be that the workspace item is cached in the object cache because it was retrieved for the mydspace page. It is cached for the default duration of 15 minutes:[ the object cache msToLive doesn't inherit the responseMsToLive time](https://github.com/DSpace/dspace-angular/blob/f980b55c1cc7977f92121c65811f12e50a19bf12/src/app/core/data/base-response-parsing.service.ts#L117), and that's a bug. 

I fixed it in [this commit](https://github.com/atmire/dspace-angular/commit/4836fd9ec8fde409081fa4228abad1a4c4de213d) ([The branch](https://github.com/atmire/dspace-angular/commits/retain-mstolive-in-object-cache) started from this PR). Once I did that though I also had to [increase the places where responseMsToLive was 0 to 10 seconds](https://github.com/atmire/dspace-angular/commit/254b7d1633f4927e0486c60c0db028bc1e172eb7) (lowering it further could work as well, haven't tested it) because otherwise I got stuck in an infinite loop of requests being removed from the cache before they could be used, and being re-requested.

I don't think it makes a lot of sense releasing this as a separate PR against master, as it will affect a lot of the same files and conflict with this PR. So I'd say @mspalti, take a look at my branch, and if you're happy it works and solves your issues, merge it with yours and push it to this PR.

mspalti:
@artlowel I reviewed your fix and merged it with this PR. I reduced the `responseMsToLive `values to 5 seconds, and then to 2 seconds. Both worked, somewhat. But if I navigated between mydspace and submissions I seemed get into a loop on occasion with 2 seconds.  Possibly with 5 as well. It was a bit difficult to test because problems were intermittent. 

Also, at 10 seconds it appeared that I could recreate the original problem that @atarix83 noticed if I was quick to select an item after first navigating to mydspace.

atarix83:
@artlowel thanks for fix
@mspalti i think that if can you reduce responseMsToLive to 1 sencond [here](https://github.com/mspalti/dspace-angular/blob/254b7d1633f4927e0486c60c0db028bc1e172eb7/src/app/core/data/request.models.ts#L395) and [here](https://github.com/mspalti/dspace-angular/blob/254b7d1633f4927e0486c60c0db028bc1e172eb7/src/app/core/submission/submission-rest.service.ts#L113) so there should not be other issues

mspalti:
@atarix83 unfortunately when I reduce times I see problems.  I just tried reducing the times as you suggested and mydspace stopped working for me. I see a loop in which requests are repeated; eventually the client stops responding. Do you see the same thing?

atarix83:
@mspalti  you're right but as you said problems were intermittent and at 10 seconds the original problem that I noticed occurs again.

I think that as is this PR can't be merged, we should find a final solution

artlowel:
I think the loading issue is due to [this `find()`](https://github.com/mspalti/dspace-angular/blob/254b7d1633f4927e0486c60c0db028bc1e172eb7/src/app/submission/submission.service.ts#L514) in the `SubmissionService.retrieveSubmission()` method. If I replace it with a `filter`, followed by a `take(1)` the issue goes away, even if you increase responseMsToLive back to 15 minutes (I'm not saying we should do that, but it does indicate that the loading issue is not directly caused by caching time, lowering the caching time will just hide the symptoms) 

I'm also not entirely sure _why_ this works. I tried it because I've seen it work in the past. The only difference between `find` and the combination of `filter` and `take(1)` I can think of, is that `find` will emit `undefined` if the source observable completes without a match, while  `filter` won't emit anything. But as far as I can tell the source observable doesn't complete in this case…

mspalti:
@artlowel I tried replacing the `find()` with a filter as you did, but in my case it didn't seem to help. I'd be curious to know if it works for @atarix83 since it's possible I did something wrong.

Looking at this again today, I decided that the issue that @atarix83 found is a consequence of looking up objects in `SearchService.search`, since `search` retrieves individual objects  and adds these to `cache/object` but doesn't create new entries in `data/request`. We're able to expire these objects by adjusting `reponseMsToLive` but if the value is too small the mydspace page is unable to render.

There's quite possibly a better, more general solution to this problem. But, one solution is to add `(href) => this.cacheService.remove(href)` to the observable in `SubmissionRestService.getDataById()`, before the call is made to `RestService.configure()`.

mspalti:
Here's the filter I added to `SubmissionService.retrieveSubmission()`

```
      filter((submissionObjects: SubmissionObject[]) => isNotUndefined(submissionObjects)),
      take(1),
```
I think that after the change I could still get to the submission link before the object expired and the result was the same as before.

mspalti:
Regarding the workspace items in `object/cache`, as an experiment I did try to use them but they appear to be incomplete. That's why I decided to try removing them before configuring the request, which did seem to fix the bug. Again, I'm not sure that it's a correct solution.

Here's the full picture:
```
  public getDataById(linkName: string, id: string): Observable<SubmitDataResponseDefinitionObject> {
    const requestId = this.requestService.generateRequestId();
    return this.halService.getEndpoint(linkName).pipe(
      map((endpointURL: string) => this.getEndpointByIDHref(endpointURL, id)),
      filter((href: string) => isNotEmpty(href)),
      distinctUntilChanged(),
      tap((href) => this.cacheService.remove(href)),
      map((endpointURL: string) => new SubmissionRequest(requestId, endpointURL)),
      map ((request: RestRequest) => {
        request.responseMsToLive = 10 * 1000;
        return request;
      }),
      tap((request: RestRequest) => this.requestService.configure(request)),
      flatMap(() => this.fetchRequest(requestId, id)),
      distinctUntilChanged());
```

atarix83:
unfortunately there another side effect for increasing [here](https://github.com/mspalti/dspace-angular/blob/254b7d1633f4927e0486c60c0db028bc1e172eb7/src/app/core/submission/submission-rest.service.ts#L113) the `responseMsToLive` to 10 seconds. 
Indeed the mydspace doesn't work properly, if with admin user perform any actions on a task within 10 seconds of page loading, the task list is not reloaded and it seems that the action has not been executed

@mspalti  About the solution to delete the cache on `getDataById`, is essentially what was done before with [clearRequestsOnTheirWayToTheStore method](https://github.com/DSpace/dspace-angular/pull/468/files#diff-c803ef5adac3ba3281e5ecd80b065dc6L153). I don't like to apply in this way because it's only a workaround for this method and it's not a final solution that works properly every time

mspalti:
@atarix83 , at least [in that example](https://github.com/mspalti/dspace-angular/blob/254b7d1633f4927e0486c60c0db028bc1e172eb7/src/app/core/submission/submission-rest.service.ts#L113) I think I noticed (yesterday) that it's possible to decrease the `responseMsToLive` to one second.  I didn't notice a problem when I did this.  Doing it in `MyDspaceRequest` does create a problem for mydspace (I think because it is the model used in SearchService).

I agree that my suggestion to delete the search object cache is a workaround.

mspalti:
I agree with @artlowel that the basic issue here is not related to cache time. Unfortunately his suggestion to replace `find()` with `filter()` still doesn't seem to fix the problem for me. Again, maybe I missed something. But I think the observable doesn't complete.

I think the `SubmissionRestService.getDataById()` and `SubmissionRestService.fetchRequest()` assume that there was no href entry in the `cache/object`, and that `RequestService` will retrieve a new workspaceitem from the server. That doesn't happen because entries have previously been added to the object cache by `SearchService` while initializing mydspace. Everything works if those entries are expired and removed from the cache, but if that happens too quickly we have other problems.

I think one way forward would be to retain forceBypassCache  in `RestRequest` and set it to false. In cases like this one, in which the default caching logic must be bypassed and we need to guarantee that a new request always happens, we could override the default, for example in `SubmissionRequest`.  

Then create a new issue and look for a different solution if we need to.

The other way forward is to block this PR until an underlying problem can be identified and fixed. That's fine with me. This has been great motivation for me to learn about the angular client so I don't feel any of the effort has been wasted.

mspalti:
[Setting reponseMsToLive to 0 works here](https://github.com/DSpace/dspace-angular/pull/468/files#diff-5ef0f3c9932bd01ea3d9ffba2a1065d1R113), I think.

mspalti:
The most recent commit adds `forceBypassCache` to the request model (RestRequest).  I'm happy to role that back if it's not the direction to go.

This property is no longer part of the `DataService`. The intent is simply to make it possible to specify that a given request type should always retrieve new data. That appears to be the intent with `SubmissionRequest`.

artlowel:
>I think one way forward would be to retain forceBypassCache in RestRequest and set it to false. In cases like this one, in which the default caching logic must be bypassed and we need to guarantee that a new request always happens, we could override the forceBypassCache property as needed, for example in SubmissionRequest. That still largely removes forceBypassCache from the rest of the code.
>
>Then create a new issue and look for a different solution if we need to.

I agree with that approach. 

I wouldn't block this PR because it improves upon the previous way of working in that the `forceBypassCache` property is more contained, that `responseMsToLive` is used instead in a lot of places, and that `responseMsToLive` is also respected by the object cache. 

So in my opinion it can be merged as-is.

atarix83:
I would like to state that I agree with the proposed solution, but I had a deep test on this PR and unfortunately I think it can't be merged as is.

Indeed the mydspace page doesn't work properly because the issue that I reported [here](https://github.com/DSpace/dspace-angular/pull/468#issuecomment-540460432) still occurs.

I've tried to add `forceBypassCache` property [here](https://github.com/mspalti/dspace-angular/blob/e38556e8214fd3e3755ae4e079426b1dbdd837db/src/app/core/data/request.models.ts#L409) but the result is that it doesn't work at all because the page stays hanging.

Additionally I also noticed that when click on workspaceitem edit the loading spinner is hanging for several seconds before to load data.

So at this state I can't approve this PR.

mspalti:
@atarix83 , right now I'm testing against the demo REST api at 4Science. It has many collections, and that seems to dramatically affect angular client performance. That said, I tried running against the master branch for comparison with this PR. I could not detect any differences between in performance or results between master and the PR branch (which includes recent updates to master).  

It might be better if I test against a smaller, local dspace instance but I can't do that at the moment (I'm travelling and using a laptop). But this does make me wonder if some of the problems here might be related to changes elsewhere.

mspalti:
@atarix83 , once I got back to my test dspace instance I think saw the problem that you discovered. For example, I deleted a workspaceitem from the list immediately after the mydspace page loaded. The delete request was processed.  But after that happened, the item was still in the list.

I tried modifying responseMsToLive and forceBypassCache in several places, and as you noted this doesn't seem to work.  

What seems to work is [emptying the object cache](https://github.com/DSpace/dspace-angular/pull/468/files#diff-b7ecbd6430bb4394fd497d33e6b1be9bR121) in `MyDSpacePageComponent ngOnInit()`, before the `SearchService` is called. That assures that a fresh list is requested every time the component is initialized.

artlowel:
Emptying the affected part of the cache after a delete makes sense. But in the past we've only done it inside the delete method on the dataservice rather than every time onInit. I guess we'll have to make an exception here, at least for now, because mydspace only seems to work if nothing is cached.

Please do [track the subscribe](https://github.com/DSpace/dspace-angular/pull/468/files#diff-b7ecbd6430bb4394fd497d33e6b1be9bR121) though, so you can destroy it onDestroy, or use a `take(1)`

mspalti:
Explicitly unsubscribing in onDestory seems better. Referring to this discussion in the code comment.

mspalti:
The e2e test failed to start on travis. The error message says incompatible chrome version.

artlowel:
That issue should be fixed if you merge in the latest master

mspalti:
OK, e2e tests succeeded after merge.

atarix83:
@mspalti thanks for the work on this issue.

your solution works properly.

However there is a better way of emptying the object cache. Indeed there is no need to clear cache every time mydspace page is load, but only after :
- a workspaceitems is deposited or deleted. This could be done by adjusting [this method](https://github.com/DSpace/dspace-angular/blob/master/src/app/submission/submission.service.ts#L462).
- an action on a task is performed (claim, return to pool, approve, reject) this could be done [here](https://github.com/DSpace/dspace-angular/blob/master/src/app/shared/mydspace-actions/mydspace-actions.ts#L59)
In this way we prevent to clear cache even when a search filter is clicked.

Feel free to resolve on this PR, otherwise please create a new issue for this.

mspalti:
@atarix83 I will make the changes you suggest on this PR.

mspalti:
@atarix83 , I moved clearing of the object cache into the methods you identified.  Code seems to be working as before.  I'm happy to make minor changes based on feedback. Otherwise, this appears ready to merge.

atarix83:
@mspalti thanks for the change I had a look again and it looks good to me. 

I think that is ready to me merged now

PR #472, opened 2019-09-04
Collections pages: add collection header and footer to all "browse by..." views

lhenze:
Also added a collection-level "home" page button (Browse by submission date).
Also added a community-level "home" page button (Subcommunities and collections) as an experiment.

Based on this: https://wiki.duraspace.org/display/DSPACE/Collections+Pages+Mockups


lhenze:
I see the error "Property binding ngVar not used by any directive on an embedded template. Make sure that the property name is spelled correctly and all directives are listed in the "@NgModule.declarations"."   Referring to this line.  https://github.com/lhenze/dspace-angular/blob/460-coll-pages/src/app/%2Bbrowse-by/%2Bbrowse-by-metadata-page/browse-by-metadata-page.component.html#L2 .     I'm not sure how to fix this. 

It compiles fine though.

lhenze:
Hi, This now passes the checks.  I realize there are a few different things going on here.  Let me know if it would be better to break this up into smaller PRs.

lhenze:
Thanks so much @artlowel .  Working on these.

artlowel:
@lhenze as discussed in the meeting:

In order to generate a UI URL you'll first need to inject the config in to your handle component. So basically add a constructor [similar to this](https://github.com/DSpace/dspace-angular/blob/dd24958f618980cdaaab1e8ff5250054944de2fd/src/app/core/shared/hal-endpoint.service.ts#L25):

```typescript

constructor(@Inject(GLOBAL_CONFIG) private EnvConfig: GlobalConfig) {
  }
```

Then you can use the UIURLCombiner to generate the url:

```typescript
const href = new UIURLCombiner(this.EnvConfig, '/handle/', this.content).toString();
```

lhenze:
@artlowel  handle component is updated, let me know if it's what you were thinking. 

Still to do is the different widgets depending on device width.  Do you have a link or suggestion for a good way to implement that?

lhenze:
Hi @artlowel  , above you say: 


> The buttons also don't work very well on xs screens, they don't fit, and are very high due to their contents.
> 
> image
> 
> Perhaps we can render them as a dropdown on xs screens, e.g.:
> 
> image
> 

To me your image looks like a \<select\> element rather than a dropdown widget.   The difference I think being that the currently active option is the visible one, and there is no button to press to reveal the list.  I don't know the bootstrap/ngb dropdown widget well but it doesn't seem like it offers those features.

artlowel:
The handle changes look good, thank you.

I meant a select yes. I have the unfortunate tendency to use the words select and dropdown interchangeably, but they are different things.

>Still to do is the different widgets depending on device width. Do you have a link or suggestion for a good way to implement that?

I'd put another level of div's inside the nav in comcol-page-browse-by.component.html and use bootstrap's [visibility helper classes](https://getbootstrap.com/docs/4.0/utilities/display/#hiding-elements). e.g. the following would show the select only on XS screens:

```html
<nav class="comcol-browse mb-4" aria-label="Browse Community or Collection" >
  <div class="d-none d-sm-block">
    <!-- Current menu goes here -->
  </div>
  <div class="d-block d-sm-none">
    <!-- Select goes here -->
  </div>
</nav>
```

lhenze:
Hi, I made some progress towards the select element for navigation for mobile, perhaps.  It works *sometimes*.  Any suggestions appreciated.

artlowel:
Ok, I finally found something that works. You can find the code in [this gist](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e)

The reason we saw the first option flash momentarily before the correct option was selected was because we're using server-side rendering. The page gets rendered on the server first, so bots and people without javascript can still use it. When it arrives on a client with JS, the JS code takes over, that's when you see the loading screen and eventually the right option.

So for some reason, the server didn't wait for the selected option to be set before deciding it has finished rendering and sending the html to the client. I think this is due to a bug in how angular initialises a select, but I'm not 100% sure. 

The way I work around it in the latest code is by, not only [attaching the currentOption to the select as its ngModel](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-html-L18), but by also [setting the selected attribute manually](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-html-L21). That last bit shouldn't be necessary, but in our case it _does_ happen server-side while angular responding to the ngModel version only happens client-side.

Here is a bit of an explanation for the changes I made to the code:
- I created a list in the component that contains all the possible values a button or an option in the select can have, that way we can get rid of [those `*ngIfs` in the template](https://github.com/DSpace/dspace-angular/blob/d4d4a6121e2e2965a6ff805db609f169ea9a4b99/src/app/shared/comcol-page-browse-by/comcol-page-browse-by.component.html#L15), and render everything with an [`*ngFor`](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-html-L19)
- Instead of putting all the data needed to select an option and to redirect to the correct option in the template, I created [an interface to represent all the info we need](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-ts-L19): ComColPageNavOption
- During ngOnInit, [I first take the browseBy types in the config, and perform a map operation on them](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-ts-L57), that basically means, turn every element of this array in to something else using the function inside the map operator.
	- So inside the map function I each BrowseByTypeConfig object in to a ComColPageNavOption object 
- Afterwards[ I check whether we're dealing with a collection or a community](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-ts-L65), and in each case I create a ComColPageNavOption object, and add it to the front of the allOptions array
	- [`...this.allOptions` means](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-ts-L70) enumerate everything that's currently in the allOptions array, so I put my new object first, then enumerate everything that was already there, and assign the result again to this.allOptions
- Next in order to know what page we're on, I use[ the route.url observable angular provides us that contains the current url](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-ts-L79), pipe that through a filter to ensure it has a value for the map below, which will just get the last part of the current url, so the community or collection id, or the type of the browse page.
- That `currentOptionId$` observable is then used in the template [to set the ngModel for the select](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-html-L18), which, as mentioned above, should have been enough to select the correct option, but as it wasn't I also use it to [set the selected attribute on the options](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-html-L21)
- Because we're working with ngModel, we're no longer getting a generic JS event in the onSelectChange, but simply the id of the selected option. I use that Id to [find the option in the allOptions array that matches](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-ts-L86), and then use the data inside that matching option [to navigate to the correct page](https://gist.github.com/artlowel/1db53990a1bcb326b4efa7e797c0bc3e#file-comcol-page-browse-by-component-ts-L89).

lhenze:
Art's changes are integrated and all works as expected, from the testing I've done.
Note -
There is still styling to do for these pages, especially typography.  I propose we consider that a separate ticket, since it's not necessarily specific to collections / community pages?

artlowel:
I did another quick check of everything, and I noticed that the site-wide browse pages are broken now. They just show an empty page.

I think the reason is [this ngIf](https://github.com/DSpace/dspace-angular/blob/300aeb435ef5c0ce4f0044770fbd55669b344e73/src/app/+browse-by/+browse-by-metadata-page/browse-by-metadata-page.component.html#L3) in browse-by-metadata-page.component.html It only ends in line 51, so that means if there's no parent, nothing is rendered. 

Looking closer at browse-by-metadata-page.component.html I also don't like the fact that it duplicates all the fields on community/collection pages. But fixing that now would lead us down another rabbit hole. So for this PR, I'll simply ask you to split up that big ngIf in two: one for the section above the ds-browse-by, and one for the section below. And then the refactor can happen in later.

PR #484, opened 2019-09-16
Update set-value to fix CVE-2019-10747

artlowel:
This PR fixes [CVE-2019-10747](https://nvd.nist.gov/vuln/detail/CVE-2019-10747)

For future reference, I [read about](https://github.com/yarnpkg/yarn/issues/4986#issuecomment-530206555) a way to make this process a lot easier: You can add a `"resolutions"` section to package.json with any minimum version requirements for transitive dependencies, in this case:

```json
  "resolutions": {
    "set-value": ">= 2.0.1"
  },
```

and then run `yarn install` and yarn will update yarn.lock to honor those requirements for you, or warn you if it can't. The package doesn't need to be a direct dependency for this to work. ~And after yarn.lock has been updated, you might as well remove it from package.json again before you commit, to keep the amount of clutter down.~ 👈 that last bit didn't work, it allowed the 0.x version back in

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,9 @@
   "engines": {
     "node": "8.* || >= 10.*"
   },
+  "resolutions": {
+    "set-value": ">= 2.0.1"
+  },
   "scripts": {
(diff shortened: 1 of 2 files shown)


artlowel:
@tdonohue I noticed that as well, but those missing scores seem to have come in now. Perhaps their servers couldn't cope?

PR #503, opened 2019-10-30
Update webdriver-manager library

artlowel:
Update for the e2e tests to work with the current version of chrome on travis


artlowel:
Merging, because it is trivial, and needed for travis tests to work

Issue #507, opened 2019-11-07
SubmissionService.getSubmissionObjectLinkName uses the route to determine the type of the submission object


artlowel:
The [`getSubmissionObjectLinkName()` method in `SubmissionService`](https://github.com/DSpace/dspace-angular/blob/959134fc89a58db9c0def5d7b8b22017fdc66024/src/app/submission/submission.service.ts#L323) is supposed to tell you wether the current submission is working with a workspace or a workflow item.

The way it determines this is by [looking at the route in angular](https://github.com/DSpace/dspace-angular/blob/959134fc89a58db9c0def5d7b8b22017fdc66024/src/app/submission/submission.service.ts#L325), whether it contains workspace or workflow. That is not robust at all, and won't work in a situation where you need to know the type of a submission object that's not the subject of the current submission.

PR #508, opened 2019-11-07
Refactor Submission Parsers

artlowel:
This PR adds the submission object (i.e. workspace or workflow item) id as a parameter to the fieldparsers. 

It also injects the fieldparsers instead of manually creating them. Though they're not singleton services, they're [still instantiated once for each field that needs to be rendered](https://github.com/atmire/dspace-angular/blob/refactor-submission-parsers/src/app/shared/form/builder/parsers/row-parser.ts#L77). Working with injection is a good idea in general in angular, because it makes things a lot easier test and mock, and it allows fieldparsers to use additional services when needed.

Finally it [adds a service](https://github.com/atmire/dspace-angular/blob/refactor-submission-parsers/src/app/core/submission/submission-object-data.service.ts) that allows you to fetch a submission object based on its id, without knowing what kind of object it is. 

These changes are necessary because without this PR, a fieldparser has only the info it has been provided by the rowparser, no context about the rest of the submission. This is problematic for some of the submission work currently in progress where we need be able to render fields with more complex behaviors that need to know about other parts of the item.

This PR is a first step towards a design for the submission where each field gets the entire submission object and knows which parts of it to use, rather than the way it works now, where the exact information needs to be spoon fed to the fieldparser. However it doesn't change anything about the way existing fields are rendered yet, because that would be a massive job, prone to introduce errors. All it does is add the ability to use services when rendering fields which, combined with the submission id and the new SubmissionObjectDataService, will allow us to write new fieldparsers that can decide for themselves which info they need.

I also created a ticket for an issue with the submission I didn't have the time to look in to for this PR #507

### DIFF
--- a/src/app/core/submission/submission-object-data.service.spec.ts
+++ b/src/app/core/submission/submission-object-data.service.spec.ts
@@ -1,0 +1,95 @@
+import { Observable } from 'rxjs';
+import { SubmissionService } from '../../submission/submission.service';
+import { RemoteData } from '../data/remote-data';
+import { SubmissionObject } from './models/submission-object.model';
+import { WorkspaceItem } from './models/workspaceitem.model';
+import { SubmissionObjectDataService } from './submission-object-data.service';
+import { SubmissionScopeType } from './submission-scope-type';
+import { WorkflowItemDataService } from './workflowitem-data.service';
+import { WorkspaceitemDataService } from './workspaceitem-data.service';
+
+describe('SubmissionObjectDataService', () => {
+  let service: SubmissionObjectDataService;
+  let submissionService: SubmissionService;
+  let workspaceitemDataService: WorkspaceitemDataService;
+  let workflowItemDataService: WorkflowItemDataService;
+
+  const submissionId = '1234';
+  const wsiResult = 'wsiResult' as any;
+  const wfiResult = 'wfiResult' as any;
+
+  beforeEach(() => {
+    workspaceitemDataService = jasmine.createSpyObj('WorkspaceitemDataService', {
+      findById: wsiResult
+    });
+    workflowItemDataService  = jasmine.createSpyObj('WorkflowItemDataService', {
+      findById: wfiResult
+    });
+  });
+
+  describe('findById', () => {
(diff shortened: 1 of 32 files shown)


paulo-graca:
Hi @artlowel ! I'm looking at submission form and something strange occurs. 
I've several collections with a submission form associated with each.
Just after login when I try to create a new Item I'm getting the error:

```
client.js:41115 ERROR TypeError: Cannot add property searchParams, object is not extensible
    at CollectionDataService../src/app/core/data/collection-data.service.ts.CollectionDataService.getAuthorizedCollectionByCommunity (client.js:214142)
    at MergeMapSubscriber.project (1.client.js:490)
    at MergeMapSubscriber../node_modules/rxjs/_esm5/internal/operators/mergeMap.js.MergeMapSubscriber._tryNext (client.js:178182)
    at MergeMapSubscriber../node_modules/rxjs/_esm5/internal/operators/mergeMap.js.MergeMapSubscriber._next (client.js:178172)
    at MergeMapSubscriber../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.next (client.js:172012)
    at MergeMapSubscriber../node_modules/rxjs/_esm5/internal/operators/mergeMap.js.MergeMapSubscriber.notifyNext (client.js:178201)
    at InnerSubscriber../node_modules/rxjs/_esm5/internal/InnerSubscriber.js.InnerSubscriber._next (client.js:171239)
    at InnerSubscriber../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.next (client.js:172012)
    at client.js:183298
    at subscribeToResult (client.js:183428)
```
and I can't change the collection. But just after it, if I access the option "registries>metadata" and after loading it, went back to "new>item" then the error no longer occurs and I'm able the change the collection. Does this PR addresses that issue? Do you want me to reporting it or is something well known?
Thanks in advance!

artlowel:
@paulo-graca That issue was introduced by https://github.com/DSpace/dspace-angular/pull/432, when the changes to the collection selector were added. I fixed it in [another PR](https://github.com/DSpace/dspace-angular/pull/531/commits/f72343bd8e5823fbd737e840d0f4e395412d7583)

Issue #510, opened 2019-11-12
Redirecting after login breaks query parameters 


LotteHofstede:
### Issue
When a user needs to authenticate to visit a certain page, they are redirected to the login page.
When they log in, they are redirected to the page they were trying to visit before the initial redirect. Unfortunately, during this second redirect the URL gets encoded, which breaks the redirect when, for example, there are query parameters in the path.
 
### Example
When I try to visit `/mydspace?configuration=workspace` as an unauthenticated user, I'm redirected to the login page. After logging in, I am redirected to `/mydspace%3Fconfiguration%3Dworkspace`, which renders a 404 page.

This issue seems to be related to issue #451 / PR #467

mspalti:
Thanks for describing this issue @LotteHofstede . I think the problem is that we didn't check the redirect with query parameters. The `router.navigate()` method adds encoding; and the solution is to use `navigateByUrl()` instead.  I'll fix this.

PR #511, opened 2019-11-12
Prevent encoding of query after login redirect.

mspalti:
Fixes https://github.com/DSpace/dspace-angular/issues/510


mspalti:
This minor revision changes the method used for redirects after login to be `router.navigateByUrl(url)`, rather than `router.navigate([url])`. 

When constructing the url, the latter was encoding query parameters. This had the effect of breaking redirects like `/mydspace?configuration=workspace`.

PR #525, opened 2019-11-22
Portuguese Translation

vitorsilverio:


bram-atmire:
@vitorsilverio looking good, the json5 validator didn't give any mistakes for me. Can you also add a change to this config file, to add/enable Portuguese in the UI:
https://github.com/DSpace/dspace-angular/blob/master/config/environment.default.js#L128

vitorsilverio:
@bram-atmire Done

PR #531, opened 2019-11-26
 Create relationships during the submission

artlowel:
This PR adds the ability to create relationships between items during the submission. 

It depends on #508

### Description
It will add a lookup button next to fields that have a `selectableRelationship` section in the submission form.

Clicking that button will open a modal that allows you to search for related items using the relationship, filter and search configuration specified in the selectableRelationship section.

The search results will have a checkbox or a radio next to them based on if you can have one or multiple relationships of that type.

Selecting a result will immediately create a relationship with the item being submitted (meaning you don't need to close the popup first).

You can see an overview of the current relationships in the current selection tab.

If name variants are enabled for the selectableRelationship. The name field of the search results will be comboboxes, they contain any name variants that are saved in the item metadata (currently always in the field `dc.title.alternative` but that will likely change and depend on the entity type in the future).

You can enter a new name variant in the input section of the combobox. If you are an admin (in the future this will check for edit rights on the related item, but currently there's no way to do that) you'll see a popup after that gives you the option to store it to the related entity.

### Still to do for this PR:
- ~~Bug: After a merge right before we opened this PR, selected relationships no longer show up in te submission, only in the modal~~
- ~~Fix linting issues~~
- ~~Write more tests and docs~~

### Features not included in this PR:
- The ability to mix relationships with plain metadata values for the same field
- A search as you type dropdown on the submission field inputs 
- The ability to reorder related entities in the submission
- Pagination for selected related entities in the submission.


artlowel:
@paulo-graca yes it is. I'll add the label

src/app/core/core.module.ts
```diff
@@ -122,6 +122,27 @@ import { BrowseDefinition } from './shared/browse-definition.model';
...
+} from './dspace-rest-v2/mocks/mock-response-map';
+import { EndpointMockingRestService } from './dspace-rest-v2/endpoint-mocking-rest.service';
+import { ENV_CONFIG, GLOBAL_CONFIG, GlobalConfig } from '../../config';
+import { SearchFilterService } from './shared/search/search-filter.service';
+import { SearchConfigurationService } from './shared/search/search-configuration.service';
+import { SelectableListService } from '../shared/object-list/selectable-list/selectable-list.service';
+import { RelationshipTypeService } from './data/relationship-type.service';
+import { SidebarService } from '../shared/sidebar/sidebar.service';
+
+export const restServiceFactory = (cfg: GlobalConfig, mocks: MockResponseMap, http: HttpClient) => {
```
tdonohue: Could you add a comment around this export to describe it.  Why is this using an `EndpointMockingRestService` when not in production mode?

src/app/core/data/relationship.service.ts
```diff
@@ -65,76 +71,78 @@ export class RelationshipService extends DataService<Relationship> {
...
-  /**
-   * Get a combined observable containing an array of all relationships in an item, as well as an array of the relationships their types
-   * This is used for easier access of a relationship's type because they exist as observables
-   * @param item
-   */
-  getItemResolvedRelsAndTypes(item: Item): Observable<[Relationship[], RelationshipType[]]> {
-    return observableCombineLatest(
-      this.getItemRelationshipsArray(item),
-      this.getItemRelationshipTypesArray(item)
+  addRelationship(typeId: string, item1: Item, item2: Item, leftwardValue?: string, rightwardValue?: string): Observable<RestResponse> {
```
tdonohue: Please add TypeDocs

src/app/core/data/relationship.service.ts
```diff
@@ -254,20 +228,113 @@ export class RelationshipService extends DataService<Relationship> {
...
-      this.objectCache.remove(rightItem.payload.self);
-      this.requestService.removeByHrefSubstring(leftItem.payload.self);
-      this.requestService.removeByHrefSubstring(rightItem.payload.self);
-    });
+      map((itemRD: RemoteData<Item>) => itemRD.payload),
+      map((item: Item) => uuids.includes(item.uuid))
+    );
+  }
+
+  getRelationshipByItemsAndLabel(item1: Item, item2: Item, label: string): Observable<Relationship> {
```
tdonohue: Please add Typedocs

src/app/core/data/relationship.service.ts
```diff
@@ -254,20 +228,113 @@ export class RelationshipService extends DataService<Relationship> {
...
+
+  private isItemMatchWithItemRD(itemRD$: Observable<RemoteData<Item>>, itemCheck: Item): Observable<boolean> {
+    return itemRD$.pipe(
+      getSucceededRemoteData(),
+      map((itemRD: RemoteData<Item>) => itemRD.payload),
+      map((item: Item) => item.uuid === itemCheck.uuid)
+    );
+  }
+
+  public setNameVariant(listID: string, itemID: string, nameVariant: string) {
```
tdonohue: Please add Typedocs or a basic comment describing each of these public methods (there are 5 more below this one).   Most of these are pretty straightforward by name, but it's immediately clear what `listID` refers to in this method.

src/app/core/dspace-rest-v2/endpoint-mocking-rest.service.ts
```diff
@@ -0,0 +1,109 @@
...
+
+import { DSpaceRESTV2Response } from './dspace-rest-v2-response.model';
+import { DSpaceRESTv2Service, HttpOptions } from './dspace-rest-v2.service';
+import { MOCK_RESPONSE_MAP, MockResponseMap } from './mocks/mock-response-map';
+import * as URL from 'url-parse';
+
+/**
+ * Service to access DSpace's REST API.
+ *
+ * If a URL is found in this.mockResponseMap, it returns the mock response instead
```
tdonohue: I'm not sure I understand this service. Could you describe why/when it would be used? If this is only for Specs, we should say that in the description here.

Also, I wonder if this should be moved to the `/mocks` subfolder along with the rest of the related files?

src/app/core/services/route.service.ts
```diff
@@ -183,4 +186,23 @@ export class RouteService {
...
+
+  public addParameter(key, value) {
+    this.store.dispatch(new AddParameterAction(key, value));
+  }
+
+  public setParameter(key, value) {
+    this.store.dispatch(new SetParameterAction(key, value));
+  }
+
+  public setCurrentRouteInfo() {
```
tdonohue: Please describe this method with TypeDocs or similar

src/app/core/utilities/equals.decorators.ts
```diff
@@ -0,0 +1,58 @@
+import { isEmpty } from '../../shared/empty.util';
+import { GenericConstructor } from '../shared/generic-constructor';
+import { EquatableObject } from './equatable';
+
+const excludedFromEquals = new Map();
+const fieldsForEqualsMap = new Map();
+
+export function inheritEquatable(parentCo: GenericConstructor<EquatableObject<any>>) {
```
tdonohue: Please add Typedocs to this function

src/app/core/utilities/equals.decorators.ts
```diff
@@ -0,0 +1,58 @@
...
+    const parentMappedFields = fieldsForEqualsMap.get(parentCo) || new Map();
+    Array.from(parentMappedFields.keys())
+      .filter((key) => !Array.from(mappedFields.keys()).includes(key))
+      .forEach((key) => {
+        fieldsForEquals(...parentMappedFields.get(key))(new childCo(), key);
+      });
+  }
+}
+
+export function excludeFromEquals(object: any, propertyName: string): any {
```
tdonohue: Please add Typedocs to this function.  It looks like this relates to the `@excludeFromEquals` annotations I'm seeing elsewhere, right?

src/app/core/utilities/equals.decorators.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+  excludedFromEquals.set(object.constructor, [...list, propertyName]);
+}
+
+// tslint:disable-next-line:ban-types
+export function getExcludedFromEqualsFor(constructor: Function): string[] {
+  return excludedFromEquals.get(constructor) || [];
+}
+
+export function fieldsForEquals(...fields: string[]): any {
```
tdonohue: Again, Typedocs please

src/app/core/utilities/equatable.ts
```diff
@@ -0,0 +1,42 @@
+import { getExcludedFromEqualsFor, getFieldsForEquals } from './equals.decorators';
+import { hasNoValue, hasValue } from '../../shared/empty.util';
+
+function equalsByFields(object1, object2, fieldList): boolean {
```
tdonohue: Again Typedocs or comment please

src/app/core/utilities/equatable.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+    if (typeof object1[key] === 'object') {
+      return !equalsByFields(object1[key], object2[key], Object.keys(object1))
+    }
+    return object1[key] !== object2[key];
+  });
+  return hasNoValue(unequalProperty);
+}
+
+export abstract class EquatableObject<T> {
```
tdonohue: Please add Typedocs for this abstract class

src/app/entity-groups/research-entities/submission/item-list-elements/org-unit/org-unit-search-result-list-submission-element.component.ts
```diff
@@ -0,0 +1,98 @@
...
+@Component({
+  selector: 'ds-person-search-result-list-submission-element',
+  styleUrls: ['./org-unit-search-result-list-submission-element.component.scss'],
+  templateUrl: './org-unit-search-result-list-submission-element.component.html'
+})
+
+/**
+ * The component for displaying a list element for an item search result of the type Person
+ */
+export class OrgUnitSearchResultListSubmissionElementComponent extends SearchResultListElementComponent<ItemSearchResult, Item> implements OnInit {
```
tdonohue: TypeDocs here are wrong :)  This looks to be OrgUnit search results

src/app/entity-groups/research-entities/submission/name-variant-modal/name-variant-modal.component.ts
```diff
@@ -0,0 +1,14 @@
+import { Component, Input } from '@angular/core';
+import { NgbActiveModal } from '@ng-bootstrap/ng-bootstrap';
+
+@Component({
+  selector: 'ds-name-variant-modal',
+  templateUrl: './name-variant-modal.component.html',
+  styleUrls: ['./name-variant-modal.component.scss']
+})
+export class NameVariantModalComponent {
```
tdonohue: Please add basic TypeDocs or comments to describe what a NameVariantModal is.

src/app/shared/form/builder/ds-dynamic-form-ui/ds-dynamic-form-control-container.component.ts
```diff
@@ -150,32 +184,95 @@ export class DsDynamicFormControlContainerComponent extends DynamicFormControlCo
...
+    private relationshipService: RelationshipService,
+    private zone: NgZone,
+    private store: Store<AppState>,
+    private submissionObjectService: SubmissionObjectDataService
   ) {
-
     super(componentFactoryResolver, layoutService, validationService);
   }
 
+  ngOnInit(): void {
```
tdonohue: Could we describe this ngOnInit() better?  Either via TypeDocs or maybe inline comments.  It's pretty complex, and it'd be worth describing what it is doing when it finds a relationship.

src/app/shared/form/builder/ds-dynamic-form-ui/ds-dynamic-form-control-container.component.ts
```diff
@@ -212,4 +309,38 @@ export class DsDynamicFormControlContainerComponent extends DynamicFormControlCo
       this.onChange(event);
     }
   }
+
+  public hasResultsSelected(): Observable<boolean> {
+    return this.model.value.pipe(map((list: Array<SearchResult<DSpaceObject>>) => isNotEmpty(list)));
+  }
+
+  openLookup() {
```
tdonohue: Again a brief description or TypeDocs.   Also see `removeSelection` just below this.

src/app/shared/form/builder/ds-dynamic-form-ui/models/disabled/dynamic-disabled.component.ts
```diff
@@ -0,0 +1,29 @@
...
+import { DynamicFormControlComponent, DynamicFormLayoutService, DynamicFormValidationService } from '@ng-dynamic-forms/core';
+import { FormGroup } from '@angular/forms';
+import { DynamicDisabledModel } from './dynamic-disabled.model';
+import { RelationshipTypeService } from '../../../../../../core/data/relationship-type.service';
+
+@Component({
+  selector: 'ds-dynamic-disabled',
+  templateUrl: './dynamic-disabled.component.html'
+})
+export class DsDynamicDisabledComponent extends DynamicFormControlComponent {
```
tdonohue: Please add TypeDocs

src/app/shared/form/builder/ds-dynamic-form-ui/models/disabled/dynamic-disabled.model.ts
```diff
@@ -0,0 +1,22 @@
+import { DynamicFormControlLayout, serializable } from '@ng-dynamic-forms/core';
+import { DsDynamicInputModel, DsDynamicInputModelConfig } from '../ds-dynamic-input.model';
+
+export const DYNAMIC_FORM_CONTROL_TYPE_DISABLED = 'EMPTY';
+
+export interface DsDynamicDisabledModelConfig extends DsDynamicInputModelConfig {
+  value?: any;
+}
+
+export class DynamicDisabledModel extends DsDynamicInputModel {
```
tdonohue: Please add TypeDocs

src/app/shared/form/builder/ds-dynamic-form-ui/relation-lookup-modal/dynamic-lookup-relation-modal.component.ts
```diff
@@ -0,0 +1,154 @@
...
+  templateUrl: './dynamic-lookup-relation-modal.component.html',
+  providers: [
+    {
+      provide: SEARCH_CONFIG_SERVICE,
+      useClass: SearchConfigurationService
+    }
+  ]
+})
+
+export class DsDynamicLookupRelationModalComponent implements OnInit, OnDestroy {
```
tdonohue: Please add TypeDocs

src/app/shared/form/builder/ds-dynamic-form-ui/relation-lookup-modal/name-variant.actions.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Item } from '../../../../../core/shared/item.model';
+
+export const NameVariantActionTypes = {
+  SET_NAME_VARIANT: type('dspace/name-variant/SET_NAME_VARIANT'),
+  REMOVE_NAME_VARIANT: type('dspace/name-variant/REMOVE_NAME_VARIANT'),
+};
+
+/* tslint:disable:max-classes-per-file */
+
+export abstract class NameVariantListAction implements Action {
```
tdonohue: Please add TypeDocs for the three classes in this file

src/app/shared/form/builder/ds-dynamic-form-ui/relation-lookup-modal/relationship.effects.ts
```diff
@@ -0,0 +1,129 @@
...
+              }
+            )
+          } else {
+            this.debounceMap[identifier].next(action.type);
+          }
+        }
+      )
+    );
+
+  @Effect({ dispatch: false }) updateNameVariantsActions$ = this.actions$
```
tdonohue: Please add TypeDocs or description

src/app/shared/form/builder/ds-dynamic-form-ui/relation-lookup-modal/search-tab/dynamic-lookup-relation-search-tab.component.ts
```diff
@@ -0,0 +1,158 @@
...
+  templateUrl: './dynamic-lookup-relation-search-tab.component.html',
+  providers: [
+    {
+      provide: SEARCH_CONFIG_SERVICE,
+      useClass: SearchConfigurationService
+    }
+  ]
+})
+
+export class DsDynamicLookupRelationSearchTabComponent implements OnInit, OnDestroy {
```
tdonohue: Please add TypeDocs to this class and its methods

src/app/shared/form/builder/ds-dynamic-form-ui/relation-lookup-modal/selection-tab/dynamic-lookup-relation-selection-tab.component.ts
```diff
@@ -0,0 +1,78 @@
...
+  templateUrl: './dynamic-lookup-relation-selection-tab.component.html',
+  providers: [
+    {
+      provide: SEARCH_CONFIG_SERVICE,
+      useClass: SearchConfigurationService
+    }
+  ]
+})
+
+export class DsDynamicLookupRelationSelectionTabComponent {
```
tdonohue: Please add Typedocs to this class and its methods

src/app/shared/form/builder/models/relationship-options.model.ts
```diff
@@ -0,0 +1,12 @@
+const RELATION_METADATA_PREFIX = 'relation.'
+
+export class RelationshipOptions {
```
tdonohue: Please add TypeDocs

src/app/shared/form/builder/parsers/disabled-field-parser.ts
```diff
@@ -0,0 +1,11 @@
+import { FieldParser } from './field-parser';
+import { FormFieldMetadataValueObject } from '../models/form-field-metadata-value.model';
+import { DsDynamicDisabledModelConfig, DynamicDisabledModel } from '../ds-dynamic-form-ui/models/disabled/dynamic-disabled.model';
+
+export class DisabledFieldParser extends FieldParser {
```
tdonohue: Please add Typedocs

src/app/shared/form/builder/parsers/parser-factory.ts
```diff
@@ -12,44 +18,102 @@ import { NameFieldParser } from './name-field-parser';
...
+const fieldParserDeps = [
+  SUBMISSION_ID,
+  CONFIG_DATA,
+  INIT_FORM_VALUES,
+  PARSER_OPTIONS,
+];
 
 export class ParserFactory {
-  public static getConstructor(type: ParserType): GenericConstructor<FieldParser> {
+  public static getProvider(type: ParserType): StaticProvider {
```
tdonohue: Could you add some basic TypeDocs to this class while we are in here?

src/app/shared/form/builder/parsers/row-parser.ts
```diff
@@ -1,62 +1,87 @@
...
 export class RowParser {
-  protected authorityOptions: IntegrationSearchOptions;
-
-  constructor(protected rowData,
-              protected scopeUUID,
-              protected initFormValues: any,
-              protected submissionScope,
-              protected readOnly: boolean) {
-    this.authorityOptions = new IntegrationSearchOptions(scopeUUID);
+  constructor(private parentInjector: Injector) {
```
tdonohue: Again, it'd be nice to add basic TypeDocs here while editing the class. I realize you didn't create this class though.

src/app/shared/object-list/selectable-list/selectable-list.actions.ts
```diff
@@ -0,0 +1,84 @@
...
+export abstract class SelectableListAction implements Action {
+  // tslint:disable-next-line:no-shadowed-variable
+  constructor(public type, public id: string) {
+  }
+}
+
+/**
+ * Used to select an item in a the selectable list
+ */
+export class SelectableListSelectAction extends SelectableListAction {
```
tdonohue: Could you similarly add comments for all the Action classes in this file?  See the ones below

src/app/shared/object-list/selectable-list/selectable-list.reducer.ts
```diff
@@ -0,0 +1,102 @@
...
+      const newListState = clearSelection(action.id);
+      return Object.assign({}, state, { [action.id]: newListState });
+    }
+    default: {
+      return state;
+    }
+  }
+}
+
+function select(state: SelectableListState, action: SelectableListSelectAction) {
```
tdonohue: Please add a brief comment describing each function in this reducer file.

tdonohue:
I gave this a test today, and it appears to mostly work. However, I did run into a few issues:

1. First, I can reproduce the bug @artlowel already noted.  After creating relations, those relations do not appear in the submission form. E.g. if I click "Lookup" next to "Author", and select some Authors, they do not appear on the form after the popup closes. However, if I click "Lookup" again, I see my already selected Authors.
2. The Search for Authors (if I click "Lookup" next to the "Author" field) does NOT appear to be limited to Author objects.  In the results list, I also see Publications, Project, OrgUnit, etc.  You can see those by simply expanding the "Item Type" facet in the sidebar of the popup modal.   (I'm just using the default submission form configuration on my local instance)

Finally, is there a good backend that this can be tested against?  In my local search results a lot of Authors have name variants that say "undefined, undefined".  I suspect this is just outdated/bad data on my local install (which was based on an early/old Postgres dump of entities used during the Preview Release) -- I think this is similar to the entities data issues on our REST API demo site.  I think we need to update that test data and/or have a better test server....otherwise, properly testing this is a bit more difficult.

artlowel:
>The Search for Authors (if I click "Lookup" next to the "Author" field) does NOT appear to be limited to Author objects. In the results list, I also see Publications, Project, OrgUnit, etc. You can see those by simply expanding the "Item Type" facet in the sidebar of the popup modal. (I'm just using the default submission form configuration on my local instance)

That is due to the server config. The selectableRelationships section in the submission config contains a search configuration to use in that modal. That search configuration needs to exist and filter on the appropriate entity types 

>Finally, is there a good backend that this can be tested against? In my local search results a lot of Authors have name variants that say "undefined, undefined". I suspect this is just outdated/bad data on my local install (which was based on an early/old Postgres dump of entities used during the Preview Release) -- I think this is similar to the entities data issues on our REST API demo site. I think we need to update that test data and/or have a better test server....otherwise, properly testing this is a bit more difficult.

Those are indeed the same issues we had with the demo site. Ask @benbosman for a database export in the meeting today, he'll give you one that's properly configured

benbosman:
@tdonohue I noticed there was a config error causing difficulties to test this.
https://github.com/DSpace/DSpace/pull/2618 should solve that problem

paulo-graca:
> 1. First, I can reproduce the bug @artlowel already noted.  After creating relations, those relations do not appear in the submission form. E.g. if I click "Lookup" next to "Author", and select some Authors, they do not appear on the form after the popup closes. However, if I click "Lookup" again, I see my already selected Authors.

I had the same doubt before https://github.com/DSpace/DSpace/pull/2608 and I spoke with @benbosman and fix that on this PR.

src/app/shared/form/builder/ds-dynamic-form-ui/relation-lookup-modal/search-tab/dynamic-lookup-relation-search-tab.component.ts
LotteHofstede: Added typedoc

src/app/core/core.module.ts
LotteHofstede: Added some comments for clarification

src/app/core/dspace-rest-v2/endpoint-mocking-rest.service.ts
LotteHofstede: Added some documentation on its usage. Also moved it to the `/mock` subfolder.

src/app/shared/mocks/dspace-rest-v2/mocks/mock-response-map.ts
```diff
@@ -0,0 +1,10 @@
+import { InjectionToken } from '@angular/core';
+import mockSubmissionResponse from './mock-submission-response.json';
+
+export class MockResponseMap extends Map<string, any> {};
+
+export const MOCK_RESPONSE_MAP: InjectionToken<MockResponseMap> = new InjectionToken<MockResponseMap>('mockResponseMap');
+
+export const mockResponseMap: MockResponseMap = new Map([
+  // [ '/config/submissionforms/traditionalpageone', mockSubmissionResponse ]
```
tdonohue: Is this meant to be commented out as an example? 

It'd be nice to add comments or Typedocs to this file to explain how this `MockResponseMap` works.  It's not self explanatory right now, and the commented out code looks a bit odd here.  However, I'm OK with this commented out example if it's explained via TypeDocs or similar.

src/app/shared/object-collection/shared/selectable-list-item-control/selectable-list-item-control.component.ts
```diff
@@ -0,0 +1,78 @@
...
+
+@Component({
+  selector: 'ds-selectable-list-item-control',
+  // styleUrls: ['./selectable-list-item-control.component.scss'],
+  templateUrl: './selectable-list-item-control.component.html'
+})
+/**
+ * Component for determining what component to use depending on the item's relationship type (relationship.type)
+ */
+export class SelectableListItemControlComponent implements OnInit {
```
tdonohue: TypeDocs are incorrect here & need updating

src/app/shared/object-list/object-list.component.ts
```diff
@@ -54,6 +58,11 @@ export class ObjectListComponent {
    * Whether or not the pager is visible when there is only a single page of results
    */
   @Input() hidePagerWhenSinglePage = true;
+  @Input() selectable = false;
+  @Input() selectionConfig: { repeatable: boolean, listId: string };
+  // @Input() previousSelection: ListableObject[] = [];
+  // allSelected = false;
+  // selectAllLoading = false;
```
tdonohue: I'm not sure why these 3 commented lines are here? Are they a mistake?

src/app/shared/object-list/selectable-list/selectable-list.actions.ts
```diff
@@ -0,0 +1,101 @@
...
+  constructor(id: string, object: ListableObject) {
+    super(SelectableListActionTypes.DESELECT_SINGLE, id);
+    this.payload = object;
+  }
+}
+
+/**
+ * Action to deselect a single object in a the selectable list
+ */
+export class SelectableListDeselectAction extends SelectableListAction {
```
tdonohue: Tiny thing, but the TypeDocs here seem to be for `SelectableListDeselectSingleAction` (just above this one). I think you accidentally swapped the TypeDocs here.

src/app/shared/page-size-selector/page-size-selector.component.ts
```diff
@@ -0,0 +1,52 @@
...
+@Component({
+  selector: 'ds-page-size-selector',
+  styleUrls: ['./page-size-selector.component.scss'],
+  templateUrl: './page-size-selector.component.html'
+})
+
+/**
+ * This component represents the part of the search sidebar that contains the general search settings.
+ */
+export class PageSizeSelectorComponent implements OnInit {
```
tdonohue: TypeDocs are incorrect here & need updating

src/app/shared/utils/route.utils.ts
```diff
@@ -0,0 +1,6 @@
+import { Router } from '@angular/router';
+
+export function currentPath(router: Router) {
```
tdonohue: Please add TypeDocs to this function.  It's a simple function, but it'd be good to still note exactly that it is doing.

src/app/submission/sections/form/section-form.component.ts
```diff
@@ -144,16 +154,20 @@ export class SubmissionSectionformComponent extends SectionModelComponent {
...
+          this.sectionService.getSectionData(this.submissionId, this.sectionData.id),
+          this.workspaceItemDataService.findById(this.submissionId).pipe(getSucceededRemoteData(), map((wsiRD: RemoteData<WorkspaceItem>) => wsiRD.payload))
+        )),
       take(1))
-      .subscribe((sectionData: WorkspaceitemSectionFormObject) => {
+      .subscribe(([sectionData, workspaceItem]: [WorkspaceitemSectionFormObject, WorkspaceItem]) => {
         if (isUndefined(this.formModel)) {
           this.sectionData.errors = [];
           // Is the first loading so init form
+          this.workspaceItem = workspaceItem;
```
tdonohue: Tiny thing, but you put this new line right after a comment that refers to the *next* line (the initForm call).  So, the comment above should be moved below.

src/app/core/core.module.ts
LotteHofstede: I think there might be a small misunderstanding. This feature doesn't have anything to do with tests. While developing, it often happens a REST endpoint necessary for the frontend is broken, unfinished or returns something unexpected. This service makes it very simple to add a mock REST response instead, so the frontend developer can simply add their expected response there, without having to worry about the actual response from the REST server. 
Of course this should never be used in production, so that's why it's disabled in production mode.

config/environment.default.js
```diff
@@ -9,11 +9,11 @@ module.exports = {
...
   // The REST API server settings.
   rest: {
-    ssl: true,
+      ssl: true,
     host: 'dspace7.4science.cloud',
-    port: 443,
+      port: 443,
     // NOTE: Space is capitalized because 'namespace' is a reserved string in TypeScript
-    nameSpace: '/server/api'
+      nameSpace: '/server/api'
```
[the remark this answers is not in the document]
LotteHofstede: Answering a few of your issues, the rest will be covered by @artlowel:

2. Fixed this issue, it might have been introduced by a merge conflict
3. This issue was already fixed in [a later PR](https://github.com/DSpace/dspace-angular/pull/541), where we changed a significant part of the way the values are rendered. 
4. Same as 3.

LotteHofstede:
@tdonohue 
Answering a few of your issues, the rest will be covered by @artlowel:

Fixed this issue, it might have been introduced by a merge conflict
This issue was already fixed in a later PR, where we changed a significant part of the way the values are rendered.
Same as 3.

artlowel:
@tdonohue 

> 1. In the popup search modal, pagination is slow. I'm using the test data provided by Ben, which has 68 authors. If I click on page "6" it takes about 3-4 seconds before the data from that page appears. It's possible my backend is just slow, but wanted to see if this is reproducible.
In Chrome Dev Tools, it looks like my backend is speedy enough, but each time I click a new page, for each item returned I see it querying the /bundles, /relationships and /owningCollection paths (so 3 separate REST requests per Item). So, it looks like it might be gathering more info than is necessary.

This is due to:
1. The fact that currently when the RemoteDataBuildService constructs an object, it will greedily go over all its relationships and retrieve any that aren't already cached, whether they're needed or not. I've been working on that for a while now, and will create a PR to solve that issue early next year
2. The fact that the submission was designed not to cache anything at all. forceByPassCache was [introduced in the submission PR](https://github.com/DSpace/dspace-angular/pull/279#discussion_r248637395), and later [when we tried to remove it](https://github.com/DSpace/dspace-angular/pull/468), it became clear that the submission just wouldn't work with the cache without major changes, so the solution then was to [use a cache time of 0](https://github.com/DSpace/dspace-angular/pull/468#issuecomment-531934206) for submission requests for the time being. As a consequence, every rest resource used for the modal needs to be fetched from the backend _every_ time. We'll likely need to address this. But perhaps we can wait and see if my solution for 1. improves performance by enough, because it doesn't look like it's easy to fix in the submission.

> 6. In the modal, the "Sort By" option does not appear to work. It defaults to "Relevance" (which seems to be a random order for Authors -- I would expect them to be ordered alphabetically). If I change to "Title Ascending" or Descending, I get 400 errors in Chrome Dev Tools.

This appears to be an issue with search configurations everywhere. If you try it on [the regular search page, with the same configuration](https://dspace7-demo.atmire.com/search?configuration=person) you have the same problem. The reason is that the sort options were hardcoded in the UI at first, but never updated after it became possible to get the available sort options from rest. I've created an issue to track it: #549 

> 7. Not a huge deal, but we may want to improve the usability of the select/deselect all options in this popup modal. I accidentally clicked it for 68 authors, and all 68 author relationships were created (so it works!), but I'm not sure that's something we'd want to have happen if your institution has thousands of authors. It should likely either warn "are you sure?" or maybe not even appear on this popup (though keeping "select/deselect page" seems fine).

Yeah I noticed the same things, but I was thinking more along the lines of: don't immediately create all relationships, but wait either a set amount of time (a few seconds), or until you close the modal to create them. In any case we'd prefer to tackle this in a separate, dedicated PR later on.

paulo-graca:
These is more an usability and expectancy things that I think can be addressed in later and smaller PRs...
when I type anything in the form, I was expecting the typed valued could be used to filter the results in the model:
image

and the introduced value is ignored:
image

Another thing is the "Current selection" tab I think it should be distinguished from the other normal tabs, otherwise, the user will ignore it:
image

And the "Close" option isn't obvious to the user that the process of item association is ended, because there isn't any visual feedback and if the related field is mandatory the submission can't be finished:
image
(mandatory field validation should also include relations verification)

config/environment.default.js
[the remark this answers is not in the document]
LotteHofstede: Reverted indenting

src/app/+search-page/search.component.html
```diff
@@ -46,5 +46,9 @@
                   [scopes]="(scopeListRD$ | async)"
                   [inPlaceSearch]="inPlaceSearch">
   </ds-search-form>
+    <div class="row mb-3 mb-md-1">
+        <div class="labels col-sm-9 offset-sm-3">
   <ds-search-labels *ngIf="searchEnabled" [inPlaceSearch]="inPlaceSearch"></ds-search-labels>
+        </div>
+    </div>
```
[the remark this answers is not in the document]
LotteHofstede: Fixed indentation

src/app/shared/form/builder/ds-dynamic-form-ui/relation-lookup-modal/selection-tab/dynamic-lookup-relation-selection-tab.component.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    fixture = TestBed.createComponent(DsDynamicLookupRelationSelectionTabComponent);
+    component = fixture.componentInstance;
+    component.selection$ = selection$;
+    component.listId = listID;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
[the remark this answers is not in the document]
LotteHofstede: Added some additional tests for this component

src/app/shared/mocks/dspace-rest-v2/mocks/mock-response-map.ts
LotteHofstede: I added some comments to explain its purpose.

src/app/shared/object-collection/shared/selectable-list-item-control/selectable-list-item-control.component.ts
LotteHofstede: Updated typedocs

src/app/shared/object-list/selectable-list/selectable-list.actions.ts
LotteHofstede: Switched the TypeDoc so it's correct.

src/app/shared/page-size-selector/page-size-selector.component.ts
LotteHofstede: Updates TypeDoc

src/app/shared/utils/route.utils.ts
LotteHofstede: Added TypeDoc

src/app/submission/sections/form/section-form.component.ts
LotteHofstede: Moved the comment to the right line.

src/app/shared/form/builder/ds-dynamic-form-ui/relation-lookup-modal/search-tab/dynamic-lookup-relation-search-tab.component.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      component.deselectPage([searchResult1, searchResult2, searchResult3]);
+    });
+
+    it('should emit the page filtered from not yet selected objects and call select on the service for all objects', () => {
+      expect(component.deselectObject.emit).toHaveBeenCalledWith(searchResult1, searchResult2);
+      expect(selectableListService.deselect).toHaveBeenCalledWith(listID, [searchResult1, searchResult2, searchResult3]);
+    });
+  });
+
+  describe('selectAll', () => {
```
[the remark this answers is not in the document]
LotteHofstede: The selectPage function takes an argument, selectAll does not. I updated the test so they differ in outcome, so the difference is more obvious.

src/app/shared/object-list/object-list.component.ts
LotteHofstede: This is just a remnant of how thing previously worked. I removed it.

src/app/core/shared/operators.ts
```diff
@@ -126,3 +125,9 @@ export const getFirstOccurrence = () =>
     source.pipe(
       map((rd) => Object.assign(rd, { payload: rd.payload.page.length > 0 ? rd.payload.page[0] : undefined }))
     );
+
+export const obsLog = (logString?: string) =>
```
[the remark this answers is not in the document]
LotteHofstede: I removed it, I only used it to test some things with observable values.

artlowel:
@tdonohue 
All of your feedback should have been addressed now. Could you re-review?

@paulo-graca 
Both of those are good points. However it's a lot less work for us to address them in #541, so we'll fix them there if you don't mind.

paulo-graca:
> @paulo-graca
> Both of those are good points. However it's a lot less work for us to address them in #541, so we'll fix them there if you don't mind

Ok, and since this PR is quite big in the number of changes. I also agree to have them addressed in a separate PR.

src/app/core/core.module.ts
artlowel: So this service allows you to store an example rest response to a json file, and then assign an URL to that json file. While in development mode, all calls to that URL will return the json from the file, all other calls will actually be sent to REST, allowing you to mock endpoints that don't exist or do something unexpected.

tdonohue: Thanks for that explanation again. I now see that I had misunderstood the purpose of `MockResponseMap` (which now has better TypeDocs to describe it): https://github.com/DSpace/dspace-angular/blob/b510894f6feddae1eeb0c778d13d423d5a56956f/src/app/shared/mocks/dspace-rest-v2/mocks/mock-response-map.ts

I see now that, by default, this `MockResponseMap` essentially "does nothing".  There are no responses listed by default to mock, so all responses will hit the REST API backend.  However, developers can use `MockResponseMap` to mock a REST response to use for development (as you said, so they can develop on the Angular side even if the REST implementation is not yet complete).

Now that I understand it, I like this idea.  Thanks!

tdonohue:
As this is now at +2 and has had a thorough code review & test by @paulo-graca and I (including several rounds of feedback over the last 3 weeks), I'm merging this as-is.  Thanks all for the hard work @artlowel , @LotteHofstede and @paulo-graca !

PR #532, opened 2019-11-29
Item page entities improvements

Atmire-Kristof:
This PR aims to further improve the changes made to item page entities in #475 by changing the "view more" link to load the next x elements instead of all at once, as the previous implementation could potentially load a lot of entities at once.

The label for "view more" has been changed to "Show x more" and "view less" is now "Hide last x".
The amount (x) defaults to 5 for related items and 10 for metadata representations, but are configured as inputs, so each individual list can easily be changed should you wish to do so.

Because both links can now be displayed at once for the same list, I've changed the layout to display the "show more" link on the bottom left side and the "hide last" link on the bottom right side.

Both metadata-representation-list and related-items component also display a loading component as (more) entities are loading. Previously, the list would completely disappear and re-appear, which could cause confusion if it takes a while to load.


Atmire-Kristof:
Update:

Refactored the code entirely to allow for the lists to load in groups of "pages".
Previously, the list size would just grow or shrink when clicking "view x more" or "hide last x", making a new REST call each time, fetching the enitre list of objects.
This has two negative effects:
- The list increasingly takes longer to expand, as the payload gets larger each time. This also means heavy load for the server.
- The list disappears entirely when loading new objects

As a solution, I made them load "by page". The list starts off by displaying 1 page of size `incrementBy`. When the user clicks "view x more", the component adds a new page of the same size to the list of pages.
This solves the issues:
- Increasing the list size takes the same amount of time to load, no matter how long the displayed list is, because it only loads the next page. The load for the server is the same each time.
- Pages already displayed will remain while the new page loads

Because related-items and metadata-representations both use the same strategy to display more/less items, I made a separate abstract component from which both can extend. This reduces code duplication and cleans the component's code.

tdonohue:
👍 Thank @Atmire-Kristof !  I've retested today & reviewed the changes. This looks good to me & seems to work well.

paulo-graca:
I'm getting angular errors with this PR that I don't get with PR531, like:
client.js:211978 Error: The server returned an invalid object
    at DSOResponseParsingService../src/app/core/data/base-response-parsing.service.ts.BaseResponseParsingService.addToObjectCache (client.js:213879)
    at DSOResponseParsingService../src/app/core/data/base-response-parsing.service.ts.BaseResponseParsingService.cache (client.js:213874)
    at DSOResponseParsingService../src/app/core/data/base-response-parsing.service.ts.BaseResponseParsingService.process (client.js:213818)
    at DSOResponseParsingService../src/app/core/data/dso-response-parsing.service.ts.DSOResponseParsingService.parse (client.js:215016)
    at MapSubscriber.project (client.js:217526)
    at MapSubscriber../node_modules/rxjs/_esm5/internal/operators/map.js.MapSubscriber._next (client.js:177921)
    at MapSubscriber../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.next (client.js:172016)
    at CatchSubscriber../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber._next (client.js:172039)
    at CatchSubscriber../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.next (client.js:172016)
    at MapSubscriber../node_modules/rxjs/_esm5/internal/operators/map.js.MapSubscriber._next (client.js:177927)

benbosman:
@paulo-graca can you let me know for which item on the https://dspace7-demo.atmire.com/server/ you're encountering https://github.com/DSpace/dspace-angular/pull/532#issuecomment-564014879?
Art also checked it out, and couldn't reproduce your error either.

paulo-graca:
It occur to me right on the homepage (using Google Chrome with and without Anonymous mode):
image and all subsequent pages I visited, even on the item page:
image

paulo-graca:
> It occur to me right on the homepage

I forgot to mention this tests are using the latest version of this PR.

artlowel:
@paulo-graca Thanks I was finally able to recreate the issue. It is not related to this PR.

It happens when you post a statistics event about the home page: the rest response will contain a property: `id: null`. This error will be fixed in the rest api, but the client should even try to parse the response, as we don't need it for anything.

This PR will fix the issue client side: https://github.com/DSpace/dspace-angular/pull/544
A REST PR will follow later.

paulo-graca:
> @paulo-graca Thanks I was finally able to recreate the issue. It is not related to this PR.
> 
> It happens when you post a statistics event about the home page: the rest response will contain a property: `id: null`. This error will be fixed in the rest api, but the client should even try to parse the response, as we don't need it for anything.
> 
> This PR will fix the issue client side: #544
> A REST PR will follow later.

Thank you @artlowel !

tdonohue:
Merging, as this is now at +3

PR #542, opened 2019-12-10
OpenAIRE4 funding agency missing translation

paulo-graca:
This PR just adds missing and required translations for OpenAIRE 4 guidelines support on the angular interface.


paulo-graca:
Thank you @artlowel !

PR #563, opened 2020-01-15
Fix history and bitstreamFormats store selector after moved in CoreState

atarix83:
This PR fixed bitstreamFormatsStateSelector and historySelector broken after latest upgrade to Angular 7


atarix83:
@artlowel 
problem resolved

PR #597, opened 2020-02-27
fix Submission cannot be completed #592

atarix83:
This PR resolves https://github.com/DSpace/dspace-angular/issues/592 and fixes other minor bug found.


paulo-graca:
Didn't had the chance to take a look before the Merge but just wanted to add the changes look ok to me.

PR #615, opened 2020-03-10
Add Administrative search

### LINKED ISSUE #258: Go To Item admin component
This task depends on #257. 

The goal is to recreate the Find By Item admin feature ([example](http://demo.dspace.org/xmlui/admin/item)). In order to avoid confusion with an upcoming administrative search feature, we'll call it "Go To Item", instead of "Find Item"

- create a page component for the feature
- configure a route for it at `/admin/content/go-to`
- Ensure the AdminGuard from #257 is used to guard everything under `/admin`
- Add a header, a label an input and a button. Use an [input group](https://getbootstrap.com/docs/4.1/components/input-group/#button-addons) to ensure the input and the button remain on the same line. Also, in md screens and wider, limit its width to 6 grid columns.
- on submit, determine whether it’s an id or a handle 
	- if it contains a `/` assume it's a handle, 
	- if not test whether it's a UUID using this regex: `[0-9a-fxA-FX]{8}-[0-9a-fxA-FX]{4}-[0-9a-fxA-FX]{4}-[0-9a-fxA-FX]{4}-[0-9a-fxA-FX]{12}`
	- If that fails [give the input an invalid status](https://getbootstrap.com/docs/4.1/components/forms/#server-side), with an error message explaining what went wrong.
- then fetch the object using the DSODataService.
	- If #256 hasn't finished yet the DSODataService won't exist. In that case add a temp method that will choose a hardcoded collection or item id, and use ItemDataService or CollectionDataService's findByID() method to fetch it
- If the object doesn't exist, or isn't an Item (using a handle you could retrieve any kind of DSpace Object), [give the input an invalid status](https://getbootstrap.com/docs/4.1/components/forms/#server-side), with an error message explaining what went wrong.
- If it does exist, and is an Item, redirect to its edit page
	- The edit page doesn't exist yet, but it will be located at `/item/:item-uuid/edit`, so simply ensure that it redirects to the correct place.

artlowel:
This PR adds an administrative search page. It depends on REST PR: https://github.com/DSpace/DSpace/pull/2706, and it replaces the "Find Item by Internal Item ID (UUID) or Handle", "Withdrawn Items (browse by Title)" and "Private Items (browse by Title)" in DSpace 6. It therefore closes #258

You can find the administrative search in the admin sidebar or by navigating to /admin/search

This page differs from the regular search in the following ways:
- It uses the new `discoverableAndUndiscoverableItems` configuration from https://github.com/DSpace/DSpace/pull/2706
- As a consequence it shows private and withdrawn items as well as archived items
- There are also facets that enable you to limit your search to search private or withdrawn items
- If an item is private or withdrawn it is shown with a badge on the search result
- It adds buttons underneath each search result with a number of actions an admin can perform on them.

### The new facets:
Screenshot 2020-03-10 at 14 13 07

---

### A private item in list view:
Screenshot 2020-03-10 at 14 39 42

---

### A withdrawn item in grid view:
Screenshot 2020-03-10 at 14 40 00


src/app/shared/object-grid/admin-search-result-grid-element/collection-search-result/collection-admin-search-result-grid-element.component.html
```diff
@@ -0,0 +1,13 @@
+<ds-collection-search-result-grid-element [object]="object"
```
benbosman: This will be a "hidden" feature while testing it since the discovery configuration only returns items. To test this, the discovery configuration will need to be adjusted to remove the item filter (similar for the communities below)
This is not something to adjust (it's good to have this available)

src/app/shared/object-grid/admin-search-result-grid-element/item-search-result/item-admin-search-result-grid-element.component.html
```diff
@@ -0,0 +1,41 @@
...
+    <div *ngIf="dso && !dso.isDiscoverable" class="private-badge">
+        <span class="badge badge-danger">{{ "admin.search.item.private" | translate }}</span>
+    </div>
+    <div *ngIf="dso && dso.isWithdrawn" class="withdrawn-badge">
+        <span class="badge badge-warning">{{ "admin.search.item.withdrawn" | translate }}</span>
+    </div>
+</div>
+<ul #buttons class="list-group list-group-flush">
+  <li class="list-group-item d-flex justify-content-between">
+    <a class="btn btn-light btn-sm my-1 edit-link" [routerLink]="[getEditPath()]" [title]="'admin.search.item.edit' | translate">
```
benbosman: I noticed the list of actions here is repeated in item-admin-search-result-list-element.component.html
I would rather expect the list of buttons to be part of a separate component and have item-admin-search-result-grid-element.component.html and item-admin-search-result-list-element.component.html only apply the style changes.
The same applies for the duplication in the ts

benbosman:
Thanks @LotteHofstede , the updates solve all my feedback.
If you can update this PR to use the configuration name https://github.com/DSpace/DSpace/pull/2706#discussion_r391741381 (small change in src/app/+admin/admin-search-page/admin-search-page.component.html to use administrativeView) I'll test it against master again.

benbosman:
Thanks for the quick update.

I've tested this branch in detail as well now, and I noticed 2 small things which would be good to adjust:
* When I actually make an item public, private, withdrawn, reinstated, I remain on the confirmation page. This happens despite the fact that the PATCH succeeded. When I start it from the /items/uuid/edit/status page, I get redirected to the /items/uuid/edit/status page, which makes more sense. I'm not sure whether the best location would be the search or also the status page (allowing to see the updated state and do more actions on the item), but I would not want the confirmation page to remain open (it feels as if it failed)
* After making an item public, private, withdrawn, reinstated, if I return to the admin search, the item's state is displayed correctly (the cache is aware of the updated state), but the discovery search results are not invalidated (at least the facets did not change).

src/app/shared/translate/missing-translation.helper.ts
```diff
@@ -0,0 +1,10 @@
+import { MissingTranslationHandler, MissingTranslationHandlerParams } from '@ngx-translate/core';
+
+export class MissingTranslationHelper implements MissingTranslationHandler {
```
tdonohue: Could we add TypeDocs to this new class?

LotteHofstede: @tdonohue 
I added the missing TypeDoc.

If I understand correctly, you would like the facets to hide when there is only a single option. I experimented with this by changing the 'available value threshold' of when to show facets from 1 to 2. Unfortunately, this means that if you perform a very specific search, there will be no filters available at all. In my opinion this looks pretty strange.
image

LotteHofstede:
@benbosman 
Thanks for the new feedback.

@artlowel fixed the first issue by making sure the stored requests are retrieved by their UUID instead of their HREF.
The user is now sent to the `edit/status` page after making the item private (or withdrawing it). 
We discovered another issue already present in the current master: after the redirect to the `edit/status` page, the item's state was not yet updated in the view. This was caused by the fact that the item resolver is only called once as long as you are on a subpath of `/items/:id`. I updated the routing modules to make sure the resolvers rerun on every route change, so the item state is updated after being redirected from the page where you edited the item's state.
I also updated this for community and collection sub-routes.

I was able to fix the second issue by removing the discovery requests from the cache when the item is withdrawn/made private/...

src/app/shared/translate/missing-translation.helper.ts
tdonohue: @LotteHofstede : thanks for the explanation and looking into this.  This can be left as-is for now. I don't feel strongly enough about it, so I'm fine with the filters remaining even if only one option exists.

tdonohue:
Merging as this is at +2 and looks good to me now. Thanks @LotteHofstede !

PR #624, opened 2020-03-17
Cache behavior for unknown type

benbosman:
This is a continuation of https://github.com/DSpace/dspace-angular/pull/611

We noticed the Angular UI currently also breaks in the BaseResponseParsingService.process() function if REST contains a type not (yet) known to Angular.

If the data is part of a type not yet known to Angular, the object is null
The null object causes an exception further down the road

The null objects will be ignored here since there's on use in caching them

### DIFF
--- a/src/app/core/data/base-response-parsing.service.ts
+++ b/src/app/core/data/base-response-parsing.service.ts
@@ -51,7 +51,7 @@ export abstract class BaseResponseParsingService {
         return this.processArray(data, request);
       } else if (isRestDataObject(data)) {
         const object = this.deserialize(data);
-        if (isNotEmpty(data._embedded)) {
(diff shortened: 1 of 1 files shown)


src/app/core/data/base-response-parsing.service.ts
```diff
@@ -51,7 +51,7 @@ export abstract class BaseResponseParsingService {
         return this.processArray(data, request);
       } else if (isRestDataObject(data)) {
         const object = this.deserialize(data);
-        if (isNotEmpty(data._embedded)) {
+        if (isNotEmpty(data._embedded) && hasValue(object)) {
```
artlowel: Perhaps it makes more sense to move this `hasValue(object)` down to the `if` on line 60. That way even if the main object in a response can't be parsed, its embeds will still be processed.

PR #645, opened 2020-04-06
#601 resource policies

atarix83:
## References
 [Edit Item - Manage Resource Policies / Roles](https://github.com/DSpace/dspace-angular/issues/601)

## Description
This is an initial PR to implement manage of resource policies. The unit tests are not yet done. 

## Instructions for Reviewers

- Added a link to the item policies page to the item status tab
- Created the policies page component
- Show the current policies for the Item, and each bundle and bitstream within the item
- Added the ability to edit or remove each exiting policy
- added a link to their edit group page
- Added the ability to add a new policy for the item and each bundle and bitstream within the item
- Added methods to the `ResourcePolicyService` to retrieve all policies for a resource and to support CRUD for policies

Known problems due to https://github.com/DSpace/dspace-angular/issues/644 :

- In the resource policy page eperson/group related to the policy should disappear from view
Schermata da 2020-04-06 20-53-52

- In the resource policy create page the list of eperson/group should be empty after the first time you access the page
Schermata da 2020-04-06 20-54-20

- editing fields for policyType and action are disabled because replace operations are not yet implemented https://jira.lyrasis.org/browse/DS-4477

## Checklist

- [ ] If my PR includes new, third-party dependencies (in `package.json`), I've made sure their licenses align with the [DSpace BSD License](https://github.com/DSpace/DSpace/blob/master/LICENSE) based on the [Licensing of Contributions](https://wiki.lyrasis.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-LicensingofContributions) documentation.


atarix83:
test are completed now, PR is ready to be reviewed

jtimal:
Hi @atarix83  
I saw two bugs,
 when you are adding policy sometimes it does not show the users, nor the groups
image
when you try to edit a group send 404 message

atarix83:
@jtimal 
thanks for reviewing. First issue is due to https://github.com/DSpace/dspace-angular/issues/644 and I already mentioned on the PR description.
I'll check the edit group link

jtimal:
Hi @atarix83, tests are successful.

artlowel:
Thanks @atarix83!

Overall it works well. A few comments though:

- It seems that you can only browse for epersons and groups to grant a permission to. In xmlui in DSpace 6, you used to be able to search. That seems more convenient. It should also be possible with the existing rest endpoints.
- You can also no longer change the eperson, group that gets the perission in the edit view, or the policy or action type. You could in DSpace 6. That may be due to a rest restriction though.
- After you delete a policy it still shows up in the overview. To prevent this you should call `requestService.removeByHrefSubstring` to remove calls to `resourcepolicies/search/resource?uuid=${current uuid}` after a delete request succeeds
- The edit policy page should set its own trail. Currently it depends on how you got there. If you come from the edit page, it uses that trail, if you go straight there it simply says "Home"
- The policy overview page looks a bit sloppy
	- There's no space between the delete and add buttons up top
		- perhaps move them somewhere they'll always fit on a single line, or perhaps just shorten the labels to "Delete selected" and "Add"
		- I'd also use the same style buttons as on other pages: `btn-success` with a plus icon for add `btn-danger` with a trash can for delete
	- The edit links aren't aligned
		- in other places (e.g. the edit item metadata or bitstream tabs) we tend to have a column labeled actions at the back, with icons representing the actions. I think for consistency it would be better to use the same aproach here

Screenshot 2020-04-23 at 14 29 38

atarix83:
> It seems that you can only browse for epersons and groups to grant a permission to. In xmlui in DSpace 6, you used to be able to search. That seems more convenient. It should also be possible with the existing rest endpoints.

I'll add a search field

> You can also no longer change the eperson, group that gets the perission in the edit view, or the policy or action type. You could in DSpace 6. That may be due to a rest restriction though.

Yes this is due to a rest restriction https://github.com/DSpace/Rest7Contract/blob/master/resourcepolicies.md#linked-entities

> After you delete a policy it still shows up in the overview. To prevent this you should call requestService.removeByHrefSubstring to remove calls to resourcepolicies/search/resource?uuid=${current uuid} after a delete request succeeds

Yes I'm aware of this, i forgot to list with the known issues. I've already tried to remove calls here https://github.com/4Science/dspace-angular/blob/863520995975694157ba259db6e9b02d161a7d99/src/app/shared/resource-policies/resource-policies.component.ts#L248 but it doesn't work anyway. Is it the right way to remove previous calls? 

> The edit policy page should set its own trail. Currently it depends on how you got there. If you come from the edit page, it uses that trail, if you go straight there it simply says "Home"

I also noticed, but it seems an issue with breadcrumbs change detection because after the first triggered change detection the right trail appears

>  There's no space between the delete and add buttons up top
        perhaps move them somewhere they'll always fit on a single line, or perhaps just shorten the labels to "Delete selected" and "Add"
        I'd also use the same style buttons as on other pages: btn-success with a plus icon for add btn-danger with a trash can for delete
    The edit links aren't aligned
        in other places (e.g. the edit item metadata or bitstream tabs) we tend to have a column labeled actions at the back, with icons representing the actions. I think for consistency it would be better to use the same aproach here

I'll work on these

atarix83:
@artlowel I should have implemented your suggestions

artlowel:
>I've already tried to remove calls here https://github.com/4Science/dspace-angular/blob/863520995975694157ba259db6e9b02d161a7d99/src/app/shared/resource-policies/resource-policies.component.ts#L248 but it doesn't work anyway. Is it the right way to remove previous calls?

No, I meant that you should remove it after you receive a successful response from a DELETE request. 

Removing it right before each request can be problematic because both `removeByHrefSubstring()` and `searchByResource()` work asynchronously. So the possibility exists that `searchByResource`'s new request gets configured and sent out before the remove is executed, this causes the request to be removed before the server can respond, and when the server response comes in it creates an invalid entry in the requestCache: a response without a request.

I've seen that happen before and it causes issues that are very hard to diagnose.

So I wouldn't clear the cache every time just to make sure. Instead, clear it when you know you've made changes and the cache is outdated. That's better for performance anyway.

>I also noticed, but it seems an issue with breadcrumbs change detection because after the first triggered change detection the right trail appears

You're right. I see why as well. I'll create a small PR to fix it.

Thanks for taking my other suggestions in to account. I think the page already looks a lot better.

I'd perhaps remove the padding classes on the edit button, and instead use `btn-sm`, that's what's used in other places.

And perhaps also move the group edit link to a button in the last column as well (you could use [this icon](https://fontawesome.com/icons/users?style=solid) to differentiate from the existing button)

I also noticed that you can't use this page on a phone screen. You can't reach the right side of the tables. I appreciate it's a complicated page, and I don't expect it to look perfect on small screens, but you should be able to use it. Perhaps simply switching to [responsive-tables](https://getbootstrap.com/docs/4.4/content/tables/#responsive-tables) can solve this issue?

I also think the search is a big improvement. Thanks!

atarix83:
@artlowel @tdonohue 
now all feedback should be addressed.
- issue occurred after policy delete should be resolved. 
- issue with eperson/group related to the policy due to cache problem is still present 
- travis build fails because of some e2e tests don't pass. Do you have any idea why?

atarix83:
@artlowel @tdonohue 
build is successful again, could you review this PR again ?

src/app/shared/resource-policies/resource-policies.component.ts
```diff
@@ -112,6 +112,7 @@ export class ResourcePoliciesComponent implements OnInit, OnDestroy {
    */
   ngOnInit(): void {
     this.isActive = true;
+    this.requestService.removeByHrefSubstring(this.resourceUUID);
```
artlowel, line 115: Is this delete necessary? it's still susceptible to the same problem I described in [my earlier comment](https://github.com/DSpace/dspace-angular/pull/645#issuecomment-624645477): it could inadvertently remove the results that you fetch in this. initResourcePolicyLIst() the line below.

atarix83: it's necessary to clean cache otherwise when you go to the authorization page directly (e.g. `http://localhost:4000/items/3834df7e-ed86-491d-8ad7-beeefa137082/edit/authorizations`), the results are not available after was retrieved on SSR
I've moved after results have been retrieved

artlowel: 👍

PR #650, opened 2020-04-09
Administer Workflow

LotteHofstede:
## References
* This PR is dependent on https://github.com/DSpace/DSpace/pull/2727

## Description
This PR adds an overview page of all workflow items, a page to delete a workflow item and a page to send a workflow item back to the submitter.

## Instructions for Reviewers
This PR adds a new page for administrators, the new page is a search page for WorkflowItems. 
Below each search entry, two buttons are provided:

* Delete
* Send back

### Delete
The `delete` button will lead the user to an overview page of the item, where the user can view all metadata of the item.
At the bottom of this page, the user can either choose to cancel or confirm the delete action.
When the user chooses to cancel the action, they are redirected to the previous page or - if there is no previous page in the history - to `/mydspace`.

Clicking the `delete` button on this page will send a DELETE request to the server, with query parameter `expunge=true`. A notification will be shown on the page, to tell the user if the action was performed successfully on the server side or not. If the action is a success, the workflow item and its item will both be deleted from the server.
After the item was deleted, the user is redirected to the previous page, similar to what would happen if they had clicked the `cancel` button.

### Send back
The `send back` button will open a similar page as the `delete` page. 
Confirming this action, however, will send the item back to the submitter instead of removing the item. This happens by sending a DELETE request with query parameter `expunge=false` instead.

### How to test
The new page is available through the admin menu by clicking 'Administer Workflow' (the last item in the menu) or surfing directly to `admin/workflow`. To test the new functionality, make sure there are workflow items in your database. 

## Checklist


jtimal:
Hi @LotteHofstede ,

The search does not return the workflow correctly, I have 12 that are shown in dspace 6 
image
I am looking for one and I cannot find it
image

A question, should all DSpace items appear?
image

LotteHofstede:
@jtimal Thanks for the review!
I'm not sure what's going wrong, could you check if the REST server you're testing with returns the same results?

jtimal:
> @jtimal Thanks for the review!
> I'm not sure what's going wrong, could you check if the REST server you're testing with returns the same results?

Hi @LotteHofstede, can I try the rest of 4sience? and if you can I don't have an admin user

LotteHofstede:
@jtimal I don't think the 4Science REST server has DSpace/DSpace#2727 deployed, so the configuration for the search endpoint would not give the correct results.

tdonohue:
Apologies, we hadn't updated the 4Science demo REST server in some time. That's just been done, so it can *now* be used to test this PR.

jtimal:
Thanks @tdonohue  I'll start testing with the Rest of  4Science

src/app/core/submission/workflowitem-data.service.ts
```diff
@@ -35,4 +39,38 @@ export class WorkflowItemDataService extends DataService<WorkflowItem> {
...
+   */
+  delete(id: string): Observable<boolean> {
+    return this.deleteWFI(id, true)
+  }
+
+  sendBack(id: string): Observable<boolean> {
+    return this.deleteWFI(id, false)
+  }
+
+  private deleteWFI(id: string, expunge: boolean): Observable<boolean> {
```
tdonohue, line 55: Tiny thing, could we add TypeDocs to this method as it's called by both functions above.  I think it's worth describing here that the WFI is deleted either way, but in one situation the result is the WorkSpaceItem still exists, and in the other it's expunged.

LotteHofstede:
@tdonohue Could you try a hard refresh? I had a similar issue yesterday, but a hard refresh fixed it!

src/app/core/submission/workflowitem-data.service.ts
LotteHofstede: Added missing TypeDoc

jtimal:
Hi @LotteHofstede , I was able to test and the operation is correct

tdonohue:
Merging as this has two approvals. Thanks again @LotteHofstede !

PR #661, opened 2020-05-08
Hotfix: use the server port from the environment file

artlowel:
After #625 was merged, the production server always used the default port, not the one in environment.prod.ts

This PR fixes that issue


artlowel:
Merging this immediately as it is trivial, and needed to get the demo server up and running

PR #697, opened 2020-06-04
Refactor registry service and associated models

Atmire-Kristof:
## References
This PR fixes [#685](https://github.com/DSpace/dspace-angular/issues/685)

## Description
This PR aims to refactor the registry service and associated models to use dataservices instead of custom response models and parsing services.

## Instructions for Reviewers

List of changes in this PR:
* Added data-services for both MetadataSchema and MetadataField models
* Refactored `get` methods on the registry-service to use `find` methods on the appropriate data-services (instead of using custom requests)
* Refactored `delete` methods on the registry-service to use `delete` methods on the appropriate data-services
* Moved implementation of `createOrUpdate...` methods from the registry-service to their respective data-services and refactored them to use existing data-service methods
* Moved `clearRequests` methods to the respective data services
* Removed redundant models and services (e.g. custom request/response models, parsing services...)
* Refactored metadata-schema component to comply with the changes made above + some consistency changes (e.g. adding $ to observables)
* Refactored test cases to comply with the changes made above
* Added breadcrumbs to registry routes

These changes shouldn't effect the behaviour of the admin registry pages at all.

## Checklist


Atmire-Kristof:
@atarix83 thanks for the feedback,

I was able to reproduce both issues. After some debugging, I discovered the second issue is caused by the metadata fields cache expiring after 10 seconds. This is added to the request in the  `searchBy` method of the abstract data-service.
I've added an override of this method in the MetadataFieldDataService without the specified `responseMsToLive`.
I've also added some safe-checks like an `isNotEmptyOperator` when retrieving a MetadataSchema in the RegistryService.

These changes seem to have fixed the issues, as I'm not able to reproduce them anymore.

src/app/core/data/metadata-field-data.service.ts
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Make a new FindListRequest with given search method
+   *
+   * @param searchMethod The search method for the object
+   * @param options The [[FindListOptions]] object
+   * @param linksToFollow The array of [[FollowLinkConfig]]
+   * @return {Observable<RemoteData<PaginatedList<MetadataField>>}
+   *    Return an observable that emits response from the server
+   */
+  searchBy(searchMethod: string, options: FindListOptions = {}, ...linksToFollow: Array<FollowLinkConfig<MetadataField>>): Observable<RemoteData<PaginatedList<MetadataField>>> {
```
atarix83, line 98: Maybe overwriting the entire method may not be the right approach. We could change the `searchBy` method of the data service in the way it works like for other [find methods](https://github.com/DSpace/dspace-angular/blob/ee9f5ec7f1d9b5af20f36460f8723cfe6961de50/src/app/core/data/data.service.ts#L243). In case of `responseMsToLive` has no value it should use the current value `10 * 1000`

artlowel: Done

PR #701, opened 2020-06-11
Misc edit community and collection fixes

samuelcambien:
## References

https://github.com/DSpace/dspace-angular/issues/665
https://github.com/DSpace/dspace-angular/issues/666

## Description

This PR addresses some issues with the Community & Collection edit pages.

## Instructions for Reviewers

List of changes in this PR:
* (DSOSelectorModalWrapperComponent): the mechanism to automatically select the current Community/Collection in the new/edit menu has been fixed
* (ComcolMetadataComponent): the request method used by the edit Community/Collection page has been changed from PUT to PATCH to repair the edit functionality
* (ComcolMetadataComponent): an appropriate error message is shown when the user doesn't have sufficient edit rights. I didn't add error message for other cases, because I don't think a 'Bad Request' or an 'Internal Server Error' message would have an added value to most end users.
* (ComcolMetadataComponent): I've split up the logo <> metadata logic, to make sure both can happen at the same time

I've tested the following cases, both for communities and collections:
* creating a new community/collection with metadata and with a logo
* creating a new community/collection with metadata and without a logo
* adding a community/collection logo without editing the metadata
* deleting a community/collection logo without editing the metadata
* adding a community/collection logo while editing the metadata
* deleting a community/collection logo while editing the metadata
* editing community/collection metadata without changing the logo
* trying one of the above as a user without permission, to verify the error message (I used a local REST server to test this, to be able to easily deny the permission)

### DIFF
--- a/src/app/shared/comcol-forms/comcol-form/comcol-form.component.spec.ts
+++ b/src/app/shared/comcol-forms/comcol-form/comcol-form.component.spec.ts
@@ -22,6 +22,7 @@ import { NotificationsService } from '../../notifications/notifications.service'
 import { NotificationsServiceStub } from '../../testing/notifications-service.stub';
 import { VarDirective } from '../../utils/var.directive';
 import { ComColFormComponent } from './comcol-form.component';
+import { Operation } from 'fast-json-patch';
 
 describe('ComColFormComponent', () => {
   let comp: ComColFormComponent<DSpaceObject>;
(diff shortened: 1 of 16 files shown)


artlowel:
@atarix83 That is an issue in master as well: #706 
Seems recent, because it wasn't there when I tested this PR last week.

But you're right it makes sense to fix it in the context of this PR. We'll look in to it.

artlowel:
@atarix83 it should be fixed now. Can you verify?

PR #705, opened 2020-06-11
Add Hungarian translation

akoscomp:
Create a new file for Hungarian translation and translate it, bases on latest en.json5 file.


bram-atmire (approved):
As provided, several of the keys had line ending problems that I corrected. It now builds and works. I also ensured that Hungarian is now by default enabled.

PR #834, opened 2020-08-07
latvian translation update

MixonZ:
* Latvian translation update


bram-atmire (approved):
@MixonZ I had to resolve a merge conflict because of two orcid related keys that were renamed (.orcid vs .orcidV2). The rest looked good: the file builds and deploys locally without problems.

PR #871, opened 2020-09-21
Ensure CSR fallback only happens on actual errors

artlowel:
## References
Fixes issue introduced by #863

## Description
This PR fixes an issue introduced by #863, which caused the error handler on the express server to be triggered when there were no errors.

## Instructions for Reviewers
In #863 we switched to the native express error handler. But we noticed on dspace-demo that the error fallback happens for every request. 

The reason is that it's not a dedicated error handler, but simply the next function in the chain that can make changes to the response. There's only an error if `err` is defined. 

This PR fixes that issue by checking if there's an error, and if not, passing the data from the previous function (the angular app) through.

To test it, triggering a 404 (as suggested in #863) is not sufficient for causing an actual express error. So I suggest you simply add `throw new Error('A test error');` to [the constructor of ServerAppModule](https://github.com/DSpace/dspace-angular/blob/main/src/modules/app/server-app.module.ts#L103), in which case you should see `Error in SSR, serving for direct CSR` in the server output with a stack trace. And if you leave it out that fallback shouldn't trigger, so that message shouldn't be there.

## Checklist

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -33,7 +33,7 @@ import { enableProdMode, NgModuleFactory, Type } from '@angular/core';
 import { REQUEST, RESPONSE } from '@nguniversal/express-engine/tokens';
 import { environment } from './src/environments/environment';
 import { createProxyMiddleware } from 'http-proxy-middleware';
-import { hasValue } from './src/app/shared/empty.util';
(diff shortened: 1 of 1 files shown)


tdonohue (approved):
👍 Looks good to me. Thanks @artlowel !

PR #938, opened 2020-11-10
Metadata identifier for Communities and Collections 

alemarte:
## References
* Fixes [#887](https://github.com/DSpace/dspace-angular/issues/887)
* Requires DSpace/DSpace [#3046](https://github.com/DSpace/DSpace/pull/3046)

## Description
The Community and Collection rest repositories should now return the handle value inside the metadata attribute 'dc.identifier.uri'. We ensure that that value is presented in place of the previously used handle field.

## Instructions for Reviewers
The community and community objects have been enhanced with a getter method for the handle field. The getter takes care of fetching the value from the metadata. The old value is still stored in a private field (_handle) but never used.

## Checklist

### DIFF
--- a/src/app/core/shared/collection.model.ts
+++ b/src/app/core/shared/collection.model.ts
@@ -1,4 +1,4 @@
-import { autoserialize, deserialize, inheritSerialization } from 'cerialize';
+import { autoserialize, deserialize, deserializeAs, inheritSerialization } from 'cerialize';
 import { Observable } from 'rxjs';
 import { link, typedObject } from '../cache/builders/build-decorators';
 import { PaginatedList } from '../data/paginated-list';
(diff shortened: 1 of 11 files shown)


src/app/core/shared/community.model.ts
```diff
@@ -64,6 +63,17 @@ export class Community extends DSpaceObject implements ChildHALResource {
   @link(COMMUNITY, false)
   parentCommunity?: Observable<RemoteData<Community>>;
 
+  /**
+   * A string representing the unique handle of this Community
+   */
+  get handle(): string {
+    return this.firstMetadataValue('dc.identifier.uri');
```
artlowel, line 70: If the getter for `handle` returns `this.firstMetadataValue('dc.identifier.uri')` do we still need the setter, or the private `_handle`  property?

src/app/core/shared/collection.model.ts
```diff
@@ -75,6 +74,17 @@ export class Collection extends DSpaceObject implements ChildHALResource {
   @link(COMMUNITY, false)
   parentCommunity?: Observable<RemoteData<Community>>;
 
+  /**
+   * A string representing the unique handle of this Collection
+   */
+  get handle(): string {
+    return this.firstMetadataValue('dc.identifier.uri');
```
artlowel, line 81: Same comment

artlowel (changes requested):
Thanks @alemarte!

I can verify that it works. I have some minor inline comments though.

There are also a few open LGTM alerts

src/app/core/shared/community.model.ts
alemarte: The setter has been introduced because in the initial solution the old field was used as a fallback value in case the metadata was not present. But since it is not used anymore, and I don't see any reason to use that value which is not so accurate through out the entire application, I agree that its presence doesn't make much sense. I'll remove it (from both collections and communities).

artlowel (approved):
Thanks @alemarte

tdonohue (commented):
The code here looks good to me.  Thanks @alemarte!  This is just waiting on a final test once the backend PR is completed (and it's nearly there).  So, once the backend is done, I'll give it a final test & merge if successful.

tdonohue (approved):
👍 Tested alongside the backend PR https://github.com/DSpace/DSpace/pull/3046 and they both work great.  Thanks @alemarte!

PR #946, opened 2020-11-17
#945 Clean node at end to avoid rimraf error

### LINKED ISSUE #945: rimraf deletes itself before yarn clean finished
**Describe the bug**
When running `yarn clean` and not having `rimraf` installed globally, a build error like the following appears. 

```
$ rimraf node_modules
$ rimraf src/environments/environment.ts
/bin/sh: rimraf: command not found
```

**To Reproduce**
1. Don't have rimraf installed globally
2. Run yarn clean
3. You'll see a build error like the above

**Expected behavior**
Finishing without reporting an error.

**Related work**
N/A

cwilper:
## References
* Fixes #945 

## Description
Cleans node at end in order to avoid rimraf deleting itself.

## Instructions for Reviewers
Quick fix


artlowel (approved):
Thanks @cwilper!

Issue #964, opened 2020-12-07
EntityTypeService `getEntityTypeByLabel` is mocked: needs proper implementation 


AndrewZWood:
**Describe the bug**
Currently the `getEntityTypeByLabel()` of `EntityTypeService` is mocked to work properly. However this mock hardcodes valid `EntityType` strings which is not very flexible for instances of DSpace with custom `EntityTypes`. I have come up with a solution to dynamically retrieve the `EntityType` by it's `label` now that there are REST endpoints available to do so.
Submitting PR shortly.

AndrewZWood:
PR: https://github.com/DSpace/dspace-angular/pull/965

PR #965, opened 2020-12-07
issue 964 properly impl getEntityTypeByLabel

AndrewZWood:
Fixes https://github.com/DSpace/dspace-angular/issues/964

### DIFF
--- a/src/app/core/data/entity-type.service.ts
+++ b/src/app/core/data/entity-type.service.ts
@@ -88,30 +88,10 @@ export class EntityTypeService extends DataService<ItemType> {
    * @param label
    */
   getEntityTypeByLabel(label: string): Observable<RemoteData<ItemType>> {
-
-    // TODO: Remove mock data once REST API supports this
-    /*
-    href$.pipe(take(1)).subscribe((href) => {
-      const request = new GetRequest(this.requestService.generateRequestId(), href);
-      this.requestService.configure(request);
+    let endPoint;
+    this.halService.getEndpoint(this.linkPath).pipe(take(1)).subscribe((result) => {
+      endPoint = result;
     });
-
-    return this.rdbService.buildSingle<EntityType>(href$);
-    */
-
-    // Mock:
-    const index = [
-      'Publication',
-      'Person',
-      'Project',
-      'OrgUnit',
-      'Journal',
-      'JournalVolume',
-      'JournalIssue',
-      'DataPackage',
-      'DataFile',
-    ].indexOf(label);
-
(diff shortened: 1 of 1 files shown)


src/app/core/data/entity-type.service.ts
```diff
@@ -88,30 +88,10 @@ export class EntityTypeService extends DataService<ItemType> {
...
-      'OrgUnit',
-      'Journal',
-      'JournalVolume',
-      'JournalIssue',
-      'DataPackage',
-      'DataFile',
-    ].indexOf(label);
-
-    return this.findById((index + 1) + '');
+    return this.findByHref(endPoint + '/label/' + label);
```
artlowel, line 95: `endPoint` will only get a value if `this.halService.getEndpoint()` is synchronous. That won't be the case if it isn't already cached, or if the cache time runs out.

 What you want is something like:
```typescript
return this.halService.getEndpoint(this.linkPath).pipe(
  take(1),
  switchMap((endPoint: string) =>
    this.findByHref(endPoint + '/label/' + label))
);
```

That way `this.findByHref` will wait for the result of `this.halService.getEndpoint` even if it isn't synchronous

artlowel (changes requested):
Thanks @AndrewZWood!

Theres an issue with the implementation. See my inline comment.

src/app/core/data/entity-type.service.ts
AndrewZWood: That neat.  Didn't know I could return like that. Will Update and test this afternoon 9 ( EST ).

AndrewZWood: Updated code and tested with local data: https://github.com/DSpace/dspace-angular/pull/965/commits/06bb2130783c2c042b1ea7698809fd3f0bec8124

artlowel (approved):
Thanks @AndrewZWood I can verify that your code works now. 

However it has brought a bug on the rest side to light. The entity types retrieved in this way can't be parsed, because they're lacking a _links section. I created an issue for it.

So this PR should not be merged until that issue is fixed on the backend

AndrewZWood:
@artlowel Already have an issue and PR for what you're describing: https://github.com/DSpace/DSpace/issues/3078

tdonohue (approved):
👍  I tested this alongside https://github.com/DSpace/DSpace/pull/3079 (corresponding REST PR) and they work great!  Thanks @AndrewZWood ! 

I'll merge both together & ensure the REST PR is deployed to the demo site.

PR #978, opened 2020-12-30
Update /claimedtasks and /pooltasks to support reloading of workflow actions on individual WorkflowItems

alemarte:
## References
* Fixes https://github.com/DSpace/dspace-angular/issues/721
* Requires https://github.com/DSpace/DSpace/pull/3054

## Description
With this pull request we manage to properly reload the workspaceitem when needed after a user interaction. Before this the entire page was reloaded every time, without exceptions. 

In particular, 4 case have been implemented and tested

1) From PoolTask to ClaimTask
2) From ClaimTask to PoolTask 
3) After Rejection
4) After Approvation

## Instructions for Reviewers
To achieve this, the angular action abstraction has been enhanced, introducing the [MyDSpaceReloadableActionsComponent](https://github.com/4Science/dspace-angular/blob/5700450eec178dc60a4f3105535adfde1df242db/src/app/shared/mydspace-actions/mydspace-reloadable-actions.ts#L25), a further level over the canonical MyDSpaceActionsComponent. The MyDSpaceReloadableActionsComponent takes care to manage the entire logic of the action operation which consists of the following phases:
1) The initialization
  Since the workspaceitem could change after the execution of the action, an anchor to reaload the objected could be necessary to reload the object (in our typical case the item uuid to reload the new workspaceitem). This is accomplished by the [initReloadAnchor()](https://github.com/4Science/dspace-angular/blob/5700450eec178dc60a4f3105535adfde1df242db/src/app/shared/mydspace-actions/mydspace-reloadable-actions.ts#L110) method.
2) The execution
  Every concrete class must implement how to perform the action. This is accomplished by the [actionExecution()](https://github.com/4Science/dspace-angular/blob/5700450eec178dc60a4f3105535adfde1df242db/src/app/shared/mydspace-actions/mydspace-reloadable-actions.ts#L42) method.
3) The reload of the object in case of success
  This is accomplished by the [reloadObjectExecution()](https://github.com/4Science/dspace-angular/blob/5700450eec178dc60a4f3105535adfde1df242db/src/app/shared/mydspace-actions/mydspace-reloadable-actions.ts#L47) method, and by the [convertReloadedObject()](https://github.com/4Science/dspace-angular/blob/5700450eec178dc60a4f3105535adfde1df242db/src/app/shared/mydspace-actions/mydspace-reloadable-actions.ts#L116) method which takes care to instantiate the proper MyDspace object given the reloadedObject.
4) Handle of the response [handleReloadableActionResponse()](https://github.com/4Science/dspace-angular/blob/5700450eec178dc60a4f3105535adfde1df242db/src/app/shared/mydspace-actions/mydspace-reloadable-actions.ts#L89)
 To show notifications and to notify the parent component that the process has been completed and a new reloaded object could be available to be displayed. This is important because if the Type has changed, and the type is managed by a different container, ListableObjectComponentLoaderComponent will have to destroy the previous Container and [instantiate a new one](https://github.com/4Science/dspace-angular/blob/5700450eec178dc60a4f3105535adfde1df242db/src/app/shared/object-collection/shared/listable-object/listable-object-component-loader.component.ts#L106).

## Checklist
 

alemarte:
Apart from 2 unit tests that keep passing in local environment and fail without any reasonable clue on github, this PR is ready to be reviewed.
I'll manage to turn test green and resolve conflicts in the meantime.

tdonohue (changes requested):
@alemarte : I tested this today, and while it "mostly works", I've also been able to cause some very odd behavior by simply clicking "Claim" and "Return to Pool" on the same WorkflowItem several times in a row.  After doing that a few times, I see the same "Operation is successful" green message displayed up to 7 times on the right side.  The message times out, but then comes back automatically a moment later....almost like there's a temporary loop in the notification message.  Eventually though it stops coming back.

I've also found that (eventually) the WorkflowItems disappear entirely from my `workflow` view (http://localhost:4000/mydspace?configuration=workflow), even though if I go to the `workspace` view (http://localhost:4000/mydspace?configuration=workspace), I see them listed with a `workflow` blue label.  I'm not certain how I caused this exactly (again I was mostly just claiming & unclaiming items), and I'm not sure if this issue is caused by this PR or not (as the WorkflowItems remain invisible after uninstalling this PR).  But, I end up seeing this error in Chrome DevTools:
```
ERROR TypeError: Cannot read property 'replace' of undefined
    at i (search.utils.ts:28)
    at o (search.utils.ts:12)
    at e.getFacetValue (search-facet-option.component.ts:115)
    at e.isChecked (search-facet-option.component.ts:86)
    at e.ngOnInit (search-facet-option.component.ts:75)
    at core.js.pre-build-optimizer.js:21097
    at core.js.pre-build-optimizer.js:29495
    at Ny (core.js.pre-build-optimizer.js:29457)
    at c_ (core.js.pre-build-optimizer.js:29998)
    at Object.updateDirectives (search-authority-filter.component.html:6)
```

Overall, I found the basics of this PR seem to work. I'm no longer seeing the page refresh.  However, either the PR has some instability, or it's made visible existing instability in the Workflow view of the MyDSpace page.   

Let me know if you can reproduce either of these issues.  If not, I can always try again by resetting my database & attempting to retest this PR with a fresh database.

tdonohue:
In today's meeting, @abollini noted that the issue above with "disappearing" WorkflowItems might actually be a side-effect of the bug solved in https://github.com/DSpace/DSpace/pull/2744   So, it may not be a fault of this PR or the corresponding backend PR.  I'll be doing some testing to see if that is true, and whether the issue I described can still be reproduced even with https://github.com/DSpace/DSpace/pull/2744 applied to the backend.

tdonohue:
@alemarte : Please rebase on latest `main`.  Keep in mind we just updated to Angular 10, which changed some lint rules. So, you may also want to ensure `yarn lint` doesn't throw errors (as its possible the PR may need minor formatting updates to pass).

alemarte:
@tdonohue I've rebased with the latest changes.
The multiple notifications and the type errors were actually bugs and I have solved them.
When you switch multiple times from pool to claim and viceversa, an error still remains. After a deep investigation supervised by @atarix83 it is very likely to be a cache problem when searching the task by item after the component reload. The rest request is the same, but the result should change since the task meanwhile has changed its identifier.

alemarte:
I've merged the second part of the cache implementation but the cache problem remains.

Long story short, these are the steps to reproduce the issue (to better debug this scenario I've added and committed the generation of a correlationId to the searchTask service which log the href and the response deliveredy by the service that in the end will be inconsistent with the payload coming from the network).

1) Given a PoolTask press on 'Claim button'. In this example I tested on this named Moloaka.
q1

When a pool task is claimed the PoolTask is deleted and a ClaimTask(87) is created. We need to retrieve the created claim task
and the strategy adopted is to call the /workflow/claimedtasks/search/findByItem of the ClaimTaskService. This is possible since the component conveniently keep the itemUUID in its state. (correlationId = 20)

q2

Then the object is emitted so that parent components can destroy the PoolTaskActionsComponent and create the fresh new ClaimedTaskActionComponent. As we can see in the second screenshot the ClaimedTaskActionComponent retrieve the workflowitem during its initialization.

q3

This game is played the same with the claimed task so we easily return in the pool task state (correlationId = 395).

From this point, clicking for the second time on 'Claim' we can see an inconsistent behaviour (correlationId = 639)

q11

As we can see we search again the claimed task by item uuid. The backend respond with a ClaimTask(88) but the service logs the previous version (87). So the ClaimTask presented to the user is the older one.

Clicking on 'Return to pool' cause a not found since it try to perform the operation on the ClaimTask(87) which doesn't exist enymore.

q5

The invoked getSearchByHref method doesn't have arguments to manage the cache strategy, so I'd expect the cache is never hit.
The ClaimedTaskDataService and the PoolTaskDataService have been configured to follow links by exploiting the new useCachedVersionIfAvailable to false, and they still have a responseMsToLive equal to 1000.

tdonohue:
@alemarte : Thanks for the screenshots, they definitely help to show the problem.  It **does** obviously show, based on your _4th screenshot_, that the REST API response has returned a new ClaimedTask (ID=88), while the Angular UI seems to still be using the cached, old/prior ClaimedTask (ID=87).

It almost seems like when the task is returned to the Pool, the original ClaimedTask with ID=87 is deleted (on the backend) but it is never removed from the cache on the Angular side.  So, when it is reclaimed, that old ClaimTask is somehow reused.  I'm not sure myself why that would be either....but, maybe @artlowel would see an obvious reason based on his work on refactoring the cache?

alemarte:
@tdonohue Just a note. I think this is not the case of a regression of the cache refactoring, because this behaviour was present before merging both parts.

src/app/shared/object-collection/shared/listable-object/listable-object-component-loader.component.ts
```diff
@@ -104,14 +125,24 @@ export class ListableObjectComponentLoaderComponent implements OnInit {
...
+
+    if ((componentRef.instance as any).reloadedObject) {
+      this.subs.push((componentRef.instance as any).reloadedObject.subscribe((reloadedObject: DSpaceObject) => {
+        if (reloadedObject) {
+          console.log('Reloaded Object from/to', this.object, reloadedObject);
+          this.object = reloadedObject;
+          this.instantiateComponent(reloadedObject);
+        }
+      }));
+    }
```
artlowel, line 145: I think the problem lies with this change.

Here, a component is created, and afterwards a subscription to an observable property of that component is created. When it emits, the component is recreated.

However the original component is never destroyed, nor is the subscription ever unsubscribed.

To make this work you should keep track of the componentref you create, and at the start of the method, check if it exists, and destroy it if it does. Same for the subscription: before you subscribe, check whether the subscription already exists and unsubscribe if it does.

However in general this seems like a bad idea. Rather than shoehorning in a way to reload here, I think you should set the list containing all the results to stale, which will cause it to be automatically reloaded. Any change we make to this component has a chance to break things in other places, as it is used for every dynamically loaded list.

alemarte: I'm 100% sure the old component is destroyed because I added logs to its onDestroy. Is destroyed because of the [viewContainerRef.clear();](https://github.com/DSpace/dspace-angular/blob/d015b6caefa06dfc72ce649e6f716cf1ea2c9095/src/app/shared/object-collection/shared/listable-object/listable-object-component-loader.component.ts#L119). Previous component is destroyed, emitter completed and subscription unsubscribed.

tdonohue: @artlowel : I wonder if you might quickly summarize the alternative approach you mentioned in today's meeting?  

You noted that this sort of individual component reloading is a bad idea, and that ideally we should be finding a way to set this individual result to "stale" and perhaps reloading it via a service?  Did I understand that correctly, or am I misstating your idea?

As noted in today's meeting, @alemarte and @atarix83, it seems like we may need to move this ticket forward mostly "as-is", just to get the basic fix in place.  However, I want to better understand the "ideal" solution/implementation, so that we can create a follow-up task to replace this component reloading with a better solution.

tdonohue (commented):
@alemarte :  I tested this again today.  The feature somewhat works, but as you've noted, it's unstable if you click "Claim" and "Return to pool" several times in a row.  The oddity is that the behavior I'm now seeing is **NOT** as you've described in your [previous comment](https://github.com/DSpace/dspace-angular/pull/978#issuecomment-775993429) 

Here's what I'm now seeing:

1. First, I've created two Workflow tasks. Both are _unclaimed_ to start with
2. Click "Claim" on one task. (Successful, ClaimedTask is created, only it is reloaded)
3. Click "Return to Pool" on that same task. (Successful, PoolTask is created, only it is reloaded)
4. Click "Claim" on that same task again (**Successful**, initial ClaimedTask appears to be **reused** according to the console log, and only it is reloaded.)
    * If a wait a minute or two before clicking "Claim" a second time, I'll get a successful message and the task will **disappear** from my list.  It only reappears if I completely reload the page.
5. Click "Return to Pool" on that same task again (404 Error occurs. Attempts to call DELETE on a ClaimedTask that doesn't exist.)   Clicking any other button on this same item also results in an error (so it cannot be Approved or Rejected either), as it seems to have loaded the cached object or component.

The behavior here is very inconsistent and worrisome.  While I wish I could merge this, the current behavior on the demo site (https://demo7.dspace.org/) at least doesn't result in errors occurring (even though it's not desirable to have the entire page reload after each action).  Do you or @atarix83  have ideas on how we can resolve this caching issue in a follow-up PR?  If so, we could merge this "as-is" and resolve it immediately afterwards.  But, I'm worried about the behavior of this PR as-is and I'd want some sort of proposal before we can move this forward.

atarix83:
@tdonohue I think that the issue you described is the same that @alemarte reported [here](https://github.com/DSpace/dspace-angular/pull/978#issuecomment-775993429), and it should have been resolved now, please check it again.
It was due to the fact that [method searchTask of the TasksService](https://github.com/4Science/dspace-angular/blob/CST-3620/src/app/core/tasks/tasks.service.ts#L107) returned the stale response before to fetch the new request without waiting for the new response.
We've resolved, with a temporary workaround, by adding a [filter to avoid to retrieve a stale response](https://github.com/4Science/dspace-angular/blob/c78cd9ad71cc361cf6f9bc6aca68ff46e614f161/src/app/core/tasks/tasks.service.ts#L113).
In our opinion this kind of filter should not be necessary because we expect to have directly the new response instead of the old stale one. Please @artlowel correct me if I'm wrong and this is not the expected behavior.

regarding what @artlowel said [there](https://github.com/DSpace/dspace-angular/pull/978#discussion_r578387637),  we [changed the code](https://github.com/4Science/dspace-angular/blob/c78cd9ad71cc361cf6f9bc6aca68ff46e614f161/src/app/shared/object-collection/shared/listable-object/listable-object-component-loader.component.ts#L138-L145) to be sure the `reloadedObject` subscription is taken once and we made sure the dynamic component instantiated by the loader is destroyed before to the creation of the new one.

tdonohue (commented):
@atarix83 : I've retested this PR again today (after pulling down your latest changes).  The behavior is changed, but it's still not working as one would expect.  Here's what I see today (bolded text is the new behavior as of today):

1. First, I've created two Workflow tasks. Both are _unclaimed_ to start with
2. Click "Claim" on one task. (Successful, ClaimedTask is created, only it is reloaded)
3. Click "Return to Pool" on that same task. (Successful, PoolTask is created, only it is reloaded)
4. Click "Claim" on that same task again (**Successful, but the task now DISAPPEARS entirely**.  The only way to get it to reappear is to reload the entire page.)

This is still worrisome behavior, as I'm sure testers will encounter this almost immediately. While claiming/unclaiming like this may not be frequent behavior, as soon as a user does this they'll worry about the stability of DSpace if their tasks disappear on them.

src/app/shared/object-collection/shared/listable-object/listable-object-component-loader.component.ts
artlowel: @tdonohue 
>You noted that this sort of individual component reloading is a bad idea, and that ideally we should be finding a way to set this individual result to "stale" and perhaps reloading it via a service? Did I understand that correctly, or am I misstating your idea?

I think replacing a single result that's part of a list is asking for problems yes, but the fact that the list was being reloaded is exactly the issue that's being reported in #721. I missed that yesterday. Because I was short on time, I only looked in to the specific error described in [@alemarte's comment](https://github.com/DSpace/dspace-angular/pull/978#issuecomment-775993429), not at the reason for the PR. I said we should set the list to stale and then it would automatically be re-retrieved, but that's likely exactly what happened before this PR and what's causing #721, so it wouldn't help in this case.

The main problem I see with the approach in this PR is that you have a list of objects coming from the my-dspace-page component. The objects get passed down to the child components. They pass them to their children etc. Halfway down the line this PRs swaps out one of those objects for a different one, and for all components below it, the new object is passed to the input, but all components above it don't know about that change and still contain references to the original object. I think that can lead to all sorts of unwanted side effects. If any component above the place the swap happens uses it for anything, you've got a confict. 

I think an ideal solution would be to change the list at the source: resultsRD$ in MyDSpacePage component. It should no longer be tied 1 to 1 to the result from the search service, but instead it should only be initialized with that result, and if we need to swap an object later, we swap it there. That would ensure all components in the chain are working with the same information, and it would fix the issues mentioned in #721

artlowel:
>In our opinion this kind of filter should not be necessary because we expect to have directly the new response instead of the old stale one. Please @artlowel correct me if I'm wrong and this is not the expected behavior.

You're right, that is a bug with the Stale state. When you retrieve an object using a dataservice, the first remote data emitted should _never_ be stale, the only way to end up with a stale remotedata object should be to be already be subscribed to it as it becomes stale. I'll create a ticket for this.

>regarding what @artlowel said there, we changed the code to be sure the reloadedObject subscription is taken once and we made sure the dynamic component instantiated by the loader is destroyed before to the creation of the new one.

Thanks

src/app/shared/object-collection/shared/listable-object/listable-object-component-loader.component.ts
tdonohue: @artlowel : I'm not sure I fully understand your "ideal solution" in the context of this PR.  Are you suggesting that we need to rethink the single object reload altogether & replace it with a "smarter" reload of the entire list (one that won't cause you to lose context within your list of results, like described in #721)?   

I'm just trying to determine proper next steps here...as from my tests yesterday & today, this PR currently isn't working well & I'm not yet sure how to move it forward.

artlowel: >Are you suggesting that we need to rethink the single object reload altogether & replace it with a "smarter" reload of the entire list (one that won't cause you to lose context within your list of results, like described in #721)?

Exactly

tdonohue: @atarix83 and @alemarte: I'd appreciate your feedback here.  I fully realize this PR has been a lot of work so far, but I'm also finding that it's really not working as expected & reloading individual objects/components is proving to be quite difficult.  I'm also noticing that within the rest of the UI, we never reload individual objects in a list (even in similar features like the Admin Search, which has button actions as well. In that example, we instead use a popup modal when clicking the button to perform a confirmation...and then reload the list as needed).

Do we need to consider another solution here?  Perhaps reloading individual objects/components is a flawed approach altogether & we need to investigate a way to reload the list while keeping your context the same?  One example might be something like this:

* Suppose you are looking at Item 2 on page 3 of the list of PoolTasks
* You click "Claim" on Item 2.
* This reloads the *entire page/list*, but you are kept at the same location in the list. Ideally you "follow" the new ClaimedTask.
    * In most situations this should just mean the reloaded list _must come back in the same order_ & you remain on the same page of that results list
    * If somehow the list size has changed (e.g. other people claimed tasks at the same time, reducing the size of the list), then you should move to the page where the ClaimedTask was. (However, this also could be logged as a known bug for now as we determine a better way to solve it)

I'd appreciate your feedback here, as I'm honestly not sure how to proceed with this PR as-is.  It seems like it's becoming problematic to reload individual objects/components, and I'm not wanting to delay Beta 5 in order to design/build a complex solution to this problem.

alemarte:
@tdonohue I finally have a local fully working version of this PR. Having targeted the cache problem help me to focus more on the reload flow and actually, there is a non deterministic situation in which after the action completion we try to reload a task with the proper ID, but that hasn't been indexed by SOLR yet. So the component disappear. Introducing a delay of 500ms between the two operation seems to solve and I'm turning state seamlessly like a charm. Of course the delay is not a proper solution, my final implementation would involve a retry with max attemps strategy in case of 404 reloading the object.

tdonohue (commented):
@alemarte and @atarix83 : I have been able to confirm today that the most recent changes now allow a single test to be claimed/unclaimed several times in succession without any errors!  I have *not* had a chance to review all the code, and I'd look to @artlowel for his expertise in this area.

I did however discover a separate bug (which appears as if it is not directly related to this code).  I've found that I'm able to see PoolTasks which I **don't seem to have access to**.   This may be a backend issue though, as it's only reproducible by doing the following:
1. First, submit several Items into a Collection with Workflow enabled (creating 1-2 PoolTasks)
2. Then, add a *new EPerson* to the list of reviewers for that Collection
3. Now, login as that *new EPerson*.  You'll be able to see those PoolTasks (created in step 1), but if you attempt to claim one you'll get a `403 Access is denied` exception, with this underlying error:
     ```
      org.dspace.app.rest.exception.RESTAuthorizationException: org.dspace.authorize.AuthorizeException: You are not allowed to to perform this task.
      ```
However, as noted, I think this may be an existing error.  If you agree, I can always create a new ticket related to this bug.

alemarte:
@artlowel About the fix details, it was easier than expected. The failing part of the process was a followlink of the workspace which causes 404. I've removed it from the task services, since it is a responsibility of the new instantiated component to fetch the workspace when needed. But at that point the backend should be consistent.

About the results inconsistency I totally agree. In fact switching the state of task for example, makes the search filter stale.
It is not clear if bubbling up the entire object is something necessary. For now I've implemented a simple notification system that tells the root component that 'something has changed'. [This is now used by the sidebar to refresh its filters](https://github.com/DSpace/dspace-angular/blob/523b7a497cbac5a08cf8dae9fcf85051d5d2648b/src/app/shared/search/search-filters/search-filters.component.ts#L73).

artlowel:
I can no longer reproduce the issue. This PR seems to work the way it should.

While the contentChange emitter makes us more able to deal with side effects that occur when swapping objects, I still think we should go with an approach that doesn't have those side effects. Where the page component is aware that a swap has been made and is the one to push the new object down to its subcomponents. That way every subcomponent on that page would be informed of that change without us having to add extra event emitters and callbacks to fix edge cases as they pop up. 

To clarify:
- Don't have MyDSpacePageComponent call searchservice.search directly, but call a method on a new intermediary service. That method initially simply emits the search results from searchservice
- When you need to swap out a result, call another method on that intermediary service with the self links of the old and the new objects to let it know which swap needs to happen.
- The intermediary service retrieves the new object, swaps the old object for the new one in the list of results, and emits the updated list in the observable that MyDSpacePageComponent is still subscribed to
- This is turn causes the MyDSpacePageComponent to push the updated list down to its subcomponents, and eventually the component that renders the actions
- Other observables on MyDSpacePageComponent, independent from resultsRD$ but that do need to refresh when a swap happens, such as configurationList$ can get a similar method in the intermediary service, that can also auto-emit again when a swap happens.

That would fix #721 in that it wouldn't cause the component to move or the page to scroll, it would also ensure that all subcomponents on the page are working with the same data at all times, and it wouldn't require changes to components that aren't really related to this issue such as ListableObjectComponentLoaderComponent

atarix83:
I could share the reasoning done by @artlowel but it makes sense only when both search results and filtering facet values are handled by the same root component.
Instead what happens now is that `MyDSpacePageComponent` provides the search results while the search filters are managed by the `SearchFiltersComponent`. In my opinion this is not the best approach indeed the consequence of this is that we had to propagate a kind of event from the parent component (MyDSpacePageComponent) down to the child (SearchFiltersComponent) to keep updated the filters value when a workflow task changes its status.

For what I said I reckon that all the search components need a refactoring in order to have a root component that take care of all search state and propagate the results down to the children, but this work is out of the scope of this PR and can be planned after the 7.0 release

tdonohue (approved):
👍 Retested today, and this seems "good enough" for 7.0.  However, I've logged two tickets for followup work:

1. First, there's a bug (on the backend) that you cannot claim tasks created *before* you were added to a workflow group.  It appears (to me) that this is a backend bug & I think it should be fixed in Beta 5 (or 7.0 final) if possible: https://github.com/DSpace/DSpace/issues/3175
2. Second, as @artlowel pointed out & @atarix83 seems to agree with, we believe it's problematic for MyDSpace to use different components than other search results pages.  This was already partially described in #773. However, I created a new ticket #1044 to attempt to summarize the discussions in this PR & link back to your comments.  I've scheduled that work for 7.1, as it looks like similar work will already be happening then.  (Please feel free to edit/update this new ticket if I've not summarized the problem accurately.)

With those two tickets, I'm +1 merging this as is.  I think we all agree this needs more work in the future, but unfortunately that work is out of scope for 7.0 (we simply don't have the time for it right now and the current solution is "good enough" until we do this larger refactor of MyDSpace).

@artlowel : With the creation of the above tickets, are you OK with merger of this PR? If you have final objections, let me know. Otherwise I'll move to merge this tomorrow & we'll readdress it in 7.1.

tdonohue:
Because this is at +1 with no objections and has follow-up tickets created, I'm merging this for the Beta 5 release.  While we all agree more work is necessary (and has been scheduled for 7.1), there is also agreement that this PR works as described/planned for 7.0.

PR #1012, opened 2021-02-04
Translate new lines to Hungarian

akoscomp:
## Description
Translate newly added English lines to Hungarian


bram-atmire (approved):
Built & deployed locally, no problems detecting when switching to Hungarian (magyar) and going through different pages.

PR #1040, opened 2021-03-04
Environment files merging with overwriteMerge

Raf-atmire:
## References
* Fixes https://github.com/DSpace/dspace-angular/issues/909

## Description
This PR ensures that you can merge the common environment with the dev environment. Any values that are specified in the dev environment, will be getting priority over the common profile. If any properties aren't present in the dev file, they'll be taken from the common file.
This PR also ensures that lists are taken from the dev file as a whole if they're present, this means that lists like the "Languages" list, won't be merged anymore. All the values will be taken from the dev file and the list in the common file will be ignored if the list is present in the dev file.

We decided against converting the arrays to objects as that'd mean that we'd have to specifically set each object to null if we didn't want to include it anymore. For example if we're specifying browse indices, with this code we'll have to specify the list of browse indices that we want to use in the dev file. If we were to convert it into objects, we'd have to copy all the browse indices to the dev file and set the ones that we don't want to use to null. This feels like a lot more overhead and more error-prone than just using the list that you want.

## Instructions for Reviewers

List of changes in this PR:
* Now using overwriteMerge for the merging of environment files

Please test this PR by setting up a common environment and a dev environment.
* Configure the property "logDirectory" in the common environment with the value '.'
* Leave out the property "logDirectory" in the dev environment
* Configure the property "debug" in the common environment with the value "false"
* Configure the property "debug" in the dev environment with the value "true"
* Configure a list of languages in the common environment
* Configure a different list of languages in the dev environment

Now run the merge script and verify:
* The logDirectory is still set to the value in the common environment
* The debug property is set to true, the value in the dev environment
* The list of languages is set to exactly what's in the dev environment

## Checklist

### DIFF
--- a/scripts/set-env.ts
+++ b/scripts/set-env.ts
@@ -10,7 +10,7 @@ const targetPath = './src/environments/environment.ts';
 const colors = require('colors');
 require('dotenv').config();
 const merge = require('deepmerge');
-
+const mergeOptions = { arrayMerge: (destinationArray, sourceArray, options) => sourceArray };
 const environment = process.argv[2];
 let environmentFilePath;
 let production = false;
@@ -45,10 +45,10 @@ const processEnv = {
 } as GlobalConfig;
 
 import(environmentFilePath)
-  .then((file) => generateEnvironmentFile(merge.all([commonEnv, file.environment, processEnv])))
+  .then((file) => generateEnvironmentFile(merge.all([commonEnv, file.environment, processEnv], mergeOptions)))
   .catch(() => {
     console.log(colors.yellow.bold(`No specific environment file found for ` + environment));
-    generateEnvironmentFile(merge(commonEnv, processEnv))
+    generateEnvironmentFile(merge(commonEnv, processEnv, mergeOptions))
   });
 
 function generateEnvironmentFile(file: GlobalConfig): void {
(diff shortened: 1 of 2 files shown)


tdonohue (changes requested):
@Raf-atmire : I tested this today by creating a small `environment.dev.ts` that looks like this (this is the entire file):
```
export const environment = {
  rest: {
    ssl: false,
    host: 'localhost',
    port: 8080,
  },
  debug: true,
  languages: [{
    code: 'en',
    label: 'English',
    active: true,
  }, {
    code: 'de',
    label: 'Deutsch',
    active: true,
  }, {
    code: 'cs',
    label: 'Čeština',
    active: true,
  }],
};
```

Then, I ran `yarn start:dev`.  I found the `rest` settings worked great.  The `debug:true` also worked.  

However, my limited list of 3 `languages` didn't seem to work. I still see all 9 languages listed in the globe dropdown in the header.  Did I misunderstand how this should work?

Raf-atmire:
Hi @tdonohue , I'm having troubles reproducing the issue. I'm using the environment.dev.ts like you've described, combined with the environment.common.ts within the git repository.
After running both `yarn run config:dev` and `yarn run start:dev` I can verify that the languages list in environment.ts holds only the 3 languages specified in the environment.dev.ts file. If I leave the languages list out of that dev file, the full list is included as expected.
I'm currently testing this with yarn version 1.22.10 and node version v12.18.3
Could you provide more information as to how I might be able to reproduce this please? Thanks!

tdonohue (approved):
👍 @Raf-atmire : My apologies...I'm somehow no longer able to reproduce my prior problem either.  My assumption therefore is either I made a mistake in testing (maybe I was running `yarn start` which obviously won't use `environment.dev.ts`) or a recent upgrade of Yarn or Node fixed the issue (I'm on Yarn v1.22.4 and Node v14.16.0).

In any case, this now works for me using the same test that I detailed in [my above comment](https://github.com/DSpace/dspace-angular/pull/1040#pullrequestreview-604630366).  Therefore, I'm approving this as-is and going ahead an merging it.  Sorry for the confusion!

PR #1223, opened 2021-06-04
Updated version of the FR wordings

davidatmire:
## Description
Since new EN wordings were added since the first version of the FR wordings, these wordings were to be translated in `fr.json5`.
No other files than `fr.json5` are modified in this PR.


bram-atmire (approved):
Added one missing comma at the end of a message and it has now built successfully.

PR #1318, opened 2021-09-16
Cst 4499

### LINKED ISSUE #776: Create new Item Version
From release plan spreadsheet

Estimate from release plan: 14 hours
Expressing interest: none

No additional notes

davidenegretti-4science:
## References
* Fixes #776 
* Requires DSpace/DSpace#3371

## Description
This feature allows to manage item versions. A new version can be created from item's page, and all versions can be managed (create, delete, edit summary) from the "Version History" tab in "Edit Item" page.

#### Item's page
* A **Create new version** button has be added

#### Version's page
* A version page `/items/version/:versionID` has been added
* This page redirects to the corresponding item's page

#### Version History table
* The **Item** column (containing the item's handler and the link to the item page) has been removed
* The **Version** column now contains the link to the version's page (redirecting to the corresponding item's page)
* An **Action** column has been added, containing the buttons to manage the version
* The **Action** column is only displayed in **Edit Item** page

#### Authorization features
* All actions on versions require authorization
* The authorization features are:
  * `canCreateVersion`
  * `canEditVersion`
  * `canDeleteVersion`

#### Data services
* New methods have been added to the following services:
  * `VersionHistoryDataService`

### Tests

All tests are still TBD

## Checklist


tdonohue (changes requested):
@davidenegretti-4science : I tested this with the backend PR today. While some of the basics work, the behavior is confusing to me & it's hard to understand.  Here's what I see in my Chrome browser:

1. Login as an Admin
2. Visit an Item
3. Click "Create new version" button.  I see a green "successful" message, but I'm not brought to the new version. This means I don't know how to get to the new version.
4. I have to go to MyDSpace and find the new version in my workspace, but it's unclear which one is a versioned document. (I expected a label or something to note it's new version, instead it just has the same name as the old version.  So, I was stuck searching for the document in MyDSpace to find it.)

Additionally ,I went through that same process again, as I thought the version wasn't created. The second time I click the "Create new version" button, I got no green message or any feedback (it's as if the button isn't working). When I look at the Chrome DevTools, I see this is because the backend returned a 422 error because a new version already existed in my workspace.  This error should be displayed to the user in some manner so they realize it has occurred.

I haven't reviewed all the code here yet, but I wanted to report these initial usability issues that I ran into.

davidenegretti-4science:
@tdonohue we could redirect the user to the "Edit item" page as soon as the item has been created.
As for the ambiguity in the MyDspace, the behaviour is the same as in DSpace 6, and we could add a label as a successive improvement.

tdonohue:
@davidenegretti-4science : A redirect to the in progress submission seems like a good enough approach.  My main concern was being able to find the new version I created (as I found it hard to do so).  If I'm automatically redirected to the new version, then that solves my concerns.

I also think the error handling needs to be improved in this PR.  So, for example, if the backend throws a 422 during creation, we know that means a new in-progress version already exists (as that's what is documented in our REST Contract: https://github.com/DSpace/RestContract/blob/main/version.md#create-version).  Therefore, if a 422 is thrown, it should display a red error message telling the user there is already an in-progress version, rather than doing nothing.

davidenegretti-4science:
@tdonohue I have completed the requested modifications

tdonohue (changes requested):
@davidenegretti-4science : Thanks for the updates!  I re-tested this today, and the previous feedback is all **fixed**!  So, it's much easier to create a new version & I see the new version button is grayed out if there's an in-progress version.

That said, I found a **new bug**.  I'll admit, I'm not sure if this bug is on the frontend or backend code, so you may need to work with @Micheleboychuk to figure out where the problem is.  Essentially, **if there is an in-progress submission**, then non-Admin users can no longer view item version history.  Here's what I see:
1. Start with `versioning.item.history.view.admin=false` and an Item that already has two versions.
2. Login as an Admin
3. Verify on the item page that two versions show in the Version History.  
4. As an Anonymous user (in a new browser or tab in Incognito mode), access that same Item page. Verify the anonymous user also sees two versions in Version History.
5. As Admin, go back to the Item, and create another new version.  This creates a third version as an in-progress submission.
6. Immediately go back to your Anonymous user.  Reload the Item page.  The Version History table will have disappeared entirely.  **(THIS IS A BUG.  Proper behavior should be the Version History table should still appear with just two versions, as those are the ones which are public / in archive)**
7. Go back to your Admin user & deposit the in-progress version, making it the latest public version.
8. Immediately go back to your Anonymous user.  Reload the Item page.  The Version History table is back, now with 3 versions.

As noted, I'm not sure if the bug is on the UI side...it might be that the Item History table information is sometimes not being sent from the REST API. 

Overall, I approve this PR.  This is the last bug that I see.  Everything else seems to be working well.

davidenegretti-4science:
@tdonohue thanks for the feedback.

I tried to reproduce the bug, but everything works fine for me. That's what I did:

1. I added `versioning.item.history.view.admin=false` to the configuration
2. I logged in as administrator
3. I opened an existing item (created as administrator) which already had 89 versions
4. I created a new version 90 and I did not click on "Deposit"
5. I went back to item's page and I checked that the version history was correct (it correctly shows version 90 and it marks version 89 as current version)
6. I opened the item page (version 89) in a new incognito window and the version history is correctly shown (without version 90)
7. I also made the additional check that version 90 can not be reached from its URL when I'm not logged in

That's what I see:

...as administrator:

versions as admin

...as anonymous user:

versions as anonymous

tdonohue:
@davidenegretti-4science : Thanks for testing on your end. Since you cannot reproduce, it's possible that I made a mistake in my testing here (or had something cached that affected this)... I'll do a full rebuild of these two PRs & see if I can reproduce the issue again today.  Will let you know what I find.

tdonohue:
@davidenegretti-4science : I've re-tested today and I'm still able to reproduce the error, but ONLY if I run `yarn start` (i.e. SSR enabled) -- and even then, sometimes the first anonymous access "works". But, if I click the reload button in my browser (which will trigger SSR), the entire Version History table disappears for the anonymous user.

If I run `yarn start:dev` then the bug is no longer reproducible -- the Version History table always appears for an anonymous user.  

This behavior implies to me that this is an issue with server-side rendering (SSR) in the UI codebase and not an issue on the backend. For some reason the Version History table is not always displayed via SSR.  Ideally, it would be.

davidenegretti-4science:
@tdonohue I've been able to reproduce the error in production mode and I fixed it

tdonohue (approved):
👍 Looks great now @davidenegretti-4science !  It now works fully via `yarn start`.

I also tested that turning versioning off works (`versioning.enabled=false`) and restricting the version history works (`versioning.item.history.view.admin=true`), and those both work seamlessly as well.

Thanks for your hard work on this.  As soon as @artlowel has a chance to review, I think this is ready to be merged.

tdonohue:
@davidenegretti-4science or @atarix83  : Please let me know when this is ready for re-review (in terms of spec updates).  As a sidenote, this now has a minor merge conflict after I merged #1327

PR #1328, opened 2021-09-23
Control collection harvest from the UI

YanaDePauw:
## References
* Fixes #[761]
* Requires DSpace/DSpace#[7952]

## Description
This PR adds additional controls to the collection UI to harvest a collection.
These controls will allow a user to:
* verify the current configuration
* start an import with the present configuration
* reset and reimport a collection

## Instructions for Reviewers
To test the new functionality, a collection will need to set up with harvesting configuration.
The support for setting the configuration is already present.

To set up a collection for harvesting, go to an edit collection page and go to the "Content Source" tab.
Check the checkbox and fill in the form with a valid source to harvest items from. This can be any DSpace with OAI enabled.

After filling in the OAI information and saving it, verify that the harvest control buttons become available and test their functionality.

**Test configuration**

Click the "Test configuration" button. Verify that after the check, a notification is shown with the output for the test.

**Import now**

Click the the "Import now" button. This will start the import which can be a quite lengthy.
When the process is finished, a notification will be shown and the "Harvest start time" will be updated.

**Reset and reimport**

The "reset and reimport" button is not implemented yet since a command will still be added to the backend that will perform both a purge and a new import.

## Checklist

### DIFF
--- a/src/app/collection-page/edit-collection-page/collection-source/collection-source-controls/collection-source-controls.component.html
+++ b/src/app/collection-page/edit-collection-page/collection-source/collection-source-controls/collection-source-controls.component.html
@@ -1,0 +1,35 @@
+<div *ngVar="(contentSource$ |async) as contentSource">
+    <div class="container-fluid" *ngIf="shouldShow">
+        <h4>{{ 'collection.source.controls.head' | translate }}</h4>
+        <div>
+            <span class="font-weight-bold">{{'collection.source.controls.harvest.status' | translate}}</span>
+            <span>{{contentSource?.harvestStatus}}</span>
+        </div>
+        <div>
+            <span class="font-weight-bold">{{'collection.source.controls.harvest.start' | translate}}</span>
+            <span>{{contentSource?.harvestStartTime ? contentSource?.harvestStartTime : 'collection.source.controls.harvest.no-information'|translate }}</span>
+        </div>
+        <div>
+            <span class="font-weight-bold">{{'collection.source.controls.harvest.last' | translate}}</span>
+            <span>{{contentSource?.lastHarvested ? contentSource?.lastHarvested : 'collection.source.controls.harvest.no-information'|translate }}</span>
+        </div>
+
+        <button class=" btn btn-secondary"
+                [disabled]="!(isEnabled)"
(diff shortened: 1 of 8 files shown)


tdonohue (changes requested):
@YanaDePauw : I gave this a test today alongside your backend PR.  The basics seem to *almost* work, but I've hit a few bugs along the way.

1. _Minor usability issue:_  When you click the "Test configuration" button (or really any of the buttons in the UI), as a user it's impossible to tell that anything is happening.  The button stays enabled and eventually a response comes back. 
    * Better behavior here would be to either send the user to the "Process" that is running behind the scenes... or, alternatively gray out the button and provide a message saying "running.." (or similar) which appears until the response comes back.  The second approach may be more of the user experience you were going for, as it seems like you are trying to keep the user on the "Edit Collection > Content Source" page while the process is kicked off.
2. _Bug:_ The "Import now" button creates a Process which immediately fails.  The reason is that you are passing the "eperson" parameter which you (correctly) disallowed to be used from the UI...it's only valid from CLI.
3. _Bug:_ Strangely, when I go to the Processes page & click "+New" & select "harvest", I see that **every single parameter** requires a `<value>`.  This means I cannot even run a simple `--help` without passing it a dummy value like `--help asdf`.  I'm not sure why this is occurring offhand, as this does NOT occur from the CLI, and it seems specific to this new Process (all other existing processes work fine).

I haven't done a full code review here yet, but I wanted to report the issues I've found so far.  Thanks!

LucaGiamminonni (commented):
Hi @YanaDePauw, I had a look at the changes and they seems more or less ok to me. In addition to the things highlighted by @tdonohue  I would have nothing else to add. From the harvest page I was able to test the connection, perform an import and perfom a reset and import.

YanaDePauw:
@tdonohue 
Concerning the "The Bug in Processes still exists.", this seems to be due to the unintuitive code in the backend that handles the the script parameters. I currently quick fixed this in the [backend PR](https://github.com/DSpace/DSpace/pull/7952). 

I also created a general [issue](https://github.com/DSpace/DSpace/issues/7992) because based on how the code works right now, this issue may occur again in the future when new scripts are added/ported.

PR #1459, opened 2021-12-16
Fix to collection selection window not loading successfully.

### LINKED ISSUE #1421: DSpace 7: When creating a new item via drag&drop or browse the collections load unsuccessfully
**Describe the bug**
When I created an item via drag&drop or browse, the collection selection window did not load successfully.

**To Reproduce**
1. Drag&drop a file or browse
grafik

2. Check the collection selection window. It keeps loading.
grafik

**Expected behavior**
The collection selection window should display the collections so the user can choose a collection for his submission.

santit96:
* Fixes #1421

## Description
* In collection dropdown component, moved the return statement that returned the collection list outside the condition that checks if it's the final page. 
* Also changed the condition that checks final page, it wasn't necessary to sum the elements per page.
* Deleted !(isLoading | async) condition that didn't allowed the show collection list if there were new collections loading
* Changed infiniteScrollDistance and infiniteScrollThrottle parameters so the load of the collection is more fluent. I took as a reference the AuthorizedCollectionSelectorComponent. 

Both CollectionDropdownComponent and AuthorizedCollectionSelectorComponent had similar behaviour, perhaps they could be combined into a single component in the future

## Instructions for Reviewers
* Go to My Dspace page 
* Create a new item via drag&drop or browse
* The collections should load without errors
Alternatively
* Edit an in progress submission
* Try to change collection of that submission
* The collections should load without errors

### DIFF
--- a/src/app/shared/collection-dropdown/collection-dropdown.component.html
+++ b/src/app/shared/collection-dropdown/collection-dropdown.component.html
@@ -11,8 +11,8 @@
      aria-labelledby="dropdownMenuButton"
      (scroll)="onScroll($event)"
      infiniteScroll
-     [infiniteScrollDistance]="5"
-     [infiniteScrollThrottle]="300"
+     [infiniteScrollDistance]="1.5"
+     [infiniteScrollThrottle]="0"
      [infiniteScrollUpDistance]="1.5"
      [fromRoot]="true"
      [scrollWindow]="false"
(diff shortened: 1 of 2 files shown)


tdonohue (approved):
👍 Thanks @santit96 , this fixes the previously reported bug.  I was able to reproduce the bug again on the `main` branch, and found that your PR fixes it (and no error appears in the DevTools console anymore).

PR #1471, opened 2021-12-22
Minor readme configuration corrections

wwelling:
Just quick readme correction.


tdonohue (approved):
👍 Looks correct to me. Thanks @wwelling !

PR #1561, opened 2022-03-17
Browse clarify filter functionality

MarieVerdonck:
## References
- Fixes https://github.com/DSpace/dspace-angular/issues/1379
- Related to https://github.com/DSpace/DSpace/issues/8203

## Description
UX related changes for the browse to clarify the startsWith function is a filter.

## Instructions for Reviewers
- 'Go back' button "All browse results" that when filtering you can go back to all browse results without having to send empty query
  - Reset on both `startsWith` (so query to filter) and `value` (when a value like subject is selected and it shows items based on that value)
- Placeholder text search bar changed to indicate it's a filter
- Hover tooltip on next button indicating no more next pages of results available

The alphabet row was removed (see comment here: https://github.com/DSpace/DSpace/issues/8203#issuecomment-1064263899) :

- The "0-9" option doesn't work with filters, since it send `startsWith=0`
  - It could for example send `startsWith=[0 TO 9]` but this is not immediately supported REST side
    - Ex: [/discover/browses/subject/entries?sort=default,ASC&page=0&size=20&startsWith=5](https://api7.dspace.org/server/#/server/api/discover/browses/subject/entries?sort=default,ASC&page=0&size=20&startsWith=5) => has 1 result
      [startsWith=[0 TO 9]](https://api7.dspace.org/server/#/server/api/discover/browses/subject/entries?sort=default,ASC&page=0&size=20&startsWith=%5B0%20TO%209%5D) => No results
      [startsWith=[0||1||2||3||4||5||6||7||8||9]](https://api7.dspace.org/server/#/server/api/discover/browses/subject/entries?sort=default,ASC&page=0&size=20&startsWith=%5B0%7C5%5D) => No results
- The alphabet letters itself only support latin alphabet, could be changed to be dependent on configured list of 'letters' based on language selected, but was outside of scope for this issue

## Testing 

- Test [[original issue](https://github.com/DSpace/dspace-angular/issues/1379)](https://github.com/DSpace/dspace-angular/issues/1379) case => Should be more clear it is filter
- Go to value browse (eg subject) => Filter by typing in prefix, test the go back button
- Go to value browse (eg subject) => Go to list of items with that subject by pressing that subject, test the go back button

## Checklist


tdonohue (changes requested):
@MarieVerdonck and @YanaDePauw : Overall, this works well and improves the usability of these Browse by pages by making it clear that they are filtering results instead of doing a "jump to". So, good job overall!

That said, I do have a few minor usability improvements to suggest in this PR.

1. When filtering by a letter (e.g. "startsWith=T"), there's no information provided in the UI as to what letter/text you are filtering by.  I would expect the top of the page to somehow let you know what you've filtered by. Here's what it looks like in this PR:
filtered
    * My expectation here would be that either the "Filter results..." input field should _not be empty_ (and instead still display the text you entered), or some sort of text should appear on the page that says "Filtered by '[text]'" (or similar).  Currently, as you can tell from my screenshot, you cannot tell what I've filtered on. I'd expect it to be clear to the user in some way.
2. The "Browsing by Issue Date" page still has "Jump to..." in the description.  It should likely say "Filter by a specific year or month" (or just "Filter results by date").  Here's what it currently says: 
date
3. (Optional) If you notice in that previous "Browsing by Issue Date" screenshot, the gear button (to change settings) appears one line *below* the filter input box.  This occurs on all Browse by pages & it looks odd (as there's essentially a blank gap before the results appear). We may want to consider moving the gear up one line.  However, I'm not sure if this change would conflict with pagination work going on in #1552 ... so, we could leave this change for that PR or other future cleanup.

MarieVerdonck:
@tdonohue Rebased & fixes done after merge of https://github.com/DSpace/dspace-angular/pull/1552 and [first two feedback points](https://github.com/DSpace/dspace-angular/pull/1561#pullrequestreview-921925413) were addressed. 
For the third, the gear is in-line with the "Now showing..." info from other PR.
As stated in [comment there](https://github.com/DSpace/dspace-angular/pull/1552#pullrequestreview-930323364) we can enable the `<<|1|2|3|>>` view by setting `showPaginator` to true in `BrowseByComponent` [here](https://github.com/DSpace/dspace-angular/pull/1561/files#diff-0d2cf76c79d3ec1ba71ee1d66f7867e33096bc92c807bcc6528422a4c68d399bR73)

tdonohue (changes requested):
@MarieVerdonck : Overall this looks good and works well.  I just found a few minor issues:

1. The Page Title (i.e. `<title>` tag in HTML `<head>`) on every Browse by page now includes `{{ startsWith}}`.  For example, on the Browse by Author page, the Title looks like this: `<title>DSpace Angular :: Browsing  by Author{{ startsWith }} </title>`.   While this is minor, it looks odd & might impact SEO.
2. When using Browse by Title & entering something into the filter textbox, the `, starting with [filter-text]` text is missing from the page heading.  This text appears fine on Browse by Date/Subject/Author, but it's missing from the Browse by Title page.
3. If you move to the *third* page of any Browse results (e.g. Browse by Title) and **then** use the textbox to filter by something that returns _less than three pages of results_ then you will end up with a "No items to show" page.  It looks like we need to remove/reset the `page` parameter whenever new filter text is added...otherwise if you are on `page=3` when adding filter text, you'll go to `page=3&startsWith=[filter-text]`... and that may or may not have results even if some items match `[filter-text]`.

Finally, I personally prefer to set `showPaginator = true` in the BrowseByComponent.  I feel the usability is improved as it lets you jump to later page more easily than continually pressing the "Next" button.  I tested the behavior of `showPaginator=true`, and I found no additional bugs when it's enabled.  So, if you agree, we can just switch this value to `true` in this PR.  However, if you have concerns about that, we can wait to do this in a future PR.

tdonohue (approved):
👍 Thanks @MarieVerdonck & @YanaDePauw  !  I've retested today and all my previously reported issues are resolved.  So, this looks good enough to merge!

PR #1574, opened 2022-03-22
Pad date range sliders to fix search facet animation issues

ybnd:
## References
* Originally noticed when working on #1138

## Description

Currently, date range sliders in search facets cover the full width of the facet, which causes the handles to jut out a little bit to the left/right.
This works fine, but causes those handles to get pushed behind the facet's padding during the collapse/expand animation.

This PR fixes this by padding the sliders so that the ends of handles at the min/max positions are flush with the slider itself

https://user-images.githubusercontent.com/31547038/159553028-db9f5275-b4fe-4d0f-94b3-67d3643ad28b.mov

## Instructions for Reviewers

Confirm that
* the slider handles are no longer cut in half during the animation when positioned at the ends of the slider
* the date range sliders still work as they did

## Checklist

### DIFF
--- a/src/app/shared/search/search-filters/search-filter/search-range-filter/search-range-filter.component.scss
+++ b/src/app/shared/search/search-filters/search-filter/search-range-filter/search-range-filter.component.scss
@@ -28,4 +28,13 @@
   .noUi-connect {
       background: var(--ds-slider-color);
   }
+
+  .noUi-target {
+    padding: 0 calc((var(--ds-slider-handle-width) / 2) - 1.5px);
+    margin: 0 1px;
+  }
+  .noUi-connect {
+    margin: 0 calc(-1 * (var(--ds-slider-handle-width) / 2));
+    width: calc(100% + var(--ds-slider-handle-width));
+  }
 }


ybnd:
One possible caveat: the current look is kinda nice too



But IMO the animation fix is worth it.

tdonohue (commented):
@ybnd : Overall, I agree that the slight alignment issue on `main` looks a little odd.  However, I found that with this tiny PR, the right slider cannot be moved all the way to the right side.  Here's a screenshot which shows a tiny sliver of green bar on the *right* of the right slider:

after

If we can make it so the bar doesn't extend past the right slider, then this PR would be fine with me.  Otherwise, I feel the current behavior is slightly "better" (even if the range slider is slightly misaligned with the start/end date fields above it).

ybnd:
Apparently this happened only at _some_ viewport widths, thanks for catching it!

When resizing slowly you can see the position of the element jitter a little, and once in a while the slider juts out from under one of the handles (also on the left side sometimes!).
There's a tradeoff between this and the sides of the handles getting clipped off by the animation, getting it to avoid both issues took a bit of time and pixel peeping.

With the current CSS I haven't been able to replicate either problem.

tdonohue (approved):
👍 Thanks @ybnd !  I've tested this PR with your updates, and the new CSS works perfectly for me as well!  I tried it in several browsers (Chrome, Firefox, MS Edge) and also various sized windows.  It all looks good now.

PR #1582, opened 2022-03-27
Updated German translation

louiserumpf:
also corrected comment search.search-form.placeholder --> home.search-form.placeholder


tdonohue (approved):
👍 Thanks @louiserumpf for the updates. I gave it a quick look (using Google Translate) and it looks good to me.

PR #1601, opened 2022-04-14
Display the access status badges

### LINKED ISSUE #1566: Displaying item access status in search results in DSpace 7
In DSpace 7, a user can only know if an item is open access or not by opening the item record. 

User must be able to know quickly in the UI if the main file of an item is Open Access or not. One first step would be to display the access status on search results.  

The item metadata dc.date available cannot be rely upon to determine access status as it is not consistently used by DSpace to display the Open Access date of the item (see https://groups.google.com/g/dspace-tech/c/k7qCSJVhRrU and https://github.com/DSpace/DSpace/issues/6583). 

In DSpace, access restrictions are set on bitstream. As an item can have one to many bitstream associated to it with different access restrictions, access status on the item should be based on the main file of the item (i.e. full text of an article, thesis, etc.). Hypothesis is that the first file is the main file in DSpace original bundle.  

4 access status can be displayed based on the restriction value used in DSPace 7 workflow:
<html>
<body>

Statut (anglais) | Description
-- | --
Open Access | Value Open access  in DSpace Workflow is selected for the file
Embargo | Value embargo in DSpace Workflow is selected for the file
Restricted | Value Administrator in DSpace Workflow is selected for the file
Metadata only | No bitstream associated

</body>
</html>

The following process can be used to determine the access status:  

1. Look at all READ policies of the primary/first bitstream, 
2. If one policy is bond to the Anonymous group and if the current date is between the start date and the end date of this policy, then item status is OpenAccess.  
3. Otherwise: if one policy for the Anonymous group has a start date inferior to the specific predefined date set in the configuration (for infinite embargo), then, it is an embargo item. 
4. If the start date is superior or equal the date specified in the configuration, then, the item is restricted. 
5. If no policy for the Anonymous group is set, the item is considered Restricted. 
6. Finally, if the item has no bitstream, the Item status will be Metadata only. 

A couple of parameters would be added to the configuration. First off, the access status display in search results can be activated or deactivated as a configuration parameter in DSpace. Next, the forever date would also be added to the configuration with a year, a month and a day. 

Access status will be displayed as tags as illustrated below:  
access-status

nibou230:
## References
* Fixes #1566
* Requires DSpace/DSpace#8247

## Description
Inclusion of a new badge to display the access status for every item.

## Instructions for Reviewers
With the configuration, the badge can be easily toggled on or off. By default, the badge is not displayed for now, as the status is recalculated each time an item is displayed and this can impact the performance. In our tests, it's fine for our use cases, but it may not be for other institutions.

List of changes in this PR:
* Addition of a new configuration `showAccessStatuses` parameter in the Item section.
* An `AccessStatusObject` to map the return of the new REST endpoint.
* Inclusion of a new data service `AccessStatusDataService`.
* A new badge component for displaying the access status.
* Some basic unit tests for this new component and the new service.
* Renamed item-page configuration for item configuration to make it generic.

**Do not forget to turn the feature on with the configuration `showAccessStatuses: true`**

## Checklist

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -13,6 +13,8 @@ ui:
   rateLimiter:
     windowMs: 60000 # 1 minute
     max: 500 # limit each IP to 500 requests per windowMs
+  # Show the file access status in items lists
+  showAccessStatuses: false
 
 # The REST API server settings
 # NOTE: these must be 'synced' with the 'dspace.server.url' setting in your backend's local.cfg.
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -13,8 +13,6 @@ ui:
   rateLimiter:
     windowMs: 60000 # 1 minute
     max: 500 # limit each IP to 500 requests per windowMs
-  # Show the file access status in items lists
-  showAccessStatuses: false
 
 # The REST API server settings
 # NOTE: these must be 'synced' with the 'dspace.server.url' setting in your backend's local.cfg.
@@ -163,10 +161,12 @@ browseBy:
   # The absolute lowest year to display in the dropdown (only used when no lowest date can be found for all items)
   defaultLowerLimit: 1900
 
-# Item Page Config
+# Item Config
 item:
   edit:
     undoTimeout: 10000 # 10 seconds
+  # Show the item access status label in items lists
+  showAccessStatuses: false
 
 # Collection Page Config
 collection:
--- a/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.html
+++ b/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.html
@@ -18,7 +18,7 @@
       </span>
       <div class="card-body">
           <ds-type-badge *ngIf="showLabel" [object]="dso"></ds-type-badge>
-          <ds-access-status-badge *ngIf="showAccessStatus" [uuid]="dso.uuid"></ds-access-status-badge>
+          <ds-access-status-badge [uuid]="dso.uuid"></ds-access-status-badge>
           <ds-truncatable-part [id]="dso.id" [minLines]="3" type="h4">
               <h4 class="card-title" [innerHTML]="firstMetadataValue('dc.title')"></h4>
           </ds-truncatable-part>
--- a/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.html
+++ b/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.html
@@ -18,6 +18,7 @@
       </span>
       <div class="card-body">
           <ds-type-badge *ngIf="showLabel" [object]="dso"></ds-type-badge>
+          <ds-access-status-badge *ngIf="showAccessStatus" [uuid]="dso.uuid"></ds-access-status-badge>
           <ds-truncatable-part [id]="dso.id" [minLines]="3" type="h4">
               <h4 class="card-title" [innerHTML]="firstMetadataValue('dc.title')"></h4>
           </ds-truncatable-part>
--- a/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.html
+++ b/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.html
@@ -18,7 +18,7 @@
       </span>
       <div class="card-body">
           <ds-type-badge *ngIf="showLabel" [object]="dso"></ds-type-badge>
-          <ds-access-status-badge [uuid]="dso.uuid"></ds-access-status-badge>
+          <ds-access-status-badge [item]="dso"></ds-access-status-badge>
           <ds-truncatable-part [id]="dso.id" [minLines]="3" type="h4">
               <h4 class="card-title" [innerHTML]="firstMetadataValue('dc.title')"></h4>
           </ds-truncatable-part>
--- a/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.ts
+++ b/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.ts
@@ -6,6 +6,7 @@ import { SearchResultGridElementComponent } from '../../search-result-grid-eleme
 import { Item } from '../../../../../core/shared/item.model';
 import { ItemSearchResult } from '../../../../object-collection/shared/item-search-result.model';
 import { getItemPageRoute } from '../../../../../item-page/item-page-routing-paths';
+import { environment } from 'src/environments/environment';
 
 @listableObjectComponent('PublicationSearchResult', ViewMode.GridElement)
 @listableObjectComponent(ItemSearchResult, ViewMode.GridElement)
@@ -23,9 +24,14 @@ export class ItemSearchResultGridElementComponent extends SearchResultGridElemen
    * Route to the item's page
    */
   itemPageRoute: string;
+  /**
+   * Whether to show the access status badge or not
+   */
+  showAccessStatus: boolean;
 
   ngOnInit(): void {
     super.ngOnInit();
     this.itemPageRoute = getItemPageRoute(this.dso);
+    this.showAccessStatus = environment.ui.showAccessStatuses;
   }
 }
--- a/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.ts
+++ b/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.ts
@@ -6,7 +6,6 @@ import { SearchResultGridElementComponent } from '../../search-result-grid-eleme
 import { Item } from '../../../../../core/shared/item.model';
 import { ItemSearchResult } from '../../../../object-collection/shared/item-search-result.model';
 import { getItemPageRoute } from '../../../../../item-page/item-page-routing-paths';
-import { environment } from 'src/environments/environment';
 
 @listableObjectComponent('PublicationSearchResult', ViewMode.GridElement)
 @listableObjectComponent(ItemSearchResult, ViewMode.GridElement)
@@ -24,14 +23,9 @@ export class ItemSearchResultGridElementComponent extends SearchResultGridElemen
    * Route to the item's page
    */
   itemPageRoute: string;
-  /**
-   * Whether to show the access status badge or not
-   */
-  showAccessStatus: boolean;
 
   ngOnInit(): void {
     super.ngOnInit();
     this.itemPageRoute = getItemPageRoute(this.dso);
-    this.showAccessStatus = environment.item.showAccessStatuses;
   }
 }
--- a/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.ts
+++ b/src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.ts
@@ -32,6 +32,6 @@ export class ItemSearchResultGridElementComponent extends SearchResultGridElemen
   ngOnInit(): void {
     super.ngOnInit();
     this.itemPageRoute = getItemPageRoute(this.dso);
-    this.showAccessStatus = environment.ui.showAccessStatuses;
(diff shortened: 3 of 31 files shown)


nibou230:
Hi guys, I was exploring the administration and I remarked that there was a couple of grids and lists where the access status badge was missing. Would you mind if I add some code to address that?

tdonohue:
@nibou230 : You are welcome to update your PR as you find issues you want to resolve. I have assigned a few reviewers as well today who will try to give this a test/review in the next week or so.

Just as a sidenote, this PR currently has minor merge conflicts.  So, it'd also be helpful to update/rebase this against the latest `main` codebase and resolve those conflicts.  (You may already know this... but while conflicts are usually infrequent, they can occur as other PRs are merged to `main` if those PRs touch some of the same files as yours does.)

nibou230:
@tdonohue I did a rebase to merge the conflicts, but for some reason the integration tests are failing. I don't know why and what I could have done to make it happen. It was fine just before the rebase with the commit `[Added the badge to some components for coherence]`.

tdonohue:
@nibou230 : We have a few tests which are known to be a bit "flakey" at times (we've yet to track down why). Restarted the tests in this PR, as it looks like that's what you hit -- all tests succeeded in this PR when it ran on Node 14, but a few failed on Node 12. Usually if tests are really failing they will fail on both v12 and v14.  

Long story short, I expect the test failures were random & they should succeed on this second build.

tdonohue (changes requested):
@nibou230 : This PR also looks great and works well.  I've verified the access status labels appear when enabled on all search/browse pages.  Currently they only appear for generic Items and Publication entities (no other entities), but that seems good enough for now.

One minor comment though about the configuration below. I think we should change where it's located.

The rest of the code here looks reasonable, though I'd ask @artlowel to give it a quick code review, as he might notice things that I overlooked (if anything).

config/config.example.yml
```diff
@@ -13,6 +13,8 @@ ui:
   rateLimiter:
     windowMs: 60000 # 1 minute
     max: 500 # limit each IP to 500 requests per windowMs
+  # Show the file access status in items lists
+  showAccessStatuses: false
```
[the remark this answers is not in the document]
nibou230: I agree, but it says `# Item Page Config` on top if this section. Since it's not related to the item's page at all, should we rename it to something more generic (like Item Configuration)? The same problem occurs in the file `item-page-config.interface.ts`. I don't know, it's a bit confusing to me.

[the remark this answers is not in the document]
tdonohue: @nibou230 : Sure, feel free to rename the comments if they are inaccurate now. It seems fine to change this to say `# Item Configuration`.  Obviously, as we add more code, sometimes these minor refactors are necessary for things to make "sense". So, please feel free to suggest/make such minor changes in your PRs & just let us know why you chose to do so (in the PR description or similar)

src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.html
```diff
@@ -18,6 +18,7 @@
       </span>
       <div class="card-body">
           <ds-type-badge *ngIf="showLabel" [object]="dso"></ds-type-badge>
+          <ds-access-status-badge *ngIf="showAccessStatus" [uuid]="dso.uuid"></ds-access-status-badge>
```
MarieVerdonck, line 21: Instead of repeated `*ngIf="showAccessStatus"` and its retrieval from config, can be moved to `ng-container` surrounding code in html of `AccessStatusBadgeComponent`

src/app/shared/object-grid/search-result-grid-element/item-search-result/item/item-search-result-grid-element.component.ts
```diff
@@ -23,9 +24,14 @@ export class ItemSearchResultGridElementComponent extends SearchResultGridElemen
    * Route to the item's page
    */
   itemPageRoute: string;
+  /**
+   * Whether to show the access status badge or not
+   */
+  showAccessStatus: boolean;
```
MarieVerdonck, line 30: Instead of repeated retrieval of `showAccessStatus` from config, can be retrieved and checked in `AccessStatusBadgeComponent`

src/app/shared/object-list/access-status-badge/access-status-badge.component.ts
```diff
@@ -0,0 +1,53 @@
...
+            return [];
+          }
+        }),
+        map((accessStatus: AccessStatusObject) => hasValue(accessStatus.status) ? accessStatus.status : 'unknown'),
+        map((status: string) => `access-status.${status.toLowerCase()}.listelement.badge`),
+        catchError(() => observableOf('access-status.unknown.listelement.badge'))
+      );
+  }
+
+  @Input() set uuid(uuid: string) {
```
MarieVerdonck, line 46: setter for `@input` & getter for component var to use in html can just be component vars:
```
@Input() uuid: string;
accessStatus$: Observable<string>;
```

src/app/shared/object-list/my-dspace-result-list-element/item-list-preview/item-list-preview.component.html
```diff
@@ -2,7 +2,10 @@
   <ng-container *ngIf="status">
     <ds-mydspace-item-status [status]="status"></ds-mydspace-item-status>
   </ng-container>
-  <ds-type-badge [object]="item"></ds-type-badge>
+  <div class="d-flex">
+    <ds-type-badge [object]="item"></ds-type-badge>
+    <ds-access-status-badge *ngIf="showAccessStatus" [uuid]="item.uuid" class="pl-1"></ds-access-status-badge>
```
MarieVerdonck, line 7: See above, avoid repeated *ngIf="showAccessStatus"

src/app/shared/object-list/my-dspace-result-list-element/item-list-preview/item-list-preview.component.ts
```diff
@@ -36,4 +37,12 @@ export class ItemListPreviewComponent {
    */
   @Input() showSubmitter = false;
 
+  /**
+   * Whether to show the access status badge or not
+   */
+  showAccessStatus: boolean;
```
MarieVerdonck, line 43: See above, avoid repeated retrieval of `showAccessStatus`

src/app/shared/object-list/search-result-list-element/item-search-result/item-types/item/item-search-result-list-element.component.html
```diff
@@ -1,4 +1,7 @@
-<ds-type-badge *ngIf="showLabel" [object]="dso"></ds-type-badge>
+<div class="d-flex">
+  <ds-type-badge *ngIf="showLabel" [object]="dso"></ds-type-badge>
+  <ds-access-status-badge *ngIf="showAccessStatus" [uuid]="dso.uuid" class="pl-1"></ds-access-status-badge>
```
MarieVerdonck, line 3: See above, avoid repeated *ngIf="showAccessStatus"

src/app/shared/object-list/search-result-list-element/item-search-result/item-types/item/item-search-result-list-element.component.ts
```diff
@@ -21,9 +22,14 @@ export class ItemSearchResultListElementComponent extends SearchResultListElemen
    * Route to the item's page
    */
   itemPageRoute: string;
+  /**
+   * Whether to show the access status badge or not
+   */
+  showAccessStatus: boolean;
```
MarieVerdonck, line 28: See above, avoid repeated retrieval of showAccessStatus

MarieVerdonck (changes requested):
**Code review**

- Improvement to prevent code duplication: move `showAccessStatus` check (retrieved from config) to `AccessStatusBadgeComponent` itself, so need not be repeated in each call of `ds-access-status-badge`

- Search result appear, then extra labels come after, since `accesStatus` retrieved after other info on item displayed via `/items/{uuid}/accesStatus`

  - Unsure of what is more performant, this or via embed

  - Could be called as embed in `src/app/shared/search/search.component.ts` as  `followLink<Item>('accessStatus', { isOptional: true, shouldEmbed: environment.ui.showAccessStatuses })` but would require:

    - A minimal `@dataService(ACCESSSTATUS) AccessDataService extends DataService<AccessStatusObject>`

    - Would also require in `item.model.ts`

      ```
        @link(ACCESS_STATUS, false, 'accessStatus')
        accessStatus?: Observable<RemoteData<Bitstream>>;
      ```

      and added to `Item._links` 

    - Resulting in call like `/discover/search/objects?sort=score,DESC&page=0&size=10&configuration=default&embed=thumbnail&embed=accessStatus`

  - As it exists now, it can not be used as embed / `followLink('accessStatus')` so the above would have to be added when you want to use it as such in for example the item call, regardless of whether you use it here in the `search.component.ts` for the `/search/objects` call.

**Functionality review**

- Applied to items (no entity type or 'Publiction') 
- In search list/grid view, recent submission list view & browse by title

Cases => all ok:

- With no bitstream/file attached: 'Metadata only'
- With 1 file open access file in ORIGINAL bundle attached: 'Open Access'
- With 1 file with future starte date file in ORIGINAL bundle attached: 'Embargo'
- With future end date 2030-05-07: 'Open Access'
- With admin only file: 'Restricted'
- With 'lease' date in the past so end date in the past => 'Embargo'

paulo-graca (approved):
As long as @tdonohue and @MarieVerdonck comments are addressed, I think it could be merged. This PR is working as expected for me.  Thank you @nibou230 for this great contribution!

My only comment (and I understand the why) is for the need of `showAccessStatuses: false`. It's a matter of personal preference, but I prefer to have it in DSpace.cfg . But is ok as is. It just another concern when setting up and deploying. Perhaps if others agree, it could be addressed it on another stage of the project.

nibou230:
Hi @MarieVerdonck, sounds good!

I added the code to test the embed route. It works fine, but only for the search page by adding the line `followLink<Item>('accessStatus', { isOptional: true, shouldEmbed: environment.ui.showAccessStatuses })`. I guess it will require a similar approach for every component showing the access status?

For now, I can let everything needed to use the access status as an embedded resource while still using the service method. The work required to use the embed route could be done later as I'm starting to run out of time for this feature.

However, I don't know if this is more performant than calling the current service method. To me, it looks identical, but it may vary for some users.

MarieVerdonck (approved):
Thank you @nibou230! It still works well and is now re-usable/used in search as embed (so there is no delay between results shown and the access status tag).

It doesn't look like it's just a matter of using the `followLink` functionality for browse, so I'm good with having it like this. To embed it in browse entries I suspect it would need changes in `BrowseService#getBrowseItemsFor` for followLinks to create a request like `/server/api/discover/browses/title/items?sort=dc.title,ASC&page=0&size=20&embed=accessStatus`. (my-dspace uses same method in the searchservice, so has the followLink you already applied)

tdonohue (approved):
👍 Thanks @nibou230 !  This is working well for me now, and all my prior feedback has been addressed.

I did want to note that I saw what @MarieVerdonck noted as well.  Currently, instead of the `accessStatus` being loaded via the `embed=` parameter, they are all loaded via separate calls to `/items/[uuid]/accessStatus` (one per item).  On a page of 10-20 items, that's 10-20 REST API calls (which doesn't take long).  But, if you load a page of 100 items, it's 100 extra REST API calls which does have a delay.

There would be a major performance improvement (for larger pages) *if we could find a way to use the `embed` param*.   As @MarieVerdonck notes, the backend already "works".  

So, if I call `/server/api/discover/browses/title/items?sort=dc.title,ASC&page=0&size=100&embed=accessStatus` (which loads 100 items, with their accessStatus embedded in the response) it loads in ~2 seconds.  But, if I load a page of 100 items with this current PR, that can take ~9 seconds.

If you feel this would not be easily possible (and it sounds like you implied it may not be), I'm OK with this PR as-is.  It works, and it's disabled by default (and has no performance problems when disabled).  However, we may want to add a note to the docs that it can cause some slowness if users change the page size to >20.

Overall, I approve of this.  If you'd prefer we delay the enhancements for later, we can accept this as-is & create a follow-up ticket to improve performance at a later time.  Thanks!

tdonohue:
Merging as this is at +3 (and just merged the backend).  Thanks @nibou230 !

@nibou230 : If you have a chance, could you add this new UI configuration to our documentation at https://wiki.lyrasis.org/display/DSDOC7x/User+Interface+Configuration?  (let me know if you need access)  If you don't have time, I can add something in the coming week.

PR #1617, opened 2022-04-27
Researcher profile

LucaGiamminonni:
## References
Requires https://github.com/DSpace/DSpace/pull/8232
## Description

With this changes we add:
- a new section in the profile page to create, view, change visibility and delete researcher profile
- a new button on the item's page to claim that item (if claim for that entity type is supported)

image
image
image
image

## Checklist


tdonohue:
@LucaGiamminonni : You may have already noticed this, but this PR has lint errors, so it's failing to run tests.

src/app/core/profile/researcher-profile.service.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+}
+
+/**
+ * A service that provides methods to make REST requests with researcher profile endpoint.
+ */
+@Injectable()
+@dataService(RESEARCHER_PROFILE)
+export class ResearcherProfileService {
```
tdonohue, line 61: I might be missing it, but I think this entire file is missing specs.  I don't see a corresponding `researcher-profile.service.spec.ts

src/app/core/breadcrumbs/dso-name.service.ts
```diff
@@ -27,7 +27,13 @@ export class DSONameService {
...
   private readonly factories = {
     Person: (dso: DSpaceObject): string => {
-      return `${dso.firstMetadataValue('person.familyName')}, ${dso.firstMetadataValue('person.givenName')}`;
+      const familyName = dso.firstMetadataValue('person.familyName');
+      const givenName = dso.firstMetadataValue('person.givenName');
+      if (isEmpty(familyName) && isEmpty(givenName)) {
+        return dso.firstMetadataValue('dc.title') || dso.name;
+      } else {
+        return `${familyName}, ${givenName}`;
+      }
```
tdonohue, line 36: Can we add a test for the new logic to `dso-name.service.spec.ts`?  There's already a test for the old logic, so it just needs to be updated to also test that when one of these name fields is empty, the dc.title is used.

src/app/entity-groups/research-entities/item-pages/person/person.component.ts
```diff
@@ -12,5 +25,81 @@ import { listableObjectComponent } from '../../../../shared/object-collection/sh
...
+
+    this.authorizationService.isAuthorized(FeatureID.ShowClaimItem, this.object._links.self.href).pipe(
+      take(1)
+    ).subscribe((isAuthorized: boolean) => {
+      this.claimable$.next(isAuthorized);
+    });
+
+  }
+
+  claim() {
```
tdonohue, line 51: This method needs TypeDocs & also needs specs added to `person.component.spec.ts`

src/app/entity-groups/research-entities/item-pages/person/person.component.ts
```diff
@@ -12,5 +25,81 @@ import { listableObjectComponent } from '../../../../shared/object-collection/sh
...
+      if (!isAuthorized) {
+        this.notificationsService.warning(this.translate.get('researcherprofile.claim.not-authorized'));
+      } else {
+        this.createFromExternalSource();
+      }
+    });
+
+  }
+
+  createFromExternalSource() {
```
tdonohue, line 65: Again, this method needs TypeDocs to describe it. And it needs Specs

src/app/entity-groups/research-entities/item-pages/person/person.component.ts
```diff
@@ -12,5 +25,81 @@ import { listableObjectComponent } from '../../../../shared/object-collection/sh
...
+          this.claimable$.next(false);
+        } else {
+          this.notificationsService.error(
+            this.translate.get('researcherprofile.error.claim.title'),
+            this.translate.get('researcherprofile.error.claim.body'));
+        }
+      });
+  }
+
+  isClaimable(): Observable<boolean> {
```
tdonohue, line 84: Same here. This method needs Typedocs & Specs

src/app/entity-groups/research-entities/item-pages/person/person.component.ts
```diff
@@ -12,5 +25,81 @@ import { listableObjectComponent } from '../../../../shared/object-collection/sh
...
+    return this.claimable$;
+  }
+
+  getTitleMetadataValues(): MetadataValue[]{
```
tdonohue, line 88: One last time. This method needs TypeDocs and Specs

src/app/profile-page/profile-claim/profile-claim.service.ts
```diff
@@ -0,0 +1,70 @@
...
+import { EPerson } from '../../core/eperson/models/eperson.model';
+import { DSpaceObject } from '../../core/shared/dspace-object.model';
+import { SearchService } from '../../core/shared/search/search.service';
+import { hasValue } from '../../shared/empty.util';
+import { PaginatedSearchOptions } from '../../shared/search/models/paginated-search-options.model';
+import { SearchResult } from '../../shared/search/models/search-result.model';
+import { getFirstSucceededRemoteData } from './../../core/shared/operators';
+
+@Injectable()
+export class ProfileClaimService {
```
tdonohue, line 16: Please add TypeDocs to describe this new service.  All methods in this class need TypeDocs as well.  Finally, it looks like this service is missing specs as I don't see a corresponding `profile-claim.service.spec.ts` file.

src/app/profile-page/profile-page-researcher-form/profile-page-researcher-form.component.ts
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * Return a boolean representing if a create operation is pending.
+   *
+   * @return {Observable<boolean>}
+   */
+  isProcessingCreate(): Observable<boolean> {
+    return this.processingCreate$.asObservable();
+  }
+
+  createProfileFromScratch() {
```
tdonohue, line 155: Please add TypeDocs or comments to describe this method

src/app/profile-page/profile-page-researcher-form/profile-page-researcher-form.component.ts
```diff
@@ -0,0 +1,181 @@
...
+      if (remoteData.isSuccess) {
+        this.initResearchProfile();
+        this.notificationService.success(this.translationService.get('researcher.profile.create.success'));
+      } else {
+        this.notificationService.error(this.translationService.get('researcher.profile.create.fail'));
+      }
+    });
+  }
+
+  private initResearchProfile(): void {
```
tdonohue, line 170: Please add TypeDocs/comments to describe this method

src/app/profile-page/profile-page.component.ts
```diff
@@ -163,4 +171,9 @@ export class ProfilePageComponent implements OnInit {
   submit() {
     this.updateProfile();
   }
+
+  isResearcherProfileEnabled(){
```
tdonohue, line 175: This new method has no Specs.  We should add basic specs to prove it's returning true/false when expected.

src/app/shared/dso-selector/modal-wrappers/claim-item-selector/claim-item-selector.component.ts
```diff
@@ -0,0 +1,68 @@
...
+import { ProfileClaimService } from '../../../../profile-page/profile-claim/profile-claim.service';
+import { CollectionElementLinkType } from '../../../object-collection/collection-element-link.type';
+
+
+
+@Component({
+  selector: 'ds-claim-item-selector',
+  templateUrl: './claim-item-selector.component.html'
+})
+export class ClaimItemSelectorComponent extends DSOSelectorModalWrapperComponent implements OnInit {
```
tdonohue, line 23: This new component needs TypeDocs to describe it.  All methods also need TypeDocs.

src/app/shared/dso-selector/modal-wrappers/claim-item-selector/claim-item-selector.component.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(ClaimItemSelectorComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
tdonohue, line 43: It looks like this specs file is missing tests for the methods in `claim-item-selector.component.ts`.

tdonohue (changes requested):
@LucaGiamminonni : Gave this a test today. The code works.  That said, I have inline comments below as there's some missing Specs & missing TypeDocs in this PR.

tdonohue:
@LucaGiamminonni : Thanks for the updates! I'll give this another review/test today.  In the meantime though, please note that [LGTM has added 13 alerts to this PR](https://github.com/DSpace/dspace-angular/pull/1617#issuecomment-1118741527).  They are all tiny code fixes (mostly unnecessary imports), but those need to get cleaned up as well please.

tdonohue (changes requested):
@LucaGiamminonni : Gave this another review/test today.  While some of my prior feedback was addressed, other parts were overlooked.  I also encountered a few minor bugs we may want to look into resolving.  Here's what I found:

1. LGTM notes that there are 13 alerts on this PR. Those should be fixed. They are all minor fixes.
2. Missing tests / specs.  See my previous review.  Tests/Specs are missing for the following
    * Entire `researcher-profile.service.ts` file
    * New logic in `dso-name.service.ts`
    * All the new methods in `person.component.ts`
    * The entire `profile-claim.service.ts` file  
    * New methods in `profile-page.component.ts`
    * New methods in `claim-item-selector.component.ts`.  This component is also missing TypeDocs
3. (Minor usability issue)  When a Researcher Profile is already linked to an EPerson, we can hide/remove the "Create New" button? Currently it's disabled, but it should probably just be removed from the screen (i.e. only "View" and "Delete" should be displayed) 
4. (Minor bug - possible caching issue?).  In some scenarios, when a matching Person Entity already exists for the logged in user, it's displayed with the name "undefined, undefined".  Here's how I've been able to reproduce this in production mode (`yarn build:prod; yarn serve:ssr`):
    * First, login as an EPerson with no researcher profile. Create a new one.
    * Now, go back and Delete the researcher profile (soft delete, just unlink it)
    * Finally, click "Create New" again.  The UI will show the existing Researcher Profile in a popup window, but it's listed as "undefined, undefined" profile-select


I've verified that the basic functionality all works well.  I can create / claim / delete (soft or hard) researcher profiles.

atarix83:
@tdonohue 

Since I wasn't involved in the implementation for this task, I've added my personal code review by implementing them directly. Here the list of what I've changed: 
- A standalone button component for profile item claim has been created
- Modal to show profile suggestions has been moved within the profile page module, since it wasn't a dso selector as the existing ones
- Fixed an issue with the change visibility toggle

I've also addressed your feedback and added missing unit tests an typedocs

src/app/core/breadcrumbs/dso-name.service.ts
atarix83: done

src/app/core/profile/researcher-profile.service.ts
atarix83: done

src/app/profile-page/profile-claim/profile-claim.service.ts
atarix83: done

src/app/profile-page/profile-page-researcher-form/profile-page-researcher-form.component.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * Initializes the researcherProfile and researcherProfileItemId attributes using the profile of the current user.
+   */
+  private initResearchProfile(): void {
+    this.researcherProfileService.findById(this.user.id, false).pipe(
+      getFirstSucceededRemoteDataPayload(),
+      tap((researcherProfile) => this.researcherProfile$.next(researcherProfile)),
+      mergeMap((researcherProfile) => this.researcherProfileService.findRelatedItemId(researcherProfile)),
```
artlowel, line 184: perhaps add `followLink('item')` to the `findById` call above? Then you wouldn't have to do this mergeMap, and the rest api can embed the item in the original call.

src/app/shared/dso-page/person-page-claim-button/person-page-claim-button.component.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Create a new researcher profile claiming the current item.
+   */
+  claim() {
+    this.researcherProfileService.createFromExternalSource(this.object._links.self.href).pipe(
+      getFirstCompletedRemoteData(),
+      mergeMap((rd: RemoteData<ResearcherProfile>) => {
+        console.log(rd);
```
artlowel, line 58: Left a console.log

src/app/shared/dso-page/person-page-claim-button/person-page-claim-button.component.ts
```diff
@@ -0,0 +1,85 @@
...
+        if (rd.hasSucceeded) {
+          return this.researcherProfileService.findRelatedItemId(rd.payload);
```
artlowel, line 60: Same  `followLink` comment. If you add a `followLink(`item`)` to the create call, the item will be embedded in the response and you don't have to fetch it separately

artlowel (changes requested):
Thanks @LucaGiamminonni!

I added some in-line comments

package.json
```diff
@@ -125,7 +125,8 @@
     "url-parse": "^1.5.6",
     "uuid": "^8.3.2",
     "webfontloader": "1.6.28",
-    "zone.js": "~0.11.5"
+    "zone.js": "~0.11.5",
+    "ngx-ui-switch": "^11.0.1"
```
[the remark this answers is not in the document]
artlowel: After testing the UI I see that you used the toggle in a way that isn't supported by the default bootstrap version (with the text inside). So I retract this comment.

artlowel (commented):
I finally got a chance to test the UI as well, and I noticed a few issues:

- It's a bit confusing that changing the state of the profile is saved immediately, while the rest of the profile page requires you to click the save button at the bottom first. I'd change it to only save when you click save.
- I'd expect the delete button to ask for confirmation first. As it stands it deletes the profile-entity immediately.
- I haven't found a way to to trigger this popup. What are the exact requirements to do so?
  screenshot
- If you create a profile, then go to the edit item page for that profile and open the Authorizations section, you'll get an error because the UI seems to be trying to do a POST to the `/api/eperson/profiles` endpoint
  - The same thing happens when you go to any of the admin "Access Control" pages
  - These errors seem to compound. The first time there's one call (and one error), the second time two, the third time three, and so on. This could point to a subscription that is created every time but never unsubscribed

tdonohue:
@LucaGiamminonni or @atarix83 : When you get a chance, I'd appreciate it if this PR could be updated/fixed. I'd like to re-test this with your backend changes (the backend https://github.com/DSpace/DSpace/pull/8232 looks good).  As noted in today's meeting, I'd like to see if we can get the Researcher Profile PRs merged as soon as possible, in order to make it easier to review/update the other ORCID-related PRs.  Thanks!

src/app/entity-groups/research-entities/item-pages/person/person.component.ts
```diff
@@ -13,4 +16,24 @@ import { listableObjectComponent } from '../../../../shared/object-collection/sh
...
+      metadataValues.push(familyName);
+    }
+    if (givenName) {
+      metadataValues.push(givenName);
+    }
+    if (metadataValues.length === 0 && title) {
+      metadataValues.push(title);
+    }
+    return metadataValues;
+  }
```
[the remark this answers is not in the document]
atarix83: here is no applicable because we need the metadataValue object, while `dsoNameService.getName` return only a string

src/app/profile-page/profile-page-researcher-form/profile-page-researcher-form.component.ts
```diff
@@ -0,0 +1,190 @@
...
+    this.processingDelete$.next(true);
+    this.researcherProfileService.delete(researcherProfile)
+      .subscribe((deleted) => {
+        if (deleted) {
+          this.researcherProfile$.next(null);
+          this.researcherProfileItemId = null;
+        }
+        this.processingDelete$.next(false);
+      });
+  }
```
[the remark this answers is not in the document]
atarix83: in this case returning the remote data is not applicable. The creation is triggered on a button click

src/app/core/profile/researcher-profile.service.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  /**
+   * Find the item id related to the given researcher profile.
+   *
+   * @param researcherProfile the profile to find for
+   */
+  public findRelatedItemId(researcherProfile: ResearcherProfile): Observable<string> {
+    return this.itemService.findByHref(researcherProfile._links.item.href, false)
+      .pipe(getFirstSucceededRemoteDataPayload(),
```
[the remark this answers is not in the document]
atarix83: done

src/app/core/profile/researcher-profile.service.ts
```diff
@@ -0,0 +1,184 @@
...
+    const requestId = this.requestService.generateRequestId();
+    const href$ = this.halService.getEndpoint(this.dataService.getLinkPath());
+
+    href$.pipe(
+      find((href: string) => hasValue(href)),
+      map((href: string) => {
+        const request = new PostRequest(requestId, href, sourceUri, options);
+        this.requestService.send(request);
+      })
+    ).subscribe();
```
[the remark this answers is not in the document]
atarix83: done

src/app/entity-groups/research-entities/item-list-elements/search-result-list-elements/person/person-search-result-list-element.component.ts
```diff
@@ -14,9 +17,15 @@ import { ItemSearchResultListElementComponent } from '../../../../../shared/obje
...
+  }
+
   get name() {
-    return this.value ?
-      this.value :
-      this.firstMetadataValue('person.familyName') + ', ' + this.firstMetadataValue('person.givenName');
+    let personName = this.dsoNameService.getName(this.dso);
+    if (isNotEmpty(this.firstMetadataValue('person.familyName')) && isNotEmpty(this.firstMetadataValue('person.givenName'))) {
+      personName = this.firstMetadataValue('person.familyName') + ', ' + this.firstMetadataValue('person.givenName');
+    }
```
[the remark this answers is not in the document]
atarix83: done

src/app/profile-page/profile-claim/profile-claim.service.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (isEmpty(query)) {
+      return of(null);
+    }
+    return this.searchService.search(new PaginatedSearchOptions({
+      configuration: 'eperson_claims',
+      query: query
+    })).pipe(
+      getFirstCompletedRemoteData()
+    );
+  }
```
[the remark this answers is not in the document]
atarix83: done

src/app/core/profile/researcher-profile.service.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    return this.rdbService.buildFromRequestUUID(requestId);
+  }
+
+  private patch(researcherProfile: ResearcherProfile, operations: Operation[]): Observable<RemoteData<ResearcherProfile>> {
+    return this.dataService.patch(researcherProfile, operations).pipe(
+      getFirstCompletedRemoteData()
+    );
```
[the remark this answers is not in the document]
atarix83: done

src/app/profile-page/profile-page-researcher-form/profile-page-researcher-form.component.ts
atarix83: added followLink and changed the findRelatedItemId in order to use embedded item if available

src/app/shared/dso-page/person-page-claim-button/person-page-claim-button.component.ts
atarix83: done

src/app/shared/dso-page/person-page-claim-button/person-page-claim-button.component.ts
atarix83: added followLink and changed the findRelatedItemId in order to use embedded item if available

atarix83:
@artlowel 

all your feedback should be addressed. In particular

> It's a bit confusing that changing the state of the profile is saved immediately, while the rest of the profile page requires you to click the save button at the bottom first. I'd change it to only save when you click save.

The toggle was added to make the status more visible and easier to edit without using the form. In our opinion it's an added value and It would be a pity if we remove


> I'd expect the delete button to ask for confirmation first. As it stands it deletes the profile-entity immediately.

done


> I haven't found a way to to trigger this popup. What are the exact requirements to do so?

it only appears when a possible matching profile is found otherwise it's not shown

> If you create a profile, then go to the edit item page for that profile and open the Authorizations section, you'll get an error because the UI seems to be trying to do a POST to the /api/eperson/profiles endpoint
    The same thing happens when you go to any of the admin "Access Control" pages
    These errors seem to compound. The first time there's one call (and one error), the second time two, the third time three, and so on. This could point to a subscription that is created every time but never unsubscribed

this should be fixed now

src/themes/custom/styles/theme.scss
```diff
@@ -11,3 +11,4 @@
 @import '../../../styles/bootstrap_variables_mapping.scss';
 @import '../../../styles/_truncatable-part.component.scss';
 @import './_global-styles.scss';
+@import '../../../../node_modules/ngx-ui-switch/ui-switch.component.scss';
```
tdonohue, line 14: This is not a comment specific to this PR, but more about our process for these SCSS imports. 

It seems **really odd** to me that individual themes don't import the `src/styles/base-theme.scss` by default... as that means anytime we update the `base-theme.scss` we have to **require** that all custom themes also update their `theme.scss`.

Shouldn't we consider changing this @atarix83 and @artlowel ?  Even adding this single import now requires anyone upgrading to 7.3 to ensure they update their `theme.scss`.  It seems like we could avoid that by just having:
```
@import '../../../styles/base-them.scss'
```
(It'd also let us remove a lot of these other imports by default)

tdonohue (changes requested):
@atarix83 : I tried testing this today, but unfortunately, I cannot seem to create a new Researcher Profile (when one doesn't yet exist).  No matter what account I use, it always results in a `400 Bad Request` error.  In the backend logs, the error seems to correspond to a WARN message:
```
 WARN  a14cfb2c-4e21-445c-bdb9-b1f53afe8843 8bcdc2c4-e777-4879-b83a-6f1af3d92dc2 org.dspace.app.rest.exception.DSpaceApiExceptionControllerAdvice @ Request is invalid or incorrect (status:400 exception: A profile item must have a valid dspace.object.owner metadata at: org.dspace.app.profile.ResearcherProfile.lambda$getDspaceObjectOwnerMetadata$3(ResearcherProfile.java:70))
```
In the UI, this looks like:
profile-create-error

To me, it appears that the backend is unable to set the required `dspace.object.owner`.  Not yet sure if this is because of a backend error in @LucaGiamminonni 's PR, or if it's because the frontend is sending an unexpected request.

In any case, this blocks me from testing further as I cannot get an initial profile created to test with.

atarix83:
@tdonohue 

> To me, it appears that the backend is unable to set the required dspace.object.owner. Not yet sure if this is because of a backend error in @LucaGiamminonni 's PR, or if it's because the frontend is sending an unexpected request.
In any case, this blocks me from testing further as I cannot get an initial profile created to test with.
NOTE: I can verify that all my other inline code feedback has been addressed. Thanks!

this is because now by default the authority configured for dspace.object.owner is disabled by default https://github.com/4Science/DSpace/blob/CST-5306/dspace/config/modules/authority.cfg#L84-L88

To test it please enable it again

>This is not a comment specific to this PR, but more about our process for these SCSS imports.
It seems really odd to me that individual themes don't import the src/styles/base-theme.scss by default... as that means anytime we update the base-theme.scss we have to require that all custom themes also update their theme.scss.
Shouldn't we consider changing this @atarix83 and @artlowel ? Even adding this single import now requires anyone upgrading to 7.3 to ensure they update their theme.scss. It seems like we could avoid that by just having:

I agree we try to achieve it within this PR

src/themes/custom/styles/theme.scss
artlowel: You can't just import `base-theme.scss`. The reason these files are separate is because the imports have to happen in a  particular order for the themes, which differs from `base-theme.scss`. The order makes it possible for us to override variables. What we _can_ do in order to make this less of a problem is create an scss file in `src/styles` for 3rd party imports, and import that file in `base-theme.scss`, and both theme.scss files. We can add [that noui slider import](https://github.com/DSpace/dspace-angular/blob/3a9783d031bd6940cdb9bcc172fb57cc292e3ffb/src/styles/base-theme.scss#L3) to it as well. That way theme css files will no longer have to change if a new 3rd party css import gets added

artlowel:
>>It's a bit confusing that changing the state of the profile is saved immediately, while the rest of the profile page requires you to click the save button at the bottom first. I'd change it to only save when you click save.
>
>The toggle was added to make the status more visible and easier to edit without using the form. In our opinion it's an added value and It would be a pity if we remove

I'm not sure I'm entirely convinced, but that shouldn't block this PR.

>>I haven't found a way to to trigger this popup. What are the exact requirements to do so?
>
>it only appears when a possible matching profile is found otherwise it's not shown

What are the specific steps to create "a possible matching profile". In the last meeting @atarix83 mentioned something about an email address, but I don't see one used in the Person metadata. The `dspace.object.owner` field seems to contain the name. If I create a new profile and remove that field, the link between the EPerson and the Person entity is broken, but creating another profile doesn't show this one as a duplicate.

artlowel:
@tdonohue explained in the meeting that a `person.email` field needs to be added to the entity in order for it to show up as a duplicate, but even so I'm not able to trigger it. The procedure I've tried is:
- go to the profile page and create a new profile
- edit the resulting person entity
- add a `person.email` field with the exact same email address used by the eperson
- unlink the profile from the eperson. I tried both:
   - clicking the delete button on the profile page
   - removing the `dspace.object.owner` field from the person entity
- log out, log back in
- go to the profile page again, and create a new profile

That last step always results in a new separate person entity, without showing the duplicate popup

atarix83:
@tdonohue  @artlowel 

I've investigated on the problem reported by Tim https://github.com/DSpace/dspace-angular/pull/1617#issuecomment-1147571106

I've fixed a problem with the profile lookup in order to not take the cached response and to make every time the request. In this way we have always the updated search result.

Unfortunately after resolved the lookup problem, the issue when creating the profile immediately after deleting it persists.
This is due to a cache problem on create method. Basically the second time the profile is created we receive a remote data with status error-stale.
It happens both when the profile is created from scratch every time (when `researcher-profile.hard-delete.enabled = true`) https://github.com/4Science/dspace-angular/blob/a830c99f908cfa0b451e1987fb8644d6b85c8d53/src/app/profile-page/profile-page-researcher-form/profile-page-researcher-form.component.ts#L183 and when an existing one it's claimed  (when `researcher-profile.hard-delete.enabled = false`)  https://github.com/4Science/dspace-angular/blob/a830c99f908cfa0b451e1987fb8644d6b85c8d53/src/app/shared/dso-page/person-page-claim-button/person-page-claim-button.component.ts#L58

Honestly i don't know how much time we could spend to find the reason of this bug. So I'd suggest to merge this PR and open a new issue and maybe the Atmire team can work on it considering that they have more knowledge on that part

artlowel:
I think what you're seeing may be the response to the first request that's returned immediately. Before the second request is even sent (as that happens in an ngrx effect). In that case we may need something like [this `skipWhile`](https://github.com/DSpace/dspace-angular/blob/2622f374b03680569953af323d241d0a678a0e23/src/app/core/data/data.service.ts#L396-L400) for create (and put and patch etc) as well, though because those methods should never return a cached object it should simply be `skipWhile((rd: RemoteData<PaginatedList<T>>) => rd.hasCompleted)`. Feel free to create an issue for it and assign it to me. We likely won't be able to get a PR ready in time for 7.3 though.

tdonohue (approved):
@atarix83 : Thanks for the updates & explanation.  I agree with @artlowel that, if you can create a ticket for the bug you've found, then I can assign to to @artlowel to work on in 7.4.

I've also re-tested this today with the latest changes, and I've found this PR is working much better. I'm again able to trigger the popup to recommend I claim an existing profile by doing the following:
1. First login as a user with no researcher profile (to avoid the create caching issue)
2. Create a Researcher Profile from `/profile`
3. Immediately delete it (with hard-delete disabled), this leaves the Person Entity but unlinks it.
4. Click "Create New" again (without logging out or reloading the page or anything)
5. I see the popup that shows the existing profile & asks if I'd like to claim it.
popup

I am able to sometimes reproduce that odd caching error... but it's more difficult to reproduce now than it was before (and I admit I'm having trouble finding a consistent way to reproduce it -- it doesn't appear for me in the "create, unlink & reclaim" scenario I described above).  So, I'm OK with leaving that issue for now and having @atarix83 create a bug ticket which we can resolve at a later time.

In summary, thanks for the hard work here @atarix83 and @LucaGiamminonni  !  I'm +1 this PR now (but please create that bug ticket for us).

(UPDATE: I did notice we never resolved the question about making [custom themes easier to maintain](https://github.com/DSpace/dspace-angular/pull/1617/files#r883938665), but I can move that into a separate ticket as well & we can do that in 7.4 as needed.)

src/themes/custom/styles/theme.scss
tdonohue: @artlowel and @atarix83 : I've moved this over into a separate ticket. https://github.com/DSpace/dspace-angular/issues/1679

I still think it'd be very nice to do at some point soon. If either of you want to do it in 7.3 (as I think it's a very small but beneficial change) let me know. Otherwise, we'll move this to the list of tasks for 7.4

atarix83:
@tdonohue @artlowel 

here the issue for the create cache problem https://github.com/DSpace/dspace-angular/issues/1680

artlowel:
I'm sorry, but I still can't trigger the popup. If I follow the procedure Tim describes above, I get the error every time. Temporarily adding a filter for stale responses in the dataservice create method bypasses that error, but it still doesn't give me the popup. It just creates a new profile immediately.

I can verify that new profiles now get the person.email field, so that's not the reason.

atarix83:
> I'm sorry, but I still can't trigger the popup. If I follow the procedure Tim describes above, I get the error every time. Temporarily adding a filter for stale responses in the dataservice create method bypasses that error, but it still doesn't give me the popup. It just creates a new profile immediately.
I can verify that new profiles now get the person.email field, so that's not the reason.

Are you running the latest rest branch?

artlowel:
Yes, commit https://github.com/DSpace/DSpace/pull/8232/commits/b801c0c69f8ecb811a5998001aa05d9d8cb7f29d

atarix83:
@artlowel 

the property `researcher-profile.hard-delete.enabled = false` is present in your local.cfg? if not can you try with that?

artlowel:
It is. If I delete the profile the entity persists

tdonohue:
@artlowel : Just a guess on what you might be seeing... I saw odd errors when I once forgot to enable the `EPersonAuthority` plugin.  Maybe that's what you are seeing?  Basically, there's 2 steps to enabling the Researcher Profiles in the backend.  First, you have to set `researcher-profile.entity-type = Person`, then you also have to enable the EPersonAuthority plugin, by uncommenting these in `dspace.cfg` or adding them to `local.cfg`:
```
choices.plugin.dspace.object.owner = EPersonAuthority
choices.presentation.dspace.object.owner = suggest
authority.controlled.dspace.object.owner = true
```
See the hints on "Enabling this feature" in the description of the backend PR: https://github.com/DSpace/DSpace/pull/8232

My guess is you might have forgotten to enable that EPersonAuthority?  

If that's not the solution, could you share more info on the error message you are seeing (either here or in Slack)?  There has to be something different about your setup, but I'm not sure what it could be, so I may need more info to help debug it.

artlowel:
No, I checked again, that EPersonAuthority plugin is configured.

The error I'm seeing is shown below. Two notifications, no useful info. There is also no corresponding error request in the network tab, so I assume it's just due to the errorStale remotedata @atarix83 mentioned.

image

If I refresh the page after getting that error, the profile has been created.

However if no-one can reproduce this issue but me, this doesn't necessarily have to block this PR. The rest looks good and this duplicate check isn't the crux of the feature.

atarix83:
@artlowel 

>No, I checked again, that EPersonAuthority plugin is configured.
The error I'm seeing is shown below. Two notifications, no useful info. There is also no corresponding error request in the network tab, so I assume it's just due to the errorStale remotedata @atarix83 mentioned.
image
If I refresh the page after getting that error, the profile has been created.
However if no-one can reproduce this issue but me, this doesn't necessarily have to block this PR. The rest looks good and this duplicate check isn't the crux of the feature.

It seems the problem due to the cache issue. Just to confirm, I've updated https://github.com/DSpace/dspace-angular/issues/1680 adding the videos of the error occurring in the two different scenarios. Could you confirm you're expericing them?

artlowel:
Yes it is exactly like the second video, however `researcher-profile.hard-delete.enabled` is set to `false` for me (first it was left out, because false is the default version, but now I've also tested setting it explicitly: no difference).

If I refresh between deleting the profile and creating a new one, I get no error, but also no duplicate popup

tdonohue:
As discussed in today's meeting, we are going to merge this "as-is".  @atarix83 , @artlowel and I all feel that this looks good.  @artlowel is still having issues reproducing the duplicate profile popup (when a profile already exists), but is going to investigate if the issue might be with Discovery configuration.  In any case, we all agree that shouldn't block this PR from being merged, and instead, _if we discover a unaddressed bug_ that can be moved to the follow-up PR #8263

PR #1618, opened 2022-04-27
Orcid settings page

LucaGiamminonni:
## References
Includes https://github.com/DSpace/dspace-angular/pull/1617
Requires https://github.com/DSpace/DSpace/pull/8263

## Description
This PR add a specific page to:
- handle orcid link/unlink with profile item
- configure orcid synchronization

This new page is accesible from the person item page with a specific button placed on the top right of the page.

## Checklist


tdonohue:
@LucaGiamminonni : You may have already noticed this, but this PR has lint errors, so it's failing to run tests.

src/app/core/profile/researcher-profile.service.ts
```diff
@@ -0,0 +1,266 @@
...
+  /**
+   * Find the researcher profile with the given uuid.
+   *
+   * @param uuid the profile uuid
+   */
+  findById(uuid: string): Observable<ResearcherProfile> {
+    return this.dataService.findById(uuid, false)
+      .pipe ( getFinishedRemoteData(),
+        map((remoteData) => remoteData.payload));
+  }
```
artlowel, line 94: Please don't "override" methods on a dataservice by using delegation and giving them a different signature or return type. `findById` on a dataservice is expected to return an `Observable<RemoteData<T>>` and have useCacheIfAvailable etc in its signature

Doing this will be made impossible after we fix https://github.com/DSpace/dspace-angular/issues/742, but for now we have to manually enforce it

src/app/core/profile/researcher-profile.service.ts
```diff
@@ -0,0 +1,266 @@
...
+  create(): Observable<RemoteData<ResearcherProfile>> {
+    return this.dataService.create( new ResearcherProfile());
+  }
+
+  /**
+   * Delete a researcher profile.
+   *
+   * @param researcherProfile the profile to delete
+   */
+  delete(researcherProfile: ResearcherProfile): Observable<boolean> {
```
artlowel, line 108: Same comment as findByID above, though if I'm not mistaken this change stems from #1617 and I already commented on it there

src/app/core/profile/researcher-profile.service.ts
```diff
@@ -0,0 +1,266 @@
...
+      this.configurationService.findByPropertyName('orcid.application-client-id').pipe(getFirstSucceededRemoteDataPayload()),
+      this.configurationService.findByPropertyName('orcid.scope').pipe(getFirstSucceededRemoteDataPayload())]
+    ).pipe(
+      map(([authorizeUrl, clientId, scopes]) => {
+        const redirectUri = environment.rest.baseUrl + '/api/eperson/orcid/' + profile.id + '/?url=' + encodeURIComponent(this.router.url);
+        return authorizeUrl.values[0] + '?client_id=' + clientId.values[0]   + '&redirect_uri=' + redirectUri + '&response_type=code&scope='
+          + scopes.values.join(' ');
+      }));
+  }
+
```
artlowel, line 235: I'd add a filter abvove the map to see if all three of the variables in the array have a value. Combine latest will emit any time any of the source observables change, there is no guarantee that will happen at the same time for all three, so I expect there will be cases in which the code in the map fails because one of those values is not defined.

src/app/item-page/orcid-page/orcid-auth/orcid-auth.component.html
```diff
@@ -0,0 +1,71 @@
...
+                {{'person.page.orcid.missing-authorizations-message' | translate}}
+                <ul>
+                    <li *ngFor="let auth of ( ( missingAuthorizations$ | async ) )"> {{getAuthorizationDescription(auth) | translate }} </li>
+                </ul>
+                </div>
+            </div>
+            </div>
+        </div>
+        </div>
+        <div *ngIf="(onlyAdminCanDisconnectProfileFromOrcid() | async)" class="row mt-3 p-3 mb-2 bg-info text-white rounded d-flex justify-content-center">
```
artlowel, line 41: It is bad for performance to call methods from a template. They will be called multiple times a second, any time Angular checks anything about the template basically. 

If the methods are trivial such as `getOrcidAuthorizations()` which only checks a metadata value it's not that big a deal (though you could still improve performance by storing their result in a component property, especially if you use them more than once). 

However `onlyAdminCanDisconnectProfileFromOrcid()` and `ownerCanDisconnectProfileFromOrcid()` ultimately depend on a backend call, so these should really be removed from the template

src/app/item-page/orcid-page/orcid-sync/orcid-setting.component.ts
```diff
@@ -0,0 +1,139 @@
...
+  constructor(private researcherProfileService: ResearcherProfileService,
+    protected translateService: TranslateService,
+    private notificationsService: NotificationsService,
+    public authService: AuthService,
+    private route: ActivatedRoute,
+    private itemService: ItemDataService
+  ) {
+     this.itemService.findById(this.route.snapshot.paramMap.get('id'), true, true).pipe(getFirstCompletedRemoteData()).subscribe((data: RemoteData<Item>) => {
+      this.item = data.payload;
+     });
```
artlowel, line 49: This component is part of the ItemPageRouting module. The item is already part of the route data. You can get the item from the route [like this](https://github.com/4Science/dspace-angular/blob/804930fbe2b5cca46343d24bf7486034cf5cc4f8/src/app/item-page/simple/item-page.component.ts#L80-L83).

src/app/shared/log-in/methods/orcid/log-in-orcid.component.ts
```diff
@@ -0,0 +1,110 @@
...
+        // change the redirect url with the current page url
+        const newRedirectUrl = `?redirectUrl=${correctRedirectUrl}`;
+        orcidServerUrl = this.location.replace(/\?redirectUrl=(.*)/g, newRedirectUrl);
+      }
+
+      // redirect to orcid authentication url
+      this.hardRedirectService.redirect(orcidServerUrl);
+    });
+
+  }
```
artlowel, line 108: This is an almost exact copy of [`redirectToShibolleth()`](https://github.com/4Science/dspace-angular/blob/716cea376da914d56f1f96a7aab3d7bc69a63f07/src/app/shared/log-in/methods/shibboleth/log-in-shibboleth.component.ts#L87-L113) in the `LogInShibbolethComponent`, only a few variable names appear to have changed. Perhaps they can be refactored to use the same method?

artlowel (changes requested):
Thanks @LucaGiamminonni!

benbosman (changes requested):
While testing https://github.com/DSpace/DSpace/pull/8263, I noticed:
* If the user didn't make their email address public in ORCID, there's no page displaying what the user should do (rather a very generic error page). I've reported this in https://github.com/DSpace/DSpace/pull/8263 as well since a REST part has to be implemented for this as well
* When on the `/entities/person/uuid/orcid` page, there's no link back to the person page

tdonohue:
@LucaGiamminonni : Just a quick note that this PR seems to be in an odd state. It didn't run any tests/verifications the last time you pushed changes.  As I mentioned in last week's meeting (May 26), it'd be good if you could simply fix a [few of the small LGTM issues](https://github.com/DSpace/dspace-angular/pull/1618#issuecomment-1133047968) -- as pushing up a small code fix might be enough to get this PR to correctly run the tests/verifications.

tdonohue:
@LucaGiamminonni & @atarix83  : Just a quick note, it appears to me that @artlowel 's prior review / feedback has been forgotten in this PR: https://github.com/DSpace/dspace-angular/pull/1618#pullrequestreview-970666301  It looks like that feedback was sent almost a month ago and still hasn't been responded to or addressed. Could this PR also be updated based on that feedback?

src/app/core/profile/researcher-profile.service.ts
atarix83: it's not needed combineLatest will not emit an initial value until each observable emits at least one value

tdonohue (changes requested):
@LucaGiamminonni and @atarix83 : Tested this today and overall it's working well with the backend ORCID Auth PR.  I just have minor comments about the ORCID settings page (as I find it confusing):
orcid-settings

1. The circled sections should have some sort of description in the boxes.  For instance, I'm assuming if you enable the "Publication Preferences" this will sync ORCID publication entries into DSpace Publication Entities? (is that correct?)   I'm unclear what "Funding Preferences" does as there is no Funding object in DSpace.  I also don't know what "Identifiers" means under Profile Preferences.
2. From the description, it's unclear what "Manual" synchronization is.  The description says "You must sync your data to ORCID manually"... but how is that done?  Is there some button you need to press that will appear (perhaps after the ORCID queue work is added in?)   It's possible this will become clearer later on, but right now I don't understand how to manually sync anything.
3. Under "Granted Authorizations" (or somewhere else on this page), I feel we must clearly state that we must store an ORCID access token within DSpace in order to provide syncing capabilities.  (I'm not sure if we need that access token for manual syncing, or if it's just for batch syncing, but we should clearer state how it's used.)
4. If I update my settings to "Batch" mode, the description doesn't change...so it's unclear what "Batch" mode means.  I'm assuming this is the automatic sync that occurs on a scheduled basis?  In any case it needs a description.

Beyond that, my initial tests were all successful.  It seems to work well to authenticate and allow me to link/unlink my ORCID with my Researcher Page.  I've not found any bugs yet, just the above minor usability issues.

atarix83:
@tdonohue 

> The circled sections should have some sort of description in the boxes. For instance, I'm assuming if you enable the "Publication Preferences" this will sync ORCID publication entries into DSpace Publication Entities? (is that correct?) I'm unclear what "Funding Preferences" does as there is no Funding object in DSpace. I also don't know what "Identifiers" means under Profile Preferences.

Please mind that the synchronization is from DSpace to ORCID registry.
Publication preferences it's used to set synchronization preference regarding the person publications present in DSPace. The same is for the Funding, but since in dspace we don't have the the `Funding` entity there are mapped to `Project` entity. We have maintained the name used by orcid. Regarding the identifiers if synchronized, it saves all Person profile identifiers to orcid registry.

> From the description, it's unclear what "Manual" synchronization is. The description says "You must sync your data to ORCID manually"... but how is that done? Is there some button you need to press that will appear (perhaps after the ORCID queue work is added in?) It's possible this will become clearer later on, but right now I don't understand how to manually sync anything.

This part will be clearer with the queue PR. Basically "Manual" synchronization will be triggered only by the user in the queue page, while "Batch" is done by a batch script that is running once a day

> Under "Granted Authorizations" (or somewhere else on this page), I feel we must clearly state that we must store an ORCID access token within DSpace in order to provide syncing capabilities. (I'm not sure if we need that access token for manual syncing, or if it's just for batch syncing, but we should clearer state how it's used.)

syncing capabilities are automatically provided once a user connects dsapce person profile with its own orcid profile, it' doesn't matter if synchronization is triggerd manually or done by batch script

> If I update my settings to "Batch" mode, the description doesn't change...so it's unclear what "Batch" mode means. I'm assuming this is the automatic sync that occurs on a scheduled basis? In any case it needs a description.

I've updated the description

@artlowel 

you feedback shoud be addresse, please check them again

tdonohue (changes requested):
@atarix83 : I see you just updated the text of the Synchronization note. As I previously noted, I'd recommend we add descriptions for each of these settings... if I was confused by them, then I'm certain others will be.  Here's my suggestions for the text to add in each box:

* Synchronization Mode box - I'd recommend minor updates to the next text to say this:
> Please select how you would like synchronization to ORCID to occur. The options include "Manual" (you must send your data to ORCID manually), or "Batch" (the system will send your data to ORCID via a scheduled script).
* Publication preferences box - I'd recommend we add descriptive text above the two options. Something like this:
> Select whether to send your linked Publication entities to your ORCID record's list of works
* Funding preferences box - Again, we should add descriptive text like this:
> Select whether to send your linked Project entities to your ORCID record's list of funding information.
* Profile Preferences box - Again, we should add descriptive text like this:
> Select whether to send your biographical data or personal identifiers to your ORCID record.

Overall, as I said, this page seems fine (however I obviously cannot test Sync capabilities until the other sync functionality is added in later PRs).  But, I do think the usability would be improved by adding simple descriptions to each of these boxes.

atarix83:
@tdonohue 

done this is the result
Schermata da 2022-06-14 09-47-17

artlowel (changes requested):
Thanks @LucaGiamminonni! It all works as it should. 

I do have some feedback on style and usability:

### ORCID button on person page
Screen Shot 2022-06-14 at 11 17 25

All other context sensitive "admin" style buttons we have here are icons. For consistency it would be better to use ORCID's "ID" logo here instead. Or even better, since we're getting multiple buttons here anyway it would be good to turn this into a dropdown menu on all dso pages, rather than putting a bunch of buttons next to each other. Though that may be out of the scope of this PR.

### Reaching the ORCID settings page
I think it's not ideal that that button is the only way to reach the settings page. I think it would make sense to add ORCID as a tab to the edit item page for Person items. In that case I'd also change the route to /entities/person/${uuid}/**edit**/orcid

Or, if we keep it as a separate page, perhaps add a direct link to the auth nav dropdown in the header, and/or the EPerson profile page


### Collapsible cards
As for the settings page itself, I'm not sure what purpose the collapsible outer cards serve. Why would you ever collapse them? Getting rid of the outer cards, and using simple headers should suffice to keep the sections separate, and it would get rid of that boxes-in-boxes look.

If they remain collapsible cards, you should be able to tell by looking at them: The header text shouldn't be a link, but styled like a card header, it should be left aligned, there should be a chevron on the right indicating the state etc.

### Spacing for authorizations boxes
On XS screens the authorizations boxes are glued together, they should get some margin-bottom

Screen Shot 2022-06-14 at 11 27 47

### Alert width

The "enable manual synchronization" alert is in a row, but not in a column. As a consequence it's wider than everything else:

image

atarix83:
@tdonohue @artlowel 

changes in order to fix issue discussed on the rest PR is done https://github.com/DSpace/DSpace/pull/8263#discussion_r896645135 Plase have a look

@artlowel 
> ORCID button on person page

updated using the orcid icon instead

> Reaching the ORCID settings page

We think this is out of scope of this PR. We agreed to address in 7.4 release

> Collapsible cards

outer boxes removed, this is the new layout
Schermata da 2022-06-14 17-59-40

> Spacing for authorizations boxes
Alert width

fixed

tdonohue (commented):
@LucaGiamminonni and @atarix83 :  Re-tested this today and it's working & looking good overall to me.  I also tested the new update by unlinking a Researcher profile, then re-claiming it and connecting it to ORCID (this results in the new PATCH request which works well.)

A few minor things I noticed in re-testing today.

1. When I update the ORCID Settings, the old ones are temporarily cached (i.e. the page doesn't refresh itself).  Here's what I'm doing
    * Login via ORCID, create a Researcher Profile. View that profile.
    * Go to the Orcid settings page (/orcid)
    * On that page change the Synchronization setting to "Batch" and enable one (or more) of the other preferences.
    * Click "Update settings".  You'll get a green success message.
    * Click "Back" button.
    * Now go immediately back to the ORCID settings page.  You'll see your newly  changed settings are missing. 
    * Click Reload in your browser.  Now, your newly changed settings are on that page.
    * NOTE: I've verified that the settings _are properly saved_.  However, the page appears to cache your old settings _until you click Reload in your browser_.  So, this is just a minor caching bug.
2. I do agree with @artlowel that we might think about (in 7.4 or later) moving the ORCID Settings page to a tab on the "Edit Item" page.  It is confusing to have the ORCID settings outside of the "Edit Item" page.  That said, I think we can schedule this for later cleanup work & leave this PR as-is.  We should just create a separate ticket for this action.

Beyond that, the updates all look good to me & they are working well.  So, I'm basically a +1, though it might be nice to cleanup that caching issue if possible. Thanks @atarix83 and @LucaGiamminonni !

atarix83:
@tdonohue  @artlowel 

cache issue is fixed, could you please check?

atarix83:
@tdonohue @artlowel 

build fails due to issue with e2e tests. it seems the rest server is not running properly. I've restarted the build several times but the error is the same. Also other PR builds are affected by the same problem

tdonohue (approved):
👍 Thanks @atarix83 and @LucaGiamminonni !  All my prior feedback (including the caching issue previously reported) has been addressed.  I retested today and everything is working well & I'm not finding any new bugs.  So, I'm OK with moving this forward as-is... if we find any additional bugs later, we can clean them up in a future PR (and/or the next ORCID PR).

tdonohue:
@artlowel : If you have a chance to re-test / re-review, this now looks good to me.

artlowel (approved):
Thanks @LucaGiamminonni and @atarix83!

tdonohue:
Merging as this is at +2 and the backend was approved & merged.  Thanks again @LucaGiamminonni and @atarix83 !

PR #1629, opened 2022-04-28
Mirador webpack build fix.

### LINKED ISSUE #1611: Mirador Build Fails
**Describe the bug**
The Mirador viewer build fails with the following: 

```
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "url": require.resolve("url/") }'
        - install 'url'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "url": false }
resolve 'url' in '/Users/michaelspalti/willamette/dura_space/dspace-angular/node_modules/normalize-url'
```

We need to add a webpack polyfill to devDependencies. Also, we should update `webpack.mirador.config.ts` to use the `copy-webpack-plugin` since it's already in our devDependencies and a good replacement for `html-webpack-plugin`

**To Reproduce**
1. execute yarn run build:mirador 

**Expected behavior**
The mirador build should succeed and the files copied to dist.

**Related work**

mspalti:
## References
* Fixes #1611

## Description
The webpack build for Mirador was failing after we upgraded to webpack version 5. It complained about a missing node dependency. This fixes the problem by updating the `webpack.mirador.config` with a fallback.

The "url" dependency is actually not required since Mirador includes a polyfill in its own dependencies, so the fallback is just an empty module: { url: false }. 

We were also missing a webpack dependency: `html-webpack-plugin`.  It appears to be removed from package.json at some point, but frankly it was unnecessary. I just modified the webpack configuration to use the existing `copy-webpack-plugin` instead.

## Instructions for Reviewers
Build Mirador (yarn run build:mirador). The build should succeed and you should be able to see the embedded Mirador viewer inside an iiif-enabled view.

List of changes in this PR:
* Updated webpack.mirador.config.ts with resolve.fallback
* Updated webpack.mirador.config.ts to use copy-webpack-plugin

## Checklist

### DIFF
--- a/src/mirador-viewer/mirador.html
+++ b/src/mirador-viewer/mirador.html
@@ -6,5 +6,6 @@
 </head>
 <body>
 <div id="mirador"></div>
+<script src="mirador.js"></script>
 </body>
 </html>
--- a/webpack/webpack.mirador.config.ts
+++ b/webpack/webpack.mirador.config.ts
@@ -1,4 +1,4 @@
-const HtmlWebpackPlugin = require('html-webpack-plugin');
+const CopyWebpackPlugin = require('copy-webpack-plugin');
 const path = require('path');
 
 module.exports = {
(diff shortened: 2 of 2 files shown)


tdonohue (approved):
👍 Thanks @mspalti !  Tested this today and it fixes the build.  Also verified the Mirador viewer works properly again in the UI.

PR #1698, opened 2022-06-22
Submission item/bitstream access conditions fixes

### LINKED ISSUE #1696: Submission bitstream access condition edit form doesn't work without maxStartDate or MaxEndDate
**Describe the bug**
During submission when a bitstream access condition has a start date or an end date (lease/embrago) without setting the maxStartDate/MaxEndDate the edit form breaks down when selection that specific access condition 

**To Reproduce**
1. on rest side configure the access condition (`config/spring/access-conditions.xml`) `lease` or `embargo` in order to not have endDateLimit or startDateLimit
```
    <bean id="lease" class="org.dspace.submit.model.AccessConditionOption">
        <property name="groupName" value="Anonymous"/>
        <property name="name" value="lease"/>
        <property name="hasStartDate" value="false"/>
        <property name="hasEndDate" value="true"/>
    </bean> 
    <bean id="embargoed" class="org.dspace.submit.model.AccessConditionOption">
        <property name="groupName" value="Anonymous"/>
        <property name="name" value="embargo"/>
        <property name="hasStartDate" value="true"/>
        <property name="hasEndDate" value="false"/>
        
    </bean>
``` 
2. create a new submission and upload a bitstream
3. edit the access condition for the bitstream
4. select one of the access condition changed lease/embrago
5. select a date
6. the form breaks down and it doesn't allow to save 

**Expected behavior**
The form should allow to select a start date or an end date even without setting the maxStartDate/MaxEndDate

### LINKED ISSUE #1697: Submission item access condition step doesn't take in account maxStartDate or MaxEndDate when set
**Describe the bug**
During submission, with the item access condition step, when a item access condition has a start date or an end date (lease/embrago) and setting the maxStartDate/MaxEndDate those values are not taken in account by the section form, so it's possible to select a bigger date.

**To Reproduce**
1. on rest side enable the access condition step by changing the item-submission.xml file
2. create a new submission 
3. in the item access condition step select one of the access condition with a date and a max date. By default both lease/embrago have it
4. select a date
5. verify that a date bigger than the limit set can be selected

**Expected behavior**
A date bigger than the limit set cannot be selected

atarix83:
## References
* Fixes #1696
* Fixes #1697

## Instructions for Reviewers
follow instructions in the issues and check they are fixed 

## Checklist

### DIFF
--- a/src/app/submission/sections/accesses/section-accesses.component.spec.ts
+++ b/src/app/submission/sections/accesses/section-accesses.component.spec.ts
@@ -16,11 +16,15 @@ import { SectionAccessesService } from './section-accesses.service';
 import { SectionFormOperationsService } from '../form/section-form-operations.service';
 import { JsonPatchOperationsBuilder } from '../../../core/json-patch/builder/json-patch-operations-builder';
 import { TranslateModule, TranslateService } from '@ngx-translate/core';
-import { SubmissionJsonPatchOperationsService } from '../../../core/submission/submission-json-patch-operations.service';
+import {
+  SubmissionJsonPatchOperationsService
+} from '../../../core/submission/submission-json-patch-operations.service';
 import { getSectionAccessesService } from '../../../shared/mocks/section-accesses.service.mock';
 import { getMockFormOperationsService } from '../../../shared/mocks/form-operations-service.mock';
 import { getMockTranslateService } from '../../../shared/mocks/translate.service.mock';
-import { SubmissionJsonPatchOperationsServiceStub } from '../../../shared/testing/submission-json-patch-operations-service.stub';
+import {
+  SubmissionJsonPatchOperationsServiceStub
+} from '../../../shared/testing/submission-json-patch-operations-service.stub';
 import { BrowserModule } from '@angular/platform-browser';
 
 import { of as observableOf } from 'rxjs';
@@ -42,8 +46,6 @@ describe('SubmissionSectionAccessesComponent', () => {
   let fixture: ComponentFixture<SubmissionSectionAccessesComponent>;
 
   const sectionsServiceStub = new SectionsServiceStub();
-  // const pathCombiner = new JsonPatchOperationPathCombiner('sections', sectionId, 'files', fileIndex);
-
   const builderService: FormBuilderService = getMockFormBuilderService();
   const submissionAccessesConfigService = getSubmissionAccessesConfigService();
   const sectionAccessesService = getSectionAccessesService();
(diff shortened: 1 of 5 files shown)


tdonohue (changes requested):
@atarix83 : I tested this today & it partially works.

I see that #1696 is fixed.  So, that's good.

However, #1697 is NOT fixed as far as I can see.  It now throws an exception (in the DevTools) and results in a 500 error on the backend.

Here's what I tried:
1. First, I used the default embargo setting for testing, which includes `<property name="startDateLimit" value="+36MONTHS"/>`
2. I started a new submission, uploading a file.
3. I selected to embargo the file, and set the start date at 5 years from now. (this should be invalid as the max is 3 years)
4. After clicking Save, nothing happens in the UI.  The popup refuses to close or save the value.  The "Save" button disappears in the popup, and there is no way to close it.
5. In my Chrome DevTools, I see that a PATCH was sent to the backend and the backend threw a 500 error.  Here's the stacktrace:
```
org.springframework.data.rest.webmvc.json.patch.PatchException: Error processing the patch request
at org.dspace.app.rest.submit.SubmissionService.evaluatePatchToInprogressSubmission(SubmissionService.java:455)
at
org.dspace.app.rest.repository.WorkspaceItemRestRepository.patch(WorkspaceItemRestRepository.java:212)
at org.dspace.app.rest.repository.WorkspaceItemRestRepository.patch(WorkspaceItemRestRepository.java:75)
at org.dspace.app.rest.repository.WorkspaceItemRestRepository$$FastClassBySpringCGLIB$$987ae
...
Caused by: java.lang.IllegalStateException: The start date of access condition embargo should be earlier than +36MONTHS
from now.
```

So, it appears that the backend knows that the date is invalid & sent back a 500 exception. But, the frontend either needs to know how to process that 500 exception or avoid sending the PATCH request if it is known to be invalid.

atarix83:
@tdonohue 

i found out it happened when the access condition was already selected. I that case the max start date wasn't set properly.
Now it should be fixed, you could not be able to select a date bigger than the limit now.
Anyway I agree the rest server should handle the invalid date case properly, so I've opened a new issue that could be address in 7.4 https://github.com/DSpace/DSpace/issues/8371

tdonohue (approved):
👍 Thanks @atarix83 !  Retested this today and the bug I saw before is fixed.  I can no longer select a date that is *past* the maximum from the UI.

This PR seems like much improved behavior based on current `main`.  So, I'm going to merge immediately to get it into 7.3.  If we find any later bugs we can address them at a later time.

PR #1779, opened 2022-08-19
Make create, edit community/collection/item dialogs theme-able.

mwoodiupui:
## Description
See the title.  This is an elaboration of local work that was needed to create a custom new-item dialog in our local theme (instead of hacking the stock dialog).

## Instructions for Reviewers
List of changes in this PR:
* Create themed wrappers for the dialog components.
* Use them in the menu resolver, in place of the un-themed components.

## Checklist


mwoodiupui:
This is draft because I'm running Node v18 but the specified Webpack and eslint-plugin-jsdoc require an earlier version.

artlowel (changes requested):
Thanks @mwoodiupui 

What you've written here looks correct at first sight, but a few things are missing:
- Those new ThemedComponents have to be declared in a module for them to work, the same module the components they're theming are declared in, I think that's SharedModule in this case
- When theming a component you should provide an example themed version in the custom theme. See [HeaderComponent](https://github.com/DSpace/dspace-angular/blob/1d4c77e48557eead566dda2f2f65ba04df2df851/src/themes/custom/app/header/header.component.ts) in the custom theme for example That way, people that want to use your themed component have something to start from. You can use that version in the custom component yourself as well to verify that everything is wired up correctly. If you modify it, and use the custom theme, those modifications should show up

src/themes/custom/app/shared/dso-selector/modal-wrappers/create-collection-parent-selector/create-collection-parent-selector.component.ts
```diff
@@ -0,0 +1,11 @@
+import { Component } from '@angular/core';
+import {
+    CreateCollectionParentSelectorComponent as BaseComponent
+} from 'src/app/shared/dso-selector/modal-wrappers/create-collection-parent-selector/create-collection-parent-selector.component';
+
+@Component({
+  selector: 'ds-create-collection-parent-selector',
+  templateUrl: '../dso-selector-modal-wrapper.component.html',
+})
```
artlowel, line 9: The convention for the custom theme is to already create html and css file for the component, but add them in comments and refer back to the version in the base theme.

That means that if you enable the theme, everything looks exactly like the base theme. Then if you want to start theming the component all you need to do is swap the comments around and you can start overwriting the css or the html right away.

Also, I've noticed that using absolute paths in imports tends to cause build problems in certain scenarios that are hard to diagnose, so I'd suggest always using relative paths, even though that can lead to a ridiculous number of `../`-es, like in this case. However most IDEs can keep count for you.

```suggestion
import {
    CreateCollectionParentSelectorComponent as BaseComponent
} from '../../../../../../../app/shared/dso-selector/modal-wrappers/create-collection-parent-selector/create-collection-parent-selector.component';

@Component({
  selector: 'ds-create-collection-parent-selector',
  // styleUrls: ['create-collection-parent-selector.component.scss'],
  // templateUrl: 'create-collection-parent-selector.component.html',
  templateUrl: '../../../../../../../app/shared/dso-selector/modal-wrappers/dso-selector-modal-wrapper.component.html',
})
```

src/themes/custom/app/shared/dso-selector/modal-wrappers/dso-selector-modal-wrapper.component.html
```diff
@@ -0,0 +1,11 @@
...
+  <div class="modal-header">{{'dso-selector.'+ action + '.' + objectType.toString().toLowerCase() + '.head' | translate}}
+    <button type="button" class="close" (click)="close()" aria-label="Close">
+      <span aria-hidden="true">×</span>
+    </button>
+  </div>
+  <div class="modal-body">
+    <h5 *ngIf="header" class="px-2">{{header | translate}}</h5>
+    <ds-dso-selector [currentDSOId]="dsoRD?.payload.uuid" [types]="selectorTypes" (onSelect)="selectObject($event)"></ds-dso-selector>
+  </div>
+</div>
```
artlowel, line 11: Rather than moving this file to the theme folder when it's not itself a themed component, I'd reference the version in the base theme as detailed in a comment above, and delete this file

artlowel (commented):
Thanks @mwoodiupui 

I have a few more comments to bring these in line with the other components in the custom theme. You picked an unconventional set of components to get started with (ones that share a single html template) which complicates things a bit

src/themes/custom/app/shared/dso-selector/modal-wrappers/create-collection-parent-selector/create-collection-parent-selector.component.ts
mwoodiupui: Done, I think.

src/themes/custom/app/shared/dso-selector/modal-wrappers/dso-selector-modal-wrapper.component.html
mwoodiupui: Yes, this file becomes useless.  Done.

mwoodiupui:
I have a knack for starting out to learn with a problem that seems simple but turns out to be difficult or unusual.  Thanks for a thorough critique.

artlowel (approved):
No problem, thanks for incorporating it in your PR!

PR #1841, opened 2022-09-20
Add label for the citation page curation task

abollini:
## References
https://github.com/DSpace/DSpace/pull/8417

## Description
Add the i18n key for the citation page curation task


tdonohue (approved):
👍 Thanks @abollini .  Verified this is correct when testing https://github.com/DSpace/DSpace/pull/8417

PR #1847, opened 2022-09-21
Bug fix for multiple IIIF image property

mspalti:
## Description
Fixes a bug in `mirador-viewer.component` that incorrectly sets the `multi` property when images are added to bundles other than `ORIGINAL`.

The PR modifies the Angular UI to look for images in ALL eligible Bundles, not just  ORIGINAL.

**Some quick background:** the `multi` property is set by the `mirador-viewer.component` component and is passed to the Mirador viewer, which in turn uses the property to create internal configuration at startup. (The default Mirador behavior is to add right thumbnail navigation when `multi` is 'true.') `multi` should be true whenever the Item has multiple image bitstreams.

_Sorry! I just realized I didn't submit this bug fix earlier!! It's a fairly minor fix so if it can make it into 7.4 that would be great._

## Instructions for Reviewers
* Create an IIIF-enabled Item with more than one image bitstream. Upload the images into a custom bundle (e.g. TestBundle).
* View the Item and check to be sure the images are found in the custom bundle by checking for right navigation thumbnail images in the Mirador viewer (see below). Remember that (unfortunately) the viewer can only be tested in `production` mode.

Screen Shot 2022-09-21 at 1 00 26 PM

 

## Checklist

### DIFF
--- a/src/app/item-page/mirador-viewer/mirador-viewer.component.spec.ts
+++ b/src/app/item-page/mirador-viewer/mirador-viewer.component.spec.ts
@@ -12,6 +12,7 @@ import { createPaginatedList } from '../../shared/testing/utils.test';
 import { of as observableOf } from 'rxjs';
 import { MiradorViewerService } from './mirador-viewer.service';
 import { HostWindowService } from '../../shared/host-window.service';
+import { BundleDataService } from '../../core/data/bundle-data.service';
 
 
 function getItem(metadata: MetadataMap) {
@@ -46,6 +47,7 @@ describe('MiradorViewerComponent with search', () => {
       declarations: [MiradorViewerComponent],
       providers: [
         { provide: BitstreamDataService, useValue: {} },
+        { provide: BundleDataService, useValue: {} },
         { provide: HostWindowService, useValue: mockHostWindowService }
       ],
       schemas: [NO_ERRORS_SCHEMA]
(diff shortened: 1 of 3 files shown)


tdonohue (commented):
@mspalti : I tried testing this today & I cannot get it to work as described.

1. I setup an Item with an existing image in the ORIGINAL bundle.  Verified it loads in Mirador & it does
2. Then, I edited that Item, created a new bundle named "Test" and added a new image to it.
3. Went back to the Item page & looked in Mirador. It still only displays the single image in the ORIGINAL bundle.
4. Reloaded page in browser (in case there's caching).  Still only displays the single image in the ORIGINAL bundle.
5. Then tried to add a _new_ image to the ORIGINAL bundle.  It also doesn't show up in Mirador.

Is there some sort of caching here that I need to refresh?  I cannot figure out how to get this to work at all.

UPDATE: Does this maybe require https://github.com/DSpace/DSpace/pull/8402 ??

mspalti:
@tdonohue I will take a look. It could be a bug in the implementation. I think it should have worked given what you tried. There's no relationship to DSpace/DSpace#8402

atarix83 (approved):
thanks @mspalti 

I made the following test : 
- tested successful with an item with bitstreams in the ORIGINAL bundle and in a custom one
- tested successful with an item with bitstream with only a custom bundle
- test failed with an item with an empty bundle ORIGINAL and bitstreams in a custom one. I found out this fails due to an issue on the rest side, indeed the request to the manifest fail with 500 error. we'll open an issue on rest side

mspalti:
Thanks @atarix83 , I see the problem on the backend!  The problem should be addressed in DSpace/DSpace/#8402 which currently is under review.

tdonohue (approved):
👍 Thanks @mspalti !  I've retested and verified everything is working perfectly now.

PR #1858, opened 2022-09-26
Hindi Translation 

abhishekd2t:
Here I am submitting new file for hindi translation for DSpace7. 

Thanks.


tdonohue (changes requested):
@abhishekd2t : Thanks for your contribution! I tried to enable this language pack today, but I'm not able to get it to work.

It should be possible to enable it on your end by adding this:
```
 { code: 'hi', label: 'हिंदी', active: true}
```
To this list of languages: https://github.com/DSpace/dspace-angular/blob/main/src/config/default-app-config.ts#L206

And then rebuild & try it out.  I keep getting errors on different lines of this file, which means the file it's fully valid yet.  I listed the *first* error I see below, but even after fixing that one, I'm getting more errors.

There's still time to fit this into 7.4 if we can get it ready on Monday or Tuesday.  I'll be glad to test it again early next week whenever you feel it is ready.

tdonohue (approved):
👍 Thanks @abhishekd2t !  This looks good now and it is working for me.  I'll merge this in time for the 7.4 release.

PR #1884, opened 2022-10-05
Misspellings corrections in pt-BR

lucaszc:
Misspellings  corrections in portuguese brazil

## Description
Misspelling corrections in portuguese brazil

## Instructions for Reviewers
  Corrections of prepositions, pronouns,  gender in the language, and adapt to the way it is translated in other parts of the interface.


tdonohue (approved):
👍 Thanks @lucaszc !  I just verified that the language pack still loads properly and it does.

PR #1893, opened 2022-10-10
[#1892] Apply type-bind properly to concat fields that are not dynamic arrays

### LINKED ISSUE #1892: Type bind submission fields with CONCAT type model are not hidden properly
**Description**
If a field in `submission-forms.xml` is configured to be of `<input-type>name</input-type>` and `<repeatable>false</repeatable>` and has a type-bind value configured, the UI does not properly hide / disable the form control or row when the bound type is not selected.
A name field which is repeatable behaves as expected with type-bind, and so does a non-name field which is not repeatable.

The backend metadata validation and type-bind enforcement appears to work well, as the metadata for these buggy fields is stripped out from the item after the submission is completed and moves to workflow / archive, but the Angular UI form services and components are not treating this specific case properly.

This was discovered in DSpace 7.3 by Agustín Alfieri and reported in Slack.

There are possible other less-common field models / config combinations that also result in this unexpected behaviour.

**To Reproduce**
1. Configure a field in submission forms default form that uses `<input-type>name</input-type>` and `<repeatable>false</repeatable>` and <type-bind>Article</type-bind> (or some type)
2. Configure another field as above, but with <repeatable>true</repeatable>
3. Configure another field as step 1 but with a different input type
4. Build DSpace and submit a new item to a collection using the modified submisison form
5. Test behaviour as the type dropdown is changed - you will see that the fields from step 1 and step 3 seem to obey type binding rules and are hidden when their type is not selected, but the field from step 2 remains visible at all times

**Expected behavior**
The field from step 2 should also be hidden when its bound type is not selected.

**Related work**
#1529 
https://github.com/DSpace/DSpace/pull/8175
https://github.com/DSpace/DSpace/pull/8296

kshepherd:
**References**
* Fixes #1892 

## Description
A bug was discovered with name fields that are not repeatable - they do not obey type bind.
This was found to be because the form control model used is `dynamic-concat-model` and this model did not specifically set required type bind configuration.

This pull request fixes the issue for concat model, but there may be other form control models around that have the same issue. If more are found in the short term, I will rename this issue / PR and continue committing fixes to other form control models here.

## Instructions for Reviewers

* See instructions at #1892 

- [x] My PR passes all specs/tests and includes new/updated specs or tests based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide)

### DIFF
--- a/src/app/shared/form/builder/ds-dynamic-form-ui/models/ds-dynamic-concat.model.ts
+++ b/src/app/shared/form/builder/ds-dynamic-form-ui/models/ds-dynamic-concat.model.ts
@@ -1,4 +1,10 @@
-import { DynamicFormControlLayout, DynamicFormGroupModel, DynamicFormGroupModelConfig, serializable } from '@ng-dynamic-forms/core';
+import {
+  DynamicFormControlLayout,
+  DynamicFormControlRelation,
+  DynamicFormGroupModel,
+  DynamicFormGroupModelConfig,
+  serializable
+} from '@ng-dynamic-forms/core';
 
 import { Subject } from 'rxjs';
 
@@ -16,6 +22,7 @@ export interface DynamicConcatModelConfig extends DynamicFormGroupModelConfig {
   separator: string;
   value?: any;
   hint?: string;
+  typeBindRelations?: DynamicFormControlRelation[];
   relationship?: RelationshipOptions;
   repeatable: boolean;
   required: boolean;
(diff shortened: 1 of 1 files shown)


jtimal:
Hi @kshepherd,
I tried these changes and they work fine for me as explained in PR 1892
image
when I change the type they all disappear
image

tdonohue (approved):
👍 Thanks @kshepherd .  I've also tested and verified that the previous reported type-bind bug is no longer reproducible.  Code looks good as well.

PR #1895, opened 2022-10-12
Fix button translation ...

lucaszc:
Fix button translation ...


tdonohue (approved):
👍 Thanks @lucaszc !  Verified this button is no longer hardcoded and is now translated.

PR #1896, opened 2022-10-12
Add missing translations in version 7.4 pt-BR

lucaszc:
Add missing keywords, and correct translations in pt-BR


tdonohue (approved):
👍 Thanks for the updates @lucaszc !  I gave this a quick test to ensure the new keys are working and they appear to be.

PR #1910, opened 2022-10-18
Update pt-BR to replace ficheiro (pt-PT)

lucaszc:
Ficheiro is used only in pt-PT in pt-BR is used arquivo


tdonohue (approved):
👍 Thanks @lucaszc !  Merging this immediately as this are just small textual replacements in the translation file.

PR #1940, opened 2022-11-01
Add: Ukrainian translation

AndrukhivAndriy:
Add: Ukrainian translation. Tested on v7.4 . Together to victory against the russian invasion


tdonohue (approved):
👍 Thank you @AndrukhivAndriy !  (And apologies for the delay in reviewing this work)   I've verified that this translation file works properly in DSpace 7, and did a very quick verification (using Google Translate) that the translations themselves look reasonable.  I'll merge this immediately and follow it up with a small PR to enable this Ukrainian language option by default.

PR #1978, opened 2022-12-01
Edit bitstream annotation metadata

mspalti:
## References
* Requires DSpace/DSpace#8595 

## Description
Adds IIIF Image Annotation field to the edit bitstream page. The metadata field is used by the backend annotation service to provide image annotations that can be viewed in IIIF viewers, including Mirador.

## Instructions for Reviewers

List of changes in this PR:
* Added form element to `edit-bitstream-page.component`.
* Minor additions to the mirador `index.js` file that can be used to configure annotation behavior in the viewer.

This PR can be used when testing DSpace/DSpace#8595. Edit a bitstream add IIIF annotations to the "IIIF Image Annotations" field (note that IIIF must be enabled for the Item). You should see these results in Mirador when viewing the Item. 

- If the annotations are valid IIIF Presentation API version 2.1.1 or 3 annotations, the mirador viewer will add an annotations icon to the left panel. 
- The icon should include an orange dot, indicating annotations are present. 
- If the annotations were created specifically for the image you are viewing, you should see sections of the image highlighted. 

This snippet might also be useful in testing. It's an IIIF Presentation API v3 AnnotationPage for a specific image in my test environment. It will mostly work in a different test environment. However, image highlights will not appear in the viewer because the canvas "id" and the coordinates are specific to this image. 

`{"id":"localStorage://?canvasId= http://localhost:8080/server/iiif/3017920a-6d27-4536-9915-10fd70e92948/canvas/c0","items":[{"body":{"type":"TextualBody","value":"<p><b>Test Annotation</b></p>\n<p>Test annotation body.</p>"},"id":"998994af-93ab-48c6-a8ae-63672658b16d","motivation":"commenting","target":{"source":"http://localhost:8080/server/iiif/3017920a-6d27-4536-9915-10fd70e92948/canvas/c0","selector":[{"type":"FragmentSelector","value":"xywh=304,612,1194,2011"}]},"type":"Annotation"}],"type":"AnnotationPage"}`

## Checklist


mspalti:
https://github.com/DSpace/DSpace/pull/8595 is being refactored to avoid using a metadata field for the Json annotation.

mspalti:
This PR needs to be changed to support annotation file uploads, rather than the current approach of just adding JSON to a bitstream metadata field. I'm happy to make the changes to this PR for the 7.5 release. 

Essentially, we need to have a new upload component that appears with the iiif options on the bitstream page. Upon upload, the component will add the annotation file to the correct bundle using the bitstream uuid as the filename.  e.g. `eb12635a-2682-413b-8c18-ec35c05badac.json` If an annotation file already exists for the bitstream, the file should be listed with an option to delete.

mspalti:
@tdonohue please feel free to move this to the 7.6 release project whenever you're ready. The annotation work is taking a different direction per @abollini's suggestion and hopefully we can get it added in the next release.

tdonohue:
@mspalti : I'll move both PRs to 7.6 then.  Thanks for the note!

tdonohue:
Per a decision in DSpace Steering (and as announced in last DSpace DevMtg), I'm moving this (and DSpace/DSpace#8595) to the 8.0 board. Steering has limited 7.6 to only features which provide "feature parity/compatibility" with 6.x. Even though this will move to the 8.0 board, our goal is to move all existing pre-8.0 features forward as soon as 7.6 development is finishing (as 7.6 will be the last 7.x release allowing new features).

mspalti:
Hi @tdonohue, I anticipated that this would get moved to v8.0 and that makes sense. The work on this version of the code is complete now. For now, I'll add some screenshots and commentary so this will make sense when we return to it.

mspalti:
I chose to incorporate the annotation upload feature into the bitstream edit page.  To add annotation file:

1. Open the bitstream edit page
2. Choose the IIIF annotation upload button
3. Upload the file.
4. The IIIF annotation button changes to show that an annotation exists for this image

You can later (or immediately) delete the annotation file

1. Open the bitstream edit page
2. Choose the IIIF annotation button
3. Delete the file
4. The IIIF annotation button changes to show the upload option
5. Upload a new file if necessary ....

This is what happens behind the scene:

1. If no annotation bundle exists, one is created
2. The uploaded image is added to the annotations bundle; the file is named _<parent.bitstream.uuid>.json_
3. When a file is deleted I don't remove the annotations bundle because deletion doesn't appear possible at the moment. Note that the bundle should ONLY be deleted if there are no remaining annotation files.
4. This process works for Items that include multiple images because that annotation bundle can contain multiple files that follow the naming convention _<parent.bitstream.uuid>.json_.

Here are a few screenshots from the workflow as currently implemented. 

Initial screen
Screenshot 2023-02-27 at 3 47 59 PM

Upload form
Screenshot 2023-02-27 at 3 43 07 PM

Upload successful
Screenshot 2023-02-27 at 3 43 46 PM

Deleting the existing file
Screenshot 2023-02-27 at 3 44 18 PM

mspalti:
You'll notice that I slipped in a new form field for `iiif.description`. This is very useful since it offers an additional way to add image descriptions to the canvas. One could use it exclusively for canvas descriptions, or in combination with `dc.description` depending on your needs and local configuration.

mspalti:
Looks like the build failure is unrelated.

mspalti:
This component needs to maintain updated copies of `item.bundles` and `item.bundle.bitstreams`. It does this by requesting local copies via `itemService.findById`, with `useCachedVersionIfAvailable` false.

I tried to optimize this by setting `useCachedVersionIfAvailable` false in the `followLinks()` configuration for bundles and bitstreams. This worked most of the time, but I noticed that the button doesn't appear in some cases. This is fairly rare, but it does happen, and that seems to indicate that either the bundles or the bitstreams are not defined on the item.

mspalti:
I considered making this feature a standalone page by adding an 'annotation' option in the bitstream options:

Screenshot 2023-03-03 at 11 19 20 AM

I think this is a good idea but in the initial implementation I opted against it for two reasons:

1. It would require IIIF status exist the bundle object so that we can tell when to add the annotations option. That will require more discussion but is simple to add if it makes sense to do so.
2. The annotation feature is important and a good selling point for DSpace but I would expect annotations to be relatively uncommon. So it seems less urgent to feature them in the bitstream options list (although I don't think this is an argument against it).

PR #2030, opened 2023-01-05
Context help tooltips

KoenP:
## Description
This PR adds a purely frontend feature to add context help buttons with tooltips in the interface. More precisely:

- We add a new "context help toggle" button in the header to toggle the visibility of context help buttons (by default, they are not visible).
- We add a new structural directive `dsContextHelp` which can be used to add a context help button to any HTML element. Its sole parameter is an object with mandatory fields `id` (an identifier to refer to the button in the store) and `content` (an identifier for the translate service); and with optional fields `iconPlacement` and `tooltipPlacement`.
    * If tooltip visibility is toggled on, a context help button will appear next to the the HTML element the directive was applied to (to the left or right, depending on the `iconPlacement` parameter).
    * When this button is clicked, a tooltip appears in a text bubble pointing towards the bubble. The relative location of the tooltip compared to the button can be configured with the `tooltipPlacement` parameter.
    * Clicking anywhere outside of the tooltip bubble makes the tooltip disappear again.
    * If the tooltip text contains Markdown-style links, these are automatically converted to clickable links. Note that only Markdown _links_ are supported (for now, we wanted to avoid pulling in a large Markdown dependency).
- We added some tooltips to the edit group page (`/access-control/groups/<group-id>`) as a demonstration of this feature.

## Instructions for Reviewers
List of changes in this PR:
* First, ...
* Second, ...

### To interact with the provided demonstration
1. Log in as admin, go to Access Control -> Groups and select a group to edit.
2. Click the question mark button in the top right, between the language menu and login menu.
3. New question mark buttons appear near "Edit Group", "Add EPeople" and "Add Subgroup": you can try clicking them, toggling visibility multiple times, try clicking the link in the "Edit group" sample text, etc.

### To add your own context help buttons
1. Open a template of your choice in an editor and choose an HTML element to provide context help for. 
2. Add a new entry to the i18n file to hold the content of your tooltip. You can use Markdown-style links in the tooltip text to create clickable links for the user.
3. Add a `*dsContextHelp` attribute to the element and fill in its parameter. Make sure at least the `content` field is set to the key for your new entry in the i18n file, and the `id` field is set to some unique string.
4. Recompile the application and navigate to the page you just edited. Make sure to toggle the context help button visibility on with the button in the top right corner.

## Checklist


src/app/header/context-help-toggle/context-help-toggle.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  selector: 'ds-context-help-toggle',
+  templateUrl: './context-help-toggle.component.html',
+  styleUrls: ['./context-help-toggle.component.scss']
+})
+export class ContextHelpToggleComponent implements OnInit, OnDestroy {
+  buttonDisabled$: Observable<boolean>;
+
+  constructor(
+    private contextHelpService: ContextHelpService,
+    private translateService: TranslateService
```
atarix83, line 17: it seems to be not used, could you remove it ?

src/app/header/context-help-toggle/context-help-toggle.component.ts
```diff
@@ -0,0 +1,40 @@
...
+import { TranslateService } from '@ngx-translate/core';
+import { Observable, Subscription, BehaviorSubject } from 'rxjs';
+import { combineLatest } from 'rxjs';
+
+@Component({
+  selector: 'ds-context-help-toggle',
+  templateUrl: './context-help-toggle.component.html',
+  styleUrls: ['./context-help-toggle.component.scss']
+})
+export class ContextHelpToggleComponent implements OnInit, OnDestroy {
```
atarix83, line 12: could you add typedocs to this class?

src/app/shared/context-help.service.ts
```diff
@@ -0,0 +1,113 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class ContextHelpService {
+  constructor(private store: Store<ContextHelpState>) { }
+
+  /**
+   * Observable keeping track of whether context help icons should be visible globally.
+   */
+  shouldShowIcons$(): Observable<boolean> {
```
atarix83, line 41: the $ suffix is usually used for variables not for the methods. I'd rename all the methods with $ suffix

atarix83 (changes requested):
Thanks @KoenP for this PR

The functionality is overall good and it works fine. I've only a few inline comments to address

src/app/shared/context-help.service.ts
artlowel: I'm not sure I agree with that. Sure it's not the current convention in DSpace, but adding the dollar suffix makes it very clear that it'll return an observable, and I can't imagine what the downside would be

KoenP: My two cents: Adding the suffix to method names makes naming the sample in lambdas easier. For instance, if we write `shouldShowIcons$().subscribe(shouldShowIcons -> ...)`, we can have both the observable(-producing method) and its sample in scope without issue. Omitting the $ suffix means we have to either shadow the outer variable (disallowed by the linter, IIRC), or come up with some name variation for the sample every time, neither of which seem satisfying.

That said, I also follow the argument for consistency with other method names, and I'll remove the suffix if you prefer it.

tdonohue: We appear to already have many methods that end in a dollar sign ($).  A quick search of the codebase shows we have methods ending in a dollar sign in `theme.service.ts`, `init.service.ts`, `menu.resolver.ts`, etc.  In all instances, it looks like those methods are returning Observables, just like this one.

We also have examples of many methods (sometimes even in the same file, see `menu.resolver.ts`) that return Observables but do *not* end in a dollar sign.

So, this is just to point out that we don't seem to have any firm agreement as to whether methods can or cannot end in a dollar sign.  Therefore, this code is fine as-is, or you can change it. It's really up to @KoenP and this request should be considered *optional* as we don't have a firm best practice.

(If we wish to establish a best practice in the future, we can always do so. But, at this point, we just don't have one.)

tdonohue (changes requested):
@KoenP : Overall this looks good & it works.  I do have a few small usability issues I see though.

1. First, the gray of the "?" icon seems to overlap into the globe in the header... when I hover over the globe, it now turns light gray (it doesn't do this on `main`).  Screenshot:
question-mark-styles
2. I don't like that the "?" icon appears in the header on all pages, as it's currently grayed out on every page except the "Edit Group" one (as that's the only page with help text).  A few possible solutions come to mind:
    * I wonder if we could only display it if help text *exists* on the page? (Rather than graying it out, it only appears on pages when it is enabled)
    * If it helps, we could even consider removing it from the header altogether and moving it into the page contents itself... placing it in the upper right of the page, similar to where the "Edit" button appears when logged in: 
Untitled

My first point is the bigger issue here.  Regarding my second point, I suspect that if we leave the "?" grayed out on (almost) every page, then we likely will get complaints as people upgrade (e.g. "Why doesn't this button work on any page?").  So, we may want to think of a different way to hide/not display it when help text doesn't (yet) exist.

artlowel:
@tdonohue:

>1. First, the gray of the "?" icon seems to overlap into the globe in the header... when I hover over the globe, it now turns light gray (it doesn't do this on `main`).

That's a bug fix. The globe always had that hover style, but it didn't work because of [an errant `!important` in the css](https://github.com/DSpace/dspace-angular/pull/2030/commits/8850f005b4831642735b24ef603fef0ffb95ba22#diff-b82ced23f0d0635fbdcce526e38729dc49e2e028cd5afd36b5150b5286801663L7).

>2. I don't like that the "?" icon appears in the header on all pages, as it's currently grayed out on every page except the "Edit Group" one (as that's the only page with help text).  A few possible solutions come to mind:
    * I wonder if we could only display it if help text *exists* on the page? (Rather than graying it out, it only appears on pages when it is enabled)
    * If that's difficult to achieve in the header, we could even consider removing it from the header altogether and moving it into the page contents itself... placing it in the upper right of the page, similar to where the "Edit" button appears when logged in: 
   
I'd prefer your first suggestion: to only show the question mark when there's context help available, since the menu in the upper right corner of the page only exists on DSO pages for now, and adding that everywhere would be a lot of work and touch a lot of files. So we'll keep it in the header, but hide it if there's no context help on the page

tdonohue:
@artlowel : Regarding the first point (the gray hover over).  I specifically called that out because the behavior in this PR looks odd to me.  Only the "?" and the globe turn gray when you hover over them.  The "Login" and the search icon do *not*.   So, it looks odd that only 2 of the 4 options in that menu behave in this manner.  In my opinion, all the options should act the same... either none have hover overs, or all do.  If we go with the hover over, we may want to look at how it appears for all icons/menus, as I'm worried it's a very light gray (which fades into the white header too much).

Overall, I'm not against the hover over...but I think we should look for consistency.

artlowel:
@tdonohue looks like the login link has the same `!important` preventing the hover style below from being activated. We'll fix it there as well. It is strange however that it looks light grey to you, to me that hover style is dark blue, same as the search icon:
image

src/app/header/context-help-toggle/context-help-toggle.component.ts
KoenP: Thanks for the feedback! I removed the `translateService` parameter from the constructor.

src/app/header/context-help-toggle/context-help-toggle.component.ts
KoenP: Added the docs.

src/app/shared/context-help.service.ts
KoenP: I've left this as-is for now.

KoenP:
@tdonohue As requested, we now hide the context help toggle if there are no tooltips on the current page, and we made the header buttons behave consistently on hover.

tdonohue (approved):
👍 Thanks @KoenP and @artlowel ! This now looks good to me.  I've also verified that the hover-over is no longer light gray for me (I suspect I may have had old styles cached in my browser when that was occurring).

After we merge this, I think it would be good to write up a basic "guide" on how to add additional "context help tooltips" (on other pages), using the example in this PR as our initial example.  Likely we should add that guide to https://wiki.lyrasis.org/display/DSDOC7x/User+Interface+Customization or perhaps to a new sub-page of that wiki page.

KoenP:
Thanks all for the reviews! 🎉 

@tdonohue, I don't think I have editing rights on the wiki so I'll just post the documentation here in this comment, as the Markdown syntax should be the same anyway, I think.

## Contextual Help Tooltips
Contextual help tooltips are a feature to provide additional information about how to use DSPace to less experienced 
users without cluttering the interface for more advanced users who do not need additional instruction.

### User perspective
If the user visits a page where contextual help tooltips are available, a "toggle context help" button appears in the
header, in between the language switch menu and user profile menu. Clicking this button toggles the visibility
of the tooltips on the page (by default, they are invisible).

Screenshot 2023-02-08 at 09 13 21

When tooltip visibility is turned on, similar looking buttons appear on the page where ever a tooltip is available.
Clicking any of these buttons makes a text bubble appear containing the contextual help; clicking anywhere outside
of the bubble makes it disappear again.

Screenshot 2023-02-08 at 09 12 52

### Adding new tooltips
Any HTML element can be given a tooltip by setting the `*dsContextHelp` attribute.
The value assigned to the attribute should be an object of type `ContextHelpDirectiveInput`:
```typescript
export interface ContextHelpDirectiveInput {
  content: string;
  id: string;
  tooltipPlacement?: PlacementArray;
  iconPlacement?: PlacementDir;
}
```
- The mandatory `content` field represents a key in i18n files. 
- `id` should be a unique identifier for this tooltip, to distinguish it from other tooltips on the page. 
- `tooltipPlacement` (optional) determines where the text bubble appears relative to the help button.
  Its type is an array of `Placement`s; see the 
  [ng-bootstrap documentation](https://ng-bootstrap.github.io/#/guides/positioning#api) for more information.
- `iconPlacement` (optional) should be assigned either `'left'` or `'right'`, and determines whether the tooltip will be 
placed on the left or on the right of the element.

This is what the template looks like for the "Edit group" example in the picture above:
```html
<h2 class="border-bottom pb-2">
          <span
            *dsContextHelp="{
              content: 'admin.access-control.groups.form.tooltip.editGroupPage',
              id: 'edit-group-page',
              iconPlacement: 'right',
              tooltipPlacement: ['right', 'bottom']
            }"
          >
            {{messagePrefix + '.head.edit' | translate}}
          </span>
</h2>
```
Note the use of the `span` tags: setting `*dsContextHelp` directly on the `h2` element makes the help button appear
all the way on the right of the page, instead of directly to the right of the "Edit group" text.

tdonohue:
Thanks @KoenP for the documentation. I've moved this over to https://wiki.lyrasis.org/display/DSDOC7x/Contextual+Help+Tooltips and made a few minor updates.

Issue #2096, opened 2023-02-15
Strange wording for file upload in submission form


alanorth:
**Describe the bug**
As of DSpace 7.5-SNAPSHOT the item submission form uses strange English when informing the user how to upload a bitstream:

> _You can update the file metadata and access conditions or upload additional files just dragging & dropping them everywhere in the page_

There should be a better way to say this in English. Also, the sentence is missing a period at the end.

PR #2097, opened 2023-02-15
src/assets/i18n: Improve English strings for file upload

alanorth:
Simple improvement to the wording in the file upload section of the item submission form. Also add a period at the end of the string.

Fixes: #2096


tdonohue (approved):
👍 Thanks @alanorth !  I agree this wording is better.  Gave it a quick test and the updates are working.

PR #1512, opened 2022-02-09
Various small changes to improve customizability

ybnd:
## Description
This PR bundles multiple small fixes & adjustments we've made to better support custom themes

Also, Fixes #1528 (see https://github.com/DSpace/dspace-angular/pull/1512/commits/fba1dfb689b18263b5c7ef9e92f301f6d2e2b528 and https://github.com/DSpace/dspace-angular/pull/1512/commits/2830da955739b61c91c15f587051b0d75f4e5464)

## List of changes in this PR
* Made a few more components themeable
  * `ComcolPageBrowseComponent`
  * `AdminSidebarComponent`
  * `SearchComponent`
  * `ItemListPreviewComponent`
  * `CommunityListComponent`
* Propagate the `SearchFormComponent.showScopeSelector` input down from `SearchComponent` so we can get at it from a theme
   Note that this change was made before https://github.com/DSpace/dspace-angular/commit/c28221623ca1405469db14459e1cffe92a65d575, so the commit in question has mostly been overwritten. Make sure to look at the file _after_ the `main` merge.
* `DSOSelectorComponent`: use `DSONameService` for more flexibility
* `ds-truncatable`
   * support Bootstrap `$light` background
   * fix RGBA/RGB gradient issue on Safari, e.g. before fix:

* Fix some minor issues with thumbnails / grid view
* `ThemedComponent`: propagate `ngOnChanges` calls down to the wrapped component in order to support theming for components that use it.

## Checklist
_This checklist provides a reminder of what we are going to look for when reviewing your PR. You need not complete this checklist prior to creating your PR (draft PRs are always welcome). If you are unsure about an item in the checklist, don't hesitate to ask. We're here to help!_

- [x] My PR is small in size (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR passes [TSLint](https://palantir.github.io/tslint/) validation using `yarn run lint`
- [x] My PR doesn't introduce circular dependencies
- [ ] My PR includes [TypeDoc](https://typedoc.org/) comments for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [x] My PR passes all specs/tests and includes new/updated specs or tests based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
[...]

PR #2235, opened 2023-05-09
Fixed `ngOnChanges` not working for themed components & decorator components + minor fixes

### LINKED ISSUE #2201: `ngOnChanges` hook does not work in themes
**Describe the bug**
When overriding a stock component in a theme, the changes hook is never called.  The instance variable `compRef` is used in `src/app/shared/theme-support/themed-component.ts:ngOnChanges` but that variable is not set until `ngInit` runs, and that happens *after* the first call to `ngOnChanges`.

I note that none of the stock themeable components implements `onChanges` so this issue has not caused a problem until now, when I want to make `section-upload-file.component.ts` themeable.

**To Reproduce**
1. Write a themed replacement component that depends on change notification.
2. Notice that no changes notifications are received.

**Expected behavior**
All component lifecycle hooks should function in themes.

alexandrevryghem:
## References
* Fixes #2201

## Description
I looked over the code again and I saw that this issue was already partially fixed in PR https://github.com/DSpace/dspace-angular/pull/1512. The `ngOnChanges()` from the child component is already being called in the `ThemedComponent` but in some rare cases the `ngOnChanges()` is triggered before the themed child component is created by the themed wrapper. When the `ngOnChanges()` isn't called multiple times this can lead to the child component's `ngOnChanges()` not being triggered at all. To fix this the component will be created in the `ngOnChanges()` if it doesn't exist yet and the child's `ngOnChanges` will be manually triggered afterwards as well.

This is also the case for the components who are rendered using the custom decorators, the rendered component's `ngOnChanges` isn't called their either when changes are detected in the loader component.

## Instructions for Reviewers
List of changes in this PR:
- Made `ThumbnailComponent` themeable
- Modified `ThemedComponent.ngOnChanges()` to initiated the themed child component when it doesn't exists yet, because previously when it didn't exist it would just skip it which caused `ngOnChanges` sometimes from not being called.
- Fixed an issue where the `src` would be `undefined` when the thumbnail isn't defined (this throws a 404 error in the network tab because it tries to retrieve the url `http://localhost:4000/undefined`, this can be seen on demo if you go to the full item page of an item)
- Added missing `offset-3` class to `CommunitySearchResultListElementComponent` to fix listable communities in the search tab that aren't aligned on mobile search page for listable communities.
- Made changes to `ListableObjectComponentLoaderComponent`, `MetadataRepresentationLoaderComponent`, `ClaimedTaskActionsLoaderComponent` to fix the same problem. The `ngOnChanges` from the components with those decorators would not automatically be called anymore.

`ThumbnailComponent`
* Set the theme to `custom` theme in `config.yml`
* Uncomment the `CustomEagerThemeModule` in `src/themes/eager-themes.module.ts`
* Test if the `ThumbnailComponent` in the `custom` theme folder still works like before (on item pages and on the listable search components)
* Test by adding a debug point in the `ngOnChanges()` method or adding a `console.log()` and checking that it passed through the `ngOnChanges()` method

The custom components loaders:
* Find a component that is rendered by `ListableObjectComponentLoaderComponent`, `MetadataRepresentationLoaderComponent` and `ClaimedTaskActionsLoaderComponent` and add in each of those components a new `ngOnChanges` and check that they are called.
* Check that all the other components using this decorator still work as before

## Checklist


mwoodiupui:
I have not thoroughly reviewed this code, but I did copy the changes in `src/app/shared/theme-support/themed.component.ts` and alter my nonworking themed component as suggested.  Debug console output shows that `ngOnChanges` is now called as expected, and my custom component has begun working.

tdonohue (approved):
👍 Thanks @alexandrevryghem !  Tested this with the given instructions above (enabled custom theme, added a `console.log` in onChanges for thumbnails, verified it is called).  Also verified other modified components seem to be functioning the same as before.  Verified the "/undefined" 404 is fixed in the "Full Item page" as well.

So, this looks great and works well.

PR #2241, opened 2023-05-11
View mode switch accessibility

doana:
## Description
Use more semantically appropriate elements for list controls and markup that better aligns with WCAG recommendations.

## Instructions for Reviewers
This issue originally brought to my attention because our accessibility auditing tools flag the links as having empty link content. While in practice they read fine to me in my testing using the NVDA screen reader, other assistive technology might not infer that the icon title should be used as the link text.

The crux of this change is to use more semantically appropriate elements for the view mode switch controls. The use of button elements is more appropriate as they don't imply a significant change of context the way that a link element does. The other change here is to explicitly include the name for the control in the button element itself, instead of as part of the title attribute on the icon inside of the link. This better aligns with the WCAG since the name of the control is more explicitly available for programmatic access. 

List of changes in this PR:
1. Use buttons instead of links for the controls
2. Move the name of the control outside of the icon title attribute and into the body of the button element
3. Use span for the icon elements as the idiomatic text element doesn't make semantic sense for an icon
4. Update tests

An alternative strategy for #2 would be to set the title element on the button itself, rather than on the icon. This would keep the tool tip that is lost from this change. 

## Checklist

### DIFF
--- a/src/app/shared/view-mode-switch/view-mode-switch.component.html
+++ b/src/app/shared/view-mode-switch/view-mode-switch.component.html
@@ -1,5 +1,5 @@
 <div class="btn-group" data-toggle="buttons">
-    <a *ngIf="isToShow(viewModeEnum.ListElement)"
+    <button *ngIf="isToShow(viewModeEnum.ListElement)"
        routerLink="."
        [queryParams]="{view: 'list'}"
        queryParamsHandling="merge"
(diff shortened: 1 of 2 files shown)


tdonohue (approved):
👍 Thanks @doana for your contribution!  I was able to test this today and verified that there are no adverse side effects to the behavior of these buttons.  I agree completely that your fixes are more appropriate and improve accessibility.

PR #2271, opened 2023-05-24
ItemStatus component should handle missing config gracefully

kshepherd:
If the "register identifier" config is missing, simply return enabled=false rather than break the rest of the page features from loading. This is a tiny fix that would be great to squeeze into 7.6

## Description
The item status page looks for the `identifiers.item-status.register-doi` config property to determine whether the Register DOI feature should be included. If this property is not exposed by REST or the request fails in some other way, the remaining features would not load because the pipe would wait forever for a successful response.

This fix checks for first _completed_ REST response and then evaluates it properly, with any errors being handled gracefully and resulting in the Register DOI button being excluded from the feature list.

## Instructions for Reviewers
**before** applying this patch
1. Ensure `identifiers.item-status.register-doi` is exposed in rest.cfg and exists in identifiers.cfg, but is false
2. As an administrator, visit an item status page and note the buttons display normally
3. Comment out the config property from rest.cfg or identifiers.cfg and restart tomcat
4. As an administrator, visit an item status page and note the buttons do not finish loading, they are all disabled
5. Apply the PR and repeat the first 4 steps, now step 4 should result in a normal, successful item status page
6. You can also change the config property to true in identifiers.cfg to ensure that it does display the button as expected (note, it will only be enabled if that item is eligible for a DOI as per usual, but it should at least display a disabled button)

## Checklist

### DIFF
--- a/src/app/item-page/edit-item-page/item-status/item-status.component.ts
+++ b/src/app/item-page/edit-item-page/item-status/item-status.component.ts
@@ -9,9 +9,9 @@ import { RemoteData } from '../../../core/data/remote-data';
 import { getItemEditRoute, getItemPageRoute } from '../../item-page-routing-paths';
 import { AuthorizationDataService } from '../../../core/data/feature-authorization/authorization-data.service';
 import { FeatureID } from '../../../core/data/feature-authorization/feature-id';
-import { hasValue } from '../../../shared/empty.util';
+import { hasValue, isNotEmpty } from '../../../shared/empty.util';
 import {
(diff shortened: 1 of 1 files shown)


nwoodward (approved):
👍 Tested and confirmed this works. After applying the PR the buttons on the Status page were operable again.

tdonohue (approved):
👍 Thanks @kshepherd !  Agreed, this looks good.  I was also able to reproduce the bug easily & found this fixes it.

PR #2286, opened 2023-05-31
Remove scripts/webpack.js

### LINKED ISSUE #2285: code cleanup: removal of webpack.js
There was a discussion about the hard coded value of `max_old_space_size` in `webpack.js` in the angular-ui Slack channel.

@artlowel was mentioning (permalink https://dspace-org.slack.com/archives/C3TTSEB1V/p1685520313753129):

> That `webpack.js` script was used at the time to run webpack with more memory than the default of 2GB: [https://github.com/DSpace/dspace-angular/commit/c58e6f62395569e3dc33c7e3c670b15b801f8a90#diff-7ae45ad102eab3b6d7e7896[…]46470d7bc6507b6481575d519R45](https://github.com/DSpace/dspace-angular/commit/c58e6f62395569e3dc33c7e3c670b15b801f8a90#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R45)
>
> It is no longer used now.

Do you mean that `webpack.js` is no longer needed and can be removed from the code base?

> It looks like it yes

alanorth:
# References

* Fixes #2285

# Description
Remove unused `scripts/webpack.js`. After some discussion with @saschaszott on Slack we realized that this file is no longer needed.

# Instructions for Reviewers
Check whether the Angular frontend builds and runs.

List of changes in this PR:
* Remove `scripts/webpack.js`
* Remove mention of `webpack.js` from`README.md`


saschaszott:
@alanorth , Webpack seems to be still used in the project as there are several Typescript files in the `webpack` directory. Is there a dependency between `webpack.js` and the `ts` files under `./webpack`?

alanorth:
Hi @saschaszott, as I understand it we are using webpack, but not the `scripts/webpack.js` file itself. I managed to build and run the Angular UI in dev mode after removing `scripts/webpack.js`. Let's wait for the tests here to finish to see if they highlight any other unit or integration issues.

artlowel (approved):
Thanks @alanorth!

Issue #2291, opened 2023-06-05
EPerson deletion leads to application freeze: unlimited OPTIONS requests on /server/api/eperson/epersons/search/byMetadata?page=0&size=5&query=


saschaszott:
The application freezes when you submit the "Delete" button in the **modal dialog** that is shown if you click the "Delete EPerson" button in the "Edit EPerson form:

### Edit EPerson form with "Delete EPerson" button

![image](https://github.com/DSpace/dspace-angular/assets/952735/eb7c93e7-4824-49eb-9733-187ad8dbfbfb)

### Modal Dialog with "Delete" button

![image](https://github.com/DSpace/dspace-angular/assets/952735/baf49c7b-6458-4840-aa74-e034c3f79d87)

A click on the "Delete" button in the modal dialog results in an infinite loop that generates a massive amount of OPTIONS HTTP requests on the REST API endpoint URL `/server/api/eperson/epersons/search/byMetadata?page=0&size=5&query=`.

![image](https://github.com/DSpace/dspace-angular/assets/952735/50e8a24b-808b-4ab5-b847-44232254a0a1)

This behaviour is reproducible with **different browsers** (Chrome, Safari) and the **current version of DSpace (7.5)** on my local dev instance as well as on the official DSpace demo instance.

**Important observation**: The browser URL does not change if you enter the "Edit EPerson" form starting from the EPeople list page (/access-control/epeople).

PR #2293, opened 2023-06-06
Delete EPerson from "Edit EPerson" form 

alisaismailati:
## References
* Fixes [#2291](https://github.com/DSpace/dspace-angular/issues/2291)

## Description
There has been a refactor of the methods participating in EPerson delete action, including avoiding nested subscriptions and removing unnecessary events.

## Instructions for Reviewers
List of changes in this PR:
* First, refactor the `delete() `method of `EPersonFormComponent`, in order to avoid nested subscriptions and manage of disabling the delete button when the delete of the entity is confirmed (since the request takes a whole and the button can be clicked again). After the delete has finished successfully, we only need to notify the user for the status of the deletion, there is no need to trigger a submit of the form. So the `this.submitForm.emit()` is removed.
* Second, the` reset() `method of `EPeopleRegistryComponent` is refactored for the same reason, to avoid nested subscriptions.
* Another change would include removing the call of `onCancel()` method on destroy of `EPersonFormComponent`. This method is called also when going back from the "edit" form to the list and the logic this method contains is also used when needed in other methods, so there is no need for an extra execution of the logic inside `onCancel()` method.

Steps to reproduce the bug are the same as the ones mentioned in [#2291](https://github.com/DSpace/dspace-angular/issues/2291) .

## Checklist

### DIFF
--- a/src/app/access-control/epeople-registry/epeople-registry.component.ts
+++ b/src/app/access-control/epeople-registry/epeople-registry.component.ts
@@ -287,14 +287,17 @@ export class EPeopleRegistryComponent implements OnInit, OnDestroy {
   /**
    * This method will set everything to stale, which will cause the lists on this page to update.
    */
-  reset() {
+  reset(): void {
     this.epersonService.getBrowseEndpoint().pipe(
-      take(1)
-    ).subscribe((href: string) => {
-      this.requestService.setStaleByHrefSubstring(href).pipe(take(1)).subscribe(() => {
-        this.epersonService.cancelEditEPerson();
-        this.isEPersonFormShown = false;
-      });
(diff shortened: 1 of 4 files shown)


tdonohue (approved):
👍 Thanks @alisaismailati !  This fixes the bug.  Verified also that it doesn't impact any of the other buttons on the same page.

saschaszott:
@tdonohue , ok, I will open a new issue for this "bug". I was wondering why the two implementations (epeople and group administration) do not have the same base components, e.g. a common list component.

saschaszott:
👉 https://github.com/DSpace/dspace-angular/issues/2302

PR #2318, opened 2023-06-16
resolve German translation inconsistency in scopenote

saschaszott:
## Description

This PR resolves a minor bug in `de.json5` in which two different translations of `scopenote` can be found. As it is not clear which translation is preferable I suggest a combined translation including both terms.


tdonohue (approved):
👍 Thanks @saschaszott .  This seems fine to me.  Merging immediately

PR #2338, opened 2023-06-23
Made ConfirmationModalComponent more generic

### LINKED ISSUE #2264: Make ConfirmationModalComponent more generic
Currently the `ConfirmationModalComponent` requires [a DSpaceObject](https://github.com/DSpace/dspace-angular/blob/383485b16ddbcb8e534862d0432c29998477f19f/src/app/shared/confirmation-modal/confirmation-modal.component.ts#L20) to be provided, so that it can add that name to the [labels](https://github.com/DSpace/dspace-angular/blob/740228695477889320e161b49626149e20a43fe2/src/app/shared/confirmation-modal/confirmation-modal.component.html#L2). It would be better to just pass the name to the item instead to make it possible to use that modal for other objects who don't extend `DSpaceObject`.

alexandrevryghem:
## References
* Fixes #2264

## Description
Currently the `ConfirmationModalComponent` uses the name from the `DSpaceObject`s to render the labels. The problem with this approach is that this modal only works for DSpace object and if you would want to use it for `MetadataField` (or any other child of `ListableObject` which isn't a child of `DSpaceObject`) it wouldn't work. Therefore it's better to only give the name to the modal instead of the whole `DSpaceObject` object.

## Instructions for Reviewers
To make this component more generic I replaced the `@Input() dso: DSpaceObject;` with `@Input() name: string` and replaced all the usages

- Open all the modals that have been changed and check that the names are still displayed correctly

## Checklist


tdonohue (approved):
👍 Thanks @alexandrevryghem !  Looks good and makes sense.  I tested all the confirmation modals/popups that are modified and they all appear to work the same as before.

PR #2344, opened 2023-06-29
Catch and handle unsuccessful "convert rels to items" responses

kshepherd:
## Description
If one or more related items in a list (eg. left items displayed on their corresponding right item's page) cannot be retrieved for some reason (for instance, authZ policies mean a 401 response for some people, or some other issue prevents a successful 200 response with good payload), the function that converts the relations to items will not continue to process the rest of the list.

This pull request might need more work - it is my first attempt at mitigating the problems / assumptions introduced by `getFirstSucceededRemoteDataPayload`. There may be more elegant ways of handling this, and we should write some new spec tests to prove the old way fails and the new way doesn't.

## Instructions for Reviewers
* Have 2-5 relationships between items eg. author and publication or project and publication
* Make sure each item component tries to display a related items list
* Restrict one of the items so that an Anonymous user gets a 401 when requesting it
* Visit the item page for the item on the *other* side of the relationship as a not-logged-in user
  * Before this PR was applied, the list would not display at all. After this PR is applied, the list displays the successfully requested related items.

## Checklist

### DIFF
--- a/src/app/item-page/simple/item-types/shared/item-relationships-utils.ts
+++ b/src/app/item-page/simple/item-types/shared/item-relationships-utils.ts
@@ -5,8 +5,7 @@ import { RemoteData } from '../../../../core/data/remote-data';
 import { Relationship } from '../../../../core/shared/item-relationships/relationship.model';
 import { Item } from '../../../../core/shared/item.model';
 import {
-  getFirstSucceededRemoteDataPayload,
-  getFirstSucceededRemoteData
+  getFirstCompletedRemoteData
 } from '../../../../core/shared/operators';
 import { hasValue } from '../../../../shared/empty.util';
 import { InjectionToken } from '@angular/core';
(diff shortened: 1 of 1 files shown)


tdonohue (approved):
@kshepherd : Thanks for the fix!  I've reviewed the code and it looks good.  Also tested it and verified it's working.

I was able to reproduce the original bug easily in the demo entities data set via the following:
1. First, I started with a Project with relationships to several Publications.  I chose this Project from the demo entities data set which is related to 5 Publications: http://localhost:4000/entities/project/18e7924c-f15b-4953-9fe3-3de370bccc97
2. Then, I access restricted *one* of the related Publications by setting all its permissions to Admin-only.
3. Finally, as an anonymous user, I went back to the Project page from step one. **None of the Publications are listed anymore as related.**

When I go through the same steps with this PR applied, in the last step (3), only the access restricted Publication is missing from the list (which is as expected).  The other four are now displayed.

PR #2396, opened 2023-07-25
[Port dspace-7_x] Search query is not set by default when opening an external sources tab

alexandrevryghem:
- Port of #2376 to `dspace-7_x`.


tdonohue (approved):
👍 Tested/verified on `main` branch in #2376.  This PR is identical for `dspace-7_x`.

PR #2404, opened 2023-07-27
[Port dspace-7_x] Fix `setStaleByHrefSubstring` not emitting true when all requests are stale

alexandrevryghem:
- Port of #2349 to `dspace-7_x`.


tdonohue (approved):
👍 Reviewed/tested against `main` in #2349.  This PR is identical and simply applies the changes to `dspace-7_x`.

PR #2412, opened 2023-08-02
remove redundant cache default values from server.ts

saschaszott:
Default values of both `botCache` and `anonymousCache` are defined in `default-app-config.ts`. 

This PR removes redundant default values from `server.ts`.


tdonohue (approved):
👍 Thanks @saschaszott .  Verified this looks good.  These are defined by default in `default-app-config.ts`, which means this PR is "safe".  It is not necessary to also have a fallback in `server.ts`.

PR #2422, opened 2023-08-08
Update fi.json5

reetagithub:
Two last translations to the Finnish file.

## Description
Two translations were still to do in the current version of the Finnish file.


alanorth:
Thanks, @reetagithub. I tested this patch by building the Angular frontend and switching to Suomi. I wasn't able to find the exact locations where these strings were used in the user interface, but as far as I can tell the strings are correct and don't break anything else.

alanorth (approved):

PR #2475, opened 2023-09-05
Fix autodiscovery Opensearch tag

J4bbi:
## References
* Fixes https://github.com/DSpace/dspace-angular/issues/791
* Requires https://github.com/DSpace/DSpace/pull/9058

## Description
The RSS component injects link tags into the document header. One of them is for autodiscovery and is incorrectly formatted (as explained in the issue).

This PR

- changes the value for the `href` and `type` attributes of the autodiscovery tag (lines [97-98](https://github.com/DSpace/dspace-angular/blob/191c0399b94c82c9a026ebb06da8cc5d500adf9d/src/app/shared/rss-feed/rss.component.ts#L97-L98))
- reformats the code to improve readability
- makes minor changes to documentation

## Instructions for Reviewers
You need to add two variables to the whitelist of exposed variables in the REST API. See DSpace [PR 9058](https://github.com/DSpace/DSpace/pull/9058).
Open up the main landing page and inspect the html.

Where it said:
`<link href="https://sandbox.dspace.org/server/opensearch/search/service" type="application/atom+xml" rel="search" title="Dspace">`

it will now read:
`<link href="https://sandbox.dspace.org/server/opensearch/service" type="application/opensearchdescription+xml" rel="search" title="Dspace">`

For consideration:

- According to the [spec](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#autodiscovery), the value for the `title` attribute, which I have left unchanged is currently "Dspace" (and should probably by DSpace, with a capital S?) but " may contain a human-readable plain text string describing the search engine." It would be possible to fetch a value from config such as `dspace.name`.
- I've hardcoded the URL to be `environment.rest.baseUrl + '/opensearch/service'`, I'm not confident this should not be configurable under [Opensearch configuration settings ](https://github.com/DSpace/DSpace/blob/c7c30227ec4a4e5e2d4eb44ee78e04a1035a95b5/dspace/config/dspace.cfg#L1331-L1359)

### DIFF
--- a/src/app/shared/rss-feed/rss.component.ts
+++ b/src/app/shared/rss-feed/rss.component.ts
@@ -5,24 +5,16 @@ import {
   OnInit,
   ViewEncapsulation
 } from '@angular/core';
-import {BehaviorSubject, Subscription} from 'rxjs';
+import {BehaviorSubject, combineLatest as observableCombineLatest, Subscription} from 'rxjs';
 import { GroupDataService } from '../../core/eperson/group-data.service';
 import { LinkHeadService } from '../../core/services/link-head.service';
 import { ConfigurationDataService } from '../../core/data/configuration-data.service';
-import {
-  getFirstCompletedRemoteData,
-  getFirstSucceededRemoteData,
-  getFirstSucceededRemoteDataPayload,
-} from '../../core/shared/operators';
+import { getFirstSucceededRemoteDataPayload } from '../../core/shared/operators';
 import { environment } from '../../../environments/environment';
 import { SearchConfigurationService } from '../../core/shared/search/search-configuration.service';
 import { PaginationService } from '../../core/pagination/pagination.service';
 import { Router } from '@angular/router';
-import {map, switchMap} from 'rxjs/operators';
-import { PaginatedSearchOptions } from '../search/models/paginated-search-options.model';
-import { RemoteData } from '../../core/data/remote-data';
-import { ConfigurationProperty } from '../../core/shared/configuration-property.model';
-
+import { map } from 'rxjs/operators';
 
 /**
  * The RSS feed button component.
@@ -44,9 +36,6 @@ import { ConfigurationProperty } from '../../core/shared/configuration-property.
 export class RSSComponent implements OnInit, OnDestroy  {
   uuid: string;
   subs: Subscription[] = [];
-  // Give default value as DSpace
-  shortname: BehaviorSubject<string> = new BehaviorSubject<string>('DSpace');
-  autolink: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(true);
   isEnabled$: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(null);
   route$: BehaviorSubject<string> = new BehaviorSubject<string>('');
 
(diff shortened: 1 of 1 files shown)


alanorth:
Ah, good catch @J4bbi! Before, my site was adding the incorrect link:

```
<link href="http://localhost:8080/server/opensearch/search/service" type="application/atom+xml" rel="search" title="Dspace">
```

After, the link is correct:

```
<link href="http://localhost:8080/server/opensearch/service" type="application/opensearchdescription+xml" rel="search" title="Dspace">
```

And I was able to see that Firefox now detects the OpenSearch endpoint and offers to add my repository as a search engine when I type in the URL bar.

> _Having another look at the config._
> _the title attribute should probably be set as [websvc.opensearch.shortname](https://github.com/DSpace/DSpace/blob/c7c30227ec4a4e5e2d4eb44ee78e04a1035a95b5/dspace/config/dspace.cfg#L1346C1-L1346C28)_

I agree. In fact, looking at my DSpace 6 instance, I see that the OpenSearch functionality sets both the short and long names.

J4bbi:
I'm converting this to draft as it is apparent that a parallel PR for the backend is required to expose more opensearch configuration variables.

There are also other issues I'll go into in more detail when I have time tomorrow or later in the week.

J4bbi:
I think[ the configuration rest endpoint](https://github.com/DSpace/RestContract/blob/main/configuration.md)  is inefficient.

It was created for a single use case, [back in 2020](https://github.com/DSpace/RestContract/pull/119). To pass the Google Analytics key to the front end from the backend. By now there are [17 backend configuration variables exposed via the rest end point](https://github.com/DSpace/DSpace/blob/c7c30227ec4a4e5e2d4eb44ee78e04a1035a95b5/dspace/config/modules/rest.cfg#L32-L54)

- You can only get one variable per request at the moment, the [findAll is not implemented](https://github.com/atmire/DSpace/blob/a7aeb102d4d714c26d9d4a138088cc40f1df1bef/dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/ConfigurationRestRepository.java#L71-L73). You have to make a separate call for each config you want, this creates overhead that puts the webserver to work. I've been looking at the RSS component, which is embedded by the way in every single Pagination component, it makes two requests for metadata properties every single time it's loaded, and will likely need to make three or more if the current approach is used.
- Another issue is that this call for the Google Analytics key is being made on every single load of a page because the front end does not know if GA is enabled and needs to ask the backend every single time.

It's not surprising that the new frontend is resource intensive if it is making several requests per page load

**Example:**
https://sandbox.dspace.org/browse/dateissued, half a sec wasted on GA. There is also another request for a metadata property there, registration.verification.enabled taking 185 ms and 1.1 kb, bandwidth and resources

image

**Solution**: have all config in a bundle and check there, no need for separate API call resulting in console error. It could possibly be exposed via /server/api endpoint

it already has:
```json
{
  "dspaceUI" : "https://sandbox.dspace.org/",
  "dspaceName" : "DSpace Sandbox",
  "dspaceServer" : "https://sandbox.dspace.org/server",
  "dspaceVersion" : "DSpace 8.0-SNAPSHOT",
  "type" : "root",
}
```
why not add
```json
{
  "dspaceUI" : "https://sandbox.dspace.org/",
  "dspaceName" : "DSpace Sandbox",
  "dspaceServer" : "https://sandbox.dspace.org/server",
  "dspaceVersion" : "DSpace 8.0-SNAPSHOT",
  "type" : "root",
  "configuration" : [
    {
       "name": "google.analytics.key",
       "values": [ 
           "UA-XXXXXX-X"
      ]
   }]
}
```

If the config is delivered via one call, you would also get rid of the 404 error in the console if GA key is not defined,
this 404 will appear for any repo that doesn't have GA configured

alanorth:
Thanks @J4bbi. I've applied this patch along with https://github.com/DSpace/DSpace/pull/9058, but I still see "DSpace" instead of my site's name in the OpenSearch head tags.

I have verified that the backend correctly exposes `websvc.opensearch.shortname` with my site name. I tried shift refreshing...

J4bbi:
I've now gone away from trying to set variables in the component's outer scope from an inner scope ca22491

Over to using `combineLatest`, which if I understand correctly combines getting four configuration properties along with pagination search options, to action on all of them at once.

All of the action therefore happens within the scope of that combined subscription as I understand it ([L64-100](https://github.com/DSpace/dspace-angular/blob/e9408d3eaf2797aba4b50f2386bf8d3774c9131e/src/app/shared/rss-feed/rss.component.ts#L64-L100)).

I am not certain I am now doing things the right way, now that `switchMap` is not being used. 

I readily admit I don't have a good understanding of how Angular works at this level.

Changes to code behaviour from the earlier version are:

- code stops executing if `websvc.opensearch.enable` is not "true" ([L81-83](https://github.com/DSpace/dspace-angular/blob/e9408d3eaf2797aba4b50f2386bf8d3774c9131e/src/app/shared/rss-feed/rss.component.ts#L81-L83))
- code now respect `websvc.opensearch.autolink `, as to whether a link tag is inserted for autodiscovery
- code now respects `websvc.opensearch.shortname` to be the description of the service for autodiscovery

Potential problems
- is the code resilient for low or no latency, what if these configuration settings are commented out completely?
- as I do not fully understand switchMap functionality I can not say if the code works as before
- I don't fully understand the ondestroy code ([L52-57](https://github.com/DSpace/dspace-angular/blob/e9408d3eaf2797aba4b50f2386bf8d3774c9131e/src/app/shared/rss-feed/rss.component.ts#L52-L57)). **A** when does that happen? Presumably the RSS component will not be toggled on and off? **B** is it necessary to unsubscribe from all subscriptions? my understanding is that some only live to deliver a single value and then self-unsubscribe? **C** the code indiscriminately removes all tags with `rel='alternate'`, is that definitely safe? If it is cleaning up after itself, this component ejects 2 or 3 link tags into the header, two of which will have `rel='alternate'` but the optional third, the autodiscovery tag, has `rel='search'`.

artlowel:
>- is the code resilient for low or no latency, what if these configuration settings are commented out completely?

I'd urge you to write tests for those cases in the spec file. If you use `observableOf` to mock the output of `findByPropertyName` (or more likely `createSuccessfulRemoteDataObject$` since you need a `RemoteData`, but that uses `observableOf` internally) it will emit without latency, so you can verify what would happen.

>as I do not fully understand switchMap functionality I can not say if the code works as before

A `switchMap` maps the source to another observable. Every time the source observable changes, it will unsubscribe from the previous mapped observable, and subscribe to the new one.

In this example `openSearchUri` comes from the source observable, and is used to do another call that results in an observable:  that `this.searchConfigurationService.paginatedSearchOption` call. So if the source observable were to emit an updated `openSearchUri` value, `switchMap` would do a new call using that value, and stop emitting responses from the call using the old URI. That's as a opposed to `mergeMap`, which _would_ simply keep emitting all values from all previous calls combined. In this case the difference doesn't matter as the source observable will only emit once anyway. But in most cases you're only interested in the most recent data, so you need a `switchMap` when turning one observable into another. That's probably why the original dev chose it.


> I don't fully understand the ondestroy code ([L52-57](https://github.com/DSpace/dspace-angular/blob/e9408d3eaf2797aba4b50f2386bf8d3774c9131e/src/app/shared/rss-feed/rss.component.ts#L52-L57)). **A** when does that happen? Presumably the RSS component will not be toggled on and off? 

`ngOnDestroy` is the cleanup method for a component. It will be executed whenever that component stops being rendered, so in this case probably when you navigate from a page that has an RSS component, to a page that doesn't

> **B** is it necessary to unsubscribe from all subscriptions? my understanding is that some only live to deliver a single value and then self-unsubscribe? 

Strictly necessary no, but it's good practice. Because even if an observable will only emit once and then complete, if for some reason it never emits, the subscription remains open, in browser memory, indefinitely 

> **C** the code indiscriminately removes all tags with `rel='alternate'`, is that definitely safe? If it is cleaning up after itself, this component ejects 2 or 3 link tags into the header, two of which will have `rel='alternate'` but the optional third, the autodiscovery tag, has `rel='search'`.

No, that's not safe. The better solution would be to give the tags added by this component a unique class, that way we can remove only the ones with that class at the end. This is [the way ThemeService does it](https://github.com/DSpace/dspace-angular/blob/e9408d3eaf2797aba4b50f2386bf8d3774c9131e/src/app/shared/theme-support/theme.service.ts#L214-L217)

PR #2505, opened 2023-09-19
remove obsolete label element in metadata-schema.component.html

saschaszott:
This PR implements a minor HTML change which improves the "find in page" of your prefered web browser. Currently, it is not possible to search for metadata field names of the form `schema.element.qualifier` using the "find in page" function, e.g. in /admin/registries/metadata/dc.

### DIFF
--- a/src/app/admin/admin-registries/metadata-schema/metadata-schema.component.html
+++ b/src/app/admin/admin-registries/metadata-schema/metadata-schema.component.html
@@ -41,7 +41,7 @@
                   </label>
                 </td>
                 <td class="selectable-row" (click)="editField(field)">{{field.id}}</td>
-                <td class="selectable-row" (click)="editField(field)">{{schema?.prefix}}.{{field.element}}{{field.qualifier ? '.' + field.qualifier : ''}}</td>                
+                <td class="selectable-row" (click)="editField(field)">{{schema?.prefix}}.{{field.element}}{{field.qualifier ? '.' + field.qualifier : ''}}</td>
                 <td class="selectable-row" (click)="editField(field)">{{field.scopeNote}}</td>
               </tr>
               </tbody>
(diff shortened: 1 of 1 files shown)


alanorth:
Thanks @saschaszott. If I understand correctly, by "find in page" you mean the Ctrl-F / Cmd-F in the browser? I tried searching for `dc.contr` using Ctrl-F in Firefox on DSpace 7.6 and it works.

<p align="center">
  
</p>

Did I understand correctly?

saschaszott:
@alanorth , the problem only arises if you search with a field name that contains **two** periods.

This screenshot was taken from Google Chrome where a search with term `dc.contributor.advisor` does not match the page content.

image

alanorth:
OK, I don't use Chrome so I don't know. I can _find_ a metadata field with two dots in Firefox.

<p align="center">
  
</p>

Is this because it's wrapping for some reason in Chrome, or perhaps just on your device?

saschaszott:
@alanorth , I cannot reproduce this odd behaviour in Safari. This seems to be a **Chrome** specific problem. Nevertheless, it should be fixed as Chrome has > 50% market coverage.

alanorth (changes requested):
This introduces trailing whitespace.

alanorth:
@saschaszott I finally looked at this. The existing code doesn't make sense—why put a label with only part of the metadata field? Could have been a bug in the original code.

Anyways, I tested this and it doesn't break my find in Firefox so I'm +1. Would you feel comfortable rebasing this commit to remove the trailing whitespace? Thanks.

saschaszott:
> This introduces trailing whitespace.

Thank you for noticing - it was fixed.

alanorth (approved):

alanorth:
Thanks @saschaszott. I will squash the two commits into one before merging so the whitespace doesn't stay in git history.

PR #2506, opened 2023-09-19
allow insertion of multi-line scope notes in MD field registry

saschaszott:
## Description

This PR adds a small improvement to the "create metadata field" form by changing the type of the scope form field to textarea. By this change the insertion of multi-line scope notes is better supported.

image


alanorth:
I tested the patch and it works as described. When I entered a multi-line note for a metadata field it was then shown with "\n" when I viewed the metadata field via the REST API, ie `/api/core/metadatafields/<:id>`. I don't know if that has ramifications we need to consider?

tdonohue (approved):
👍 Thanks @saschaszott !  Works as described.  I'm not noticing any side effects to this change, so I think it's safe.

PR #2508, opened 2023-09-20
Serbian (Latin) translation

imilos:
## References
* Replaces #2507

## Description
Serbian (Latin) translation added.

## Instructions for Reviewers

List of changes in this PR:
* Added file sr-lat.json5
* Modified default default config to include Serbian (Latin) translation.


alanorth:
Thanks @imilos! I tested this patch and was able to switch to Serbian latin in the Angular user interface.

I notice that the order of languages in the drop down mostly alphabetical (at least for Latin script). Can you move `Srpski (lat)` up, above `Suomi`? Thanks.

imilos:
Thanks for your quick response @alanorth ! I moved Serbian (lat) just above Suomi, as you suggested. I also hope to have Serbian with Cyrillic script soon.

alanorth (approved):

alanorth:
@imilos Hvala! Looks good. Merging now (I squashed the commits into one).

PR #2527, opened 2023-09-26
Minor header button improvements

alexandrevryghem:
## References
- Partially resolves larger ticket #1174
  - one of these was probably for the `ExpandableNavbarSectionComponent`: 468407, 468408, 468409, 468410
  - Issue ID: 468472

## Description
Minor header improvements for header icons spacing (by refactoring the code to use flex gap) & improved search navbar bar accessibility.

image

## Instructions for Reviewers
List of changes in this PR:
* Fixed a html issue: the `ExpandableNavbarSectionComponent` has a `<ul>` containing non-`<li>` elements ([here](https://github.com/DSpace/dspace-angular/blob/7caacdb20aa465bb0b60d5789831bb910faa6b60/src/app/navbar/expandable-navbar-section/expandable-navbar-section.component.html#L15))
* Added flex gap to the navbar button container
* Removed the individual paddings on all the buttons
* Improved the search navbar by preventing it to focus on the input when it's not expanded (Deque issue ID: 468472)
* Themed the `LangSwitchComponent` for testing purposes in order to check that when the language button is hidden that there is no double padding being applied & that it works for themed components as well

**Include guidance for how to test or review your PR:**
* Test this by just looking at the header icons in both mobile & desktop mode
* Test it with the `dspace` theme & the base theme (done by commenting out the `dspace` theme in `src/config/default-app-config.ts`)

## Checklist

### DIFF
--- a/src/app/admin/admin-sidebar/admin-sidebar-section/admin-sidebar-section.component.ts
+++ b/src/app/admin/admin-sidebar/admin-sidebar-section/admin-sidebar-section.component.ts
@@ -12,8 +12,7 @@ import { Router } from '@angular/router';
  * Represents a non-expandable section in the admin sidebar
  */
 @Component({
-  /* eslint-disable @angular-eslint/component-selector */
-  selector: 'li[ds-admin-sidebar-section]',
+  selector: 'ds-admin-sidebar-section',
   templateUrl: './admin-sidebar-section.component.html',
   styleUrls: ['./admin-sidebar-section.component.scss'],
 
--- a/src/app/admin/admin-sidebar/admin-sidebar.component.html
+++ b/src/app/admin/admin-sidebar/admin-sidebar.component.html
@@ -26,10 +26,10 @@
               </div>
             </li>
 
-            <ng-container *ngFor="let section of (sections | async)">
+            <li *ngFor="let section of (sections | async)">
                 <ng-container
                         *ngComponentOutlet="(sectionMap$ | async).get(section.id).component; injector: (sectionMap$ | async).get(section.id).injector;"></ng-container>
-            </ng-container>
+            </li>
         </ul>
     </div>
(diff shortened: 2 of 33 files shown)


alanorth:
Thanks @alexandrevryghem. Tested by applying to my local DSpace 7.6 development environment and updating my own custom theme (based on the `dspace` theme) to use the new classes and styles.  The buttons in the navbar look and behave exactly as before, but in effect have slightly more padding:

diff-fs8

If we merge this we need to make sure people know that they will most likely need to update their custom themes.

alexandrevryghem:
Isn't that always recommended when upgrading your angular version (an alternative is to not just copy the original css file but to extend it instead, this way minor updates like this will always be included). I did however recalculate the pixels today and found out that the search icon in the search bar is also not correctly vertically aligned (see picture I added a background color to make it more visible) so I updated that and I also increased the gap width to match the previous spacing a little bit more.

alanorth:
> (an alternative is to not just copy the original css file but to extend it instead, this way minor updates like this will always be included).

Yeah that's a good idea: instead of copying the SCSS with the component to a theme, importing it instead and overriding the only styles you need. Less duplicated code that goes out of date...

Anyways, I'm +1 on this by testing now. There are tiny differences in the spacing of the buttons, but the functionality is the same. If this addresses the Dequeue accessibility feedback then I'm in support of merging it.

alexandrevryghem:
@tdonohue I indeed forgot to check if the `dspace` theme still behaved correctly, I fixed the issue so hopefully the CICD should pass now

tdonohue (approved):
👍 Thanks @alexandrevryghem !  Finally got to testing/reviewing this and it all looks good.  I can verify that it resolves several accessibility issues in the header buttons and works well.

PR #2551, opened 2023-10-19
[Port main] Created separate ePerson pages for create/edit

alexandrevryghem:
Port of #2391 to `main`


tdonohue (approved):
👍 Verified this is a manual port of #2391 (which was already approved/tested).  Code is identical.  I restarted the tests though as they showed (seemingly random) failures.  Once they succeed, I'll merge both PRs.

PR #2561, opened 2023-10-21
Fix RequestService test sometimes failing in CICD

alexandrevryghem:
## Description
Because tests are executed slower on CICD it can sometimes take one second or more to execute the `RequestService#setStaleByHref(href)`, when the test arrives inside the subscribe, a create new `RequestStaleAction` with the current time is created (this can already be a few seconds after `RequestService#setStaleByHref(href)` was called). To prevent such errors we can just expect that `lastUpdated` will be a number instead of comparing the actual values. The purpose of that test is to check the type of `RequestUpdateAction` returned by the function and not the date when it was created, so this doesn't have any side effects. [This is an example of such a failure.](https://github.com/DSpace/dspace-angular/actions/runs/6580387915/job/17878258713)

## Checklist

### DIFF
--- a/src/app/core/data/request.service.spec.ts
+++ b/src/app/core/data/request.service.spec.ts
@@ -661,7 +661,9 @@ describe('RequestService', () => {
       spyOn(service, 'getByHref').and.returnValue(observableOf(staleRE));
       spyOn(store, 'dispatch');
       service.setStaleByHref(href).subscribe(() => {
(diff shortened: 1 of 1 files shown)


tdonohue (approved):
👍 Thanks @alexandrevryghem for tracking this down, as I've seen this random failure as well.

This seems like a reasonable fix, so I'll merge it immediately as it only impacts the specs.

PR #2579, opened 2023-10-26
adding new access-status-list-element-badge css classes

### LINKED ISSUE #2402: Visual differentiation needed for Access Status
**Describe the bug**
Currently, when displaying different access status like open access, restricted access, ... there is no visual differentiation.
image

all look the same. 

**To Reproduce**
1. Firstly you should have content with different types of access (specific resource policies for open access, embargoed access)
2. Activate the required configuration on your client site (angular):
``` yaml
# Item Config
item:
  showAccessStatuses: true
```
3. Perform a query (can be an empty one and you will have a list of items). Please mind that this only applies to generic Items or Publications.

**Expected behavior**
I hoped to have a visual differentiation for different access status. Could be something based on a different color, a specific style class. Of course I can theme that component and do that on my side, but I think it makes more sense to have it like that for all implementations.

paulo-graca:
## References
* Fixes #2402

## Description
This PR adds CSS classes to access-status-badge component in order to make possible to differentiate access status based on the value itself.

## Instructions for Reviewers
In order to test it you need to activate in your angular configuration:
``` yaml
# Item Config
item:
  showAccessStatuses: true
```
and for Items and Publications (if you use Entities), inspect the access status value, like: "Open Access" and you should see something like:
``` html
<span class="badge badge-secondary access-status-list-element-badge access-status-open-access-listelement-badge">Open Access</span>
```

List of changes in this PR:
* We just changed the HTML part of the component and replace "." with "-" of the accessStatus value.


artlowel (changes requested):
Thanks @paulo-graca 

It works, but I just have an inline comment about the performance.

It would also be good to already provide default colors for the access statuses. Note that you can reuse bootstrap's own badge mixin for that e.g.:

```scss
.access-status-open-access-listelement-badge {
  @include badge-variant(map-get($theme-colors, warning));
}
```

this would make the `access-status-open-access-listelement-badge` class idential to the `badge-warning` class (i.e. orange)

src/app/shared/object-collection/shared/badges/access-status-badge/access-status-badge.component.html
```diff
@@ -1,5 +1,5 @@
 <ng-container *ngIf="showAccessStatus">
   <span *ngIf="accessStatus$ | async as accessStatus">
-    <span class="badge badge-secondary">{{ accessStatus | translate }}</span>
+    <span [class]="'badge badge-secondary access-status-list-element-badge ' + accessStatus.replaceAll('.','-')">{{ accessStatus | translate }}</span>
```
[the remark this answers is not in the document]
paulo-graca: Thank you @artlowel we weren't aware of this re-render issue. We will do as you suggest. Thank you for your feedback!

paulo-graca:
@artlowel I think I changed it in the way you suggested. I've also added an empty scss file, but without specifying anything. I think we all don't have a clear idea what access status should look like and you can confirm that on the issue discussion page (https://github.com/DSpace/dspace-angular/issues/2402). 
We are currently working in a custom theme that will address this, when finished, if all agree, we can proposed it to be the default. It will be something like:
image
and 
image

src/app/shared/object-collection/shared/badges/access-status-badge/access-status-badge.component.ts
```diff
@@ -57,5 +63,12 @@ export class AccessStatusBadgeComponent {
       map((status: string) => `access-status.${status.toLowerCase()}.listelement.badge`),
       catchError(() => observableOf('access-status.unknown.listelement.badge'))
     );
+
+    // stylesheet based on the access status value
+    this.accessStatus$.pipe(
+      map((accessStatusClass: string) => accessStatusClass.replace(/\./g, '-'))
+    ).subscribe((accessStatusClass: string) => {
+      this.accessStatusClass = accessStatusClass;
+    });
```
artlowel, line 72: Any time you add a subscription, please [store it in a component property](https://github.com/DSpace/dspace-angular/blob/8bf4ba812614e5ed34b2808b79d7e364fd01c2db/src/app/dso-shared/dso-edit-metadata/dso-edit-metadata.component.ts#L122) and then [unsubscribe from it, in the ngOnDestroy method of the component](https://github.com/DSpace/dspace-angular/blob/8bf4ba812614e5ed34b2808b79d7e364fd01c2db/src/app/dso-shared/dso-edit-metadata/dso-edit-metadata.component.ts#L255-L259). That means that we'll unsubscribe whenever that component stops being rendered.

If you don't do that these subscriptions will remain in memory indefinitely, and especially in a case like this, where a new subscription is opened for every search result, that could lead to problems

paulo-graca: I've added the onDestroy to deal with the subscription.

artlowel (approved):
Thanks @paulo-graca!

artlowel:
>We are currently working in a custom theme that will address this, when finished, if all agree, we can proposed it to be the default. It will be something like:
image
and 
image

I like the way that looks

PR #2590, opened 2023-10-29
fix(pt-BR.json5): fix and update the language file

marcoaureliocardoso:
Fix and update the pt-BR language file

## References
None

## Description
Fix and update the pt-BR language file

## List of changes in this PR:
* Many to list, but only in dspace-angular/src/assets/i18n/pt-BR.json5 file.


tdonohue (approved):
👍 Thanks @marcoaureliocardoso !  I gave this a quick test locally & verified some of the translations via Google Translate.  Looks good to me.

PR #2694, opened 2023-12-06
 [Port dspace-7_x]  Enable type-bind for checkbox inputs during submission

hutattedonmyarm:
Backport to 7.x of #2693


tdonohue (approved):
👍 Tested via `main` version #2693.  This is the same code so it looks good.  Thanks again @hutattedonmyarm !

PR #2736, opened 2024-01-10
use localized error message

saschaszott:
## Description

This PR removes the hard coded error message in `AuthService` that is displayed when a user supplies invalid username and/or password.

Please note, that `AuthService` still contains two other hard coded error messages ("Not authenticated").


nwoodward (approved):
👍 looks good after deploying. It would be nice to also localize the "not authenticated" messages you mentioned.

tdonohue (approved):
👍 Thanks @saschaszott !  This seems like a very obvious win.  Merging immediately.

PR #1732, opened 2022-07-20
Duplicate Detection ported from DSpace CRIS 7 Angular

kshepherd:
This pull request ports the Duplicate Detection framework from DSpace CRIS 7 into DSpace.

*Attribution: Ported from DSpace-CRIS and extended by The Library Code with support of Technische Universität Berlin*

DSpace Backend PR: https://github.com/DSpace/DSpace/pull/8415
REST Contract PR: TBD

**One outstanding issue**
_For a new submission, the default behaviour when no matches are found should be to keep the section hidden, but instead, the default message indicates that decisions are all registered._
Fixed, see comment https://github.com/DSpace/dspace-angular/pull/1732#issuecomment-1211459672

**To test**
- Make sure both the dspace-angular and DSpace PRs are applied
- Enable the ‘dedup’ event consumer in dspace.cfg and add the ‘detect-duplicate’ step to a submission process in item-submission.xml
- For best results, configure autosave on the dc.title field and any others you’re planning to test (eg. dc.identifier.doi, dc.identifier.isbn or so on). This means that whenever a change is saved to dc.title, the duplicate detection section will update with a new list of potential matches
- Submit an item to a collection with the detect-duplicate step enabled. Enter some text in the title so that 

**Notes for PR reviewers**
- There is a new ‘dedup’ table in the database, this should be added by flyway
- There is a new ‘dedup’ Solr core
- The ‘dedup’ consumer is not in the default event consumers by default, so it’ll need to be added for practical testing
- The fuzzy search signature is commented out of the deduplication.xml spring config by default, so uncomment that to test fuzzy matches (the default MD5 signatures for exact metadata matches still work)

**Notes for 4Science devs (regarding changes from CRIS)**
- SearchDeduplication was renamed to DeduplicationPluginService as that seemed more fitting to what it actually does
- tmpMapFilter, tmpFilter, other similar names of maps and lists were renamed to suit their purpose (eg. signatureSearchFilters)
- A lot of unused methods, variables, classes were stripped out during the port to keep code lean and easier to review. If some were partial implementations of future work, we can bring them in with a subsequent PR

As I’ve added an authorization check (configurable, enabled by default) so that the user needs read access to each item in the potential matches list, I manually disabled it for some integration tests and leave it for others, and explicitly test it in a new integration test

I’ve added javadoc and most is pretty straightforward but please correct me if I misunderstood terms such as ‘fake', ‘reader’

## Checklist
[...]

PR #2749, opened 2024-01-18
Basic Duplicate Detection

kshepherd:
## References
* Replaces #1732
* Requires DSpace/DSpace#9265

The initial work on this feature was supported by TU Berlin. This work was supported by FHNW and ZHAW.

## Description
This feature introduces basic Duplicate Detection into DSpace submission and item links, using Solr's ability to search by levenshtein distance.

A new solr field `item_signature` is indexed for in-progress and archived items, with a normalised version of the title to allow it to be treated as a single term using the `~n` operator, where `n` is a configurable edit distance.

The potential duplicates detected by this search will be displayed to submitters or reviewers as a submission step, for their information (they could then choose to ignore the report if they know this is a new different item, or cancel their submission if there is a legitimate duplicate already present).

If a potential duplicate is in workspace, only the submitter of an item can see this information. 

Workflow reviewers are given a small hint indicating how many potential duplicates are detected for a pooled or claimed task, and directing them to the 'Edit' (form) where the step will show details. If the potential duplicate is in workflow, only a reviewer or admin will be able to see this information.

If versioning is used, only the latest version is included in comparison.

## Instructions for Reviewers
* Make sure you are using the backend PR DSpace/DSpace#9265
* Configure submission to include the new `duplicate` step
* Enable duplication in `dspace/config/modules/duplicate-detection.cfg` and configure as desired
* Add items with the exact same (or play around with edit distance) titles, and see what shows up in the Duplicate section 

## Checklist

### DIFF
--- a/src/app/core/data/duplicate-search.service.ts
+++ b/src/app/core/data/duplicate-search.service.ts
@@ -1,0 +1,101 @@
+/* eslint-disable max-classes-per-file */
+import { Observable } from 'rxjs';
+import { Injectable } from '@angular/core';
+import { map } from 'rxjs/operators';
+import { FollowLinkConfig } from '../../shared/utils/follow-link-config.model';
+import { ResponseParsingService } from './parsing.service';
+import { RemoteData } from './remote-data';
+import { GetRequest } from './request.models';
+import { RequestService } from './request.service';
+import { GenericConstructor } from '../shared/generic-constructor';
+import { HALEndpointService } from '../shared/hal-endpoint.service';
+import { SearchResponseParsingService } from './search-response-parsing.service';
+import { RemoteDataBuildService } from '../cache/builders/remote-data-build.service';
+import { RestRequest } from './rest-request.model';
+import { BaseDataService } from './base/base-data.service';
+import { FindListOptions } from './find-list-options.model';
+import { Duplicate } from '../../shared/object-list/duplicate-data/duplicate.model';
+import { PaginatedList } from './paginated-list.model';
+import { RequestParam } from '../cache/models/request-param.model';
+import { ObjectCacheService } from '../cache/object-cache.service';
+
+
+/**
+ * Service that performs all general actions that have to do with the search page
+ */
+@Injectable()
+export class DuplicateDataService extends BaseDataService<Duplicate> {
+
+  /**
+   * The ResponseParsingService constructor name
+   */
+  private parser: GenericConstructor<ResponseParsingService> = SearchResponseParsingService;
+
+  /**
+   * The RestRequest constructor name
+   */
+  private request: GenericConstructor<RestRequest> = GetRequest;
+
+  /**
+   * Subscription to unsubscribe from
+   */
+  private sub;
+
+  constructor(
+    protected requestService: RequestService,
+    protected rdbService: RemoteDataBuildService,
+    protected objectCache: ObjectCacheService,
+    protected halService: HALEndpointService,
+  ) {
+    super('duplicates', requestService, rdbService, objectCache, halService);
+  }
+
+  protected getEndpoint(): Observable<string> {
+    return this.halService.getEndpoint(this.linkPath);
+  }
+
+  /**
+   * Method to set service options
+   * @param {GenericConstructor<ResponseParsingService>} parser The ResponseParsingService constructor name
+   * @param {boolean} request The RestRequest constructor name
+   */
+  setServiceOptions(parser: GenericConstructor<ResponseParsingService>, request: GenericConstructor<RestRequest>) {
+    if (parser) {
+      this.parser = parser;
+    }
+    if (request) {
+      this.request = request;
+    }
+  }
+
+  private getSearchUrl(): Observable<string> {
+    const href$ = this.getEndpoint();
+    return href$.pipe(
+      map((href) => href + '/search')
+    );
+  }
+
+  public findDuplicates(uuid: string, options?: FindListOptions, useCachedVersionIfAvailable = true, reRequestOnStale = true, ...linksToFollow: FollowLinkConfig<Duplicate>[]): Observable<RemoteData<PaginatedList<Duplicate>>> {
+    const searchParams = [new RequestParam('uuid', uuid)];
+    let findListOptions = new FindListOptions();
+    if (options) {
+      findListOptions = Object.assign(new FindListOptions(), options);
+    }
+    if (findListOptions.searchParams) {
+      findListOptions.searchParams = [...findListOptions.searchParams, ...searchParams];
+    } else {
+      findListOptions.searchParams = searchParams;
+    }
+
+    return this.findListByHref(this.getSearchUrl(), findListOptions, useCachedVersionIfAvailable, reRequestOnStale, ...linksToFollow);
+  }
+
+  /**
+   * Unsubscribe from the subscription
+   */
+  ngOnDestroy(): void {
+    if (this.sub !== undefined) {
+      this.sub.unsubscribe();
+    }
+  }
+}
--- a/src/app/core/data/item-data.service.ts
+++ b/src/app/core/data/item-data.service.ts
@@ -46,6 +46,9 @@ import { RestRequestMethod } from './rest-request-method';
 import { CreateData, CreateDataImpl } from './base/create-data';
 import { RequestParam } from '../cache/models/request-param.model';
 import { dataService } from './base/data-service.decorator';
+import { Duplicate } from '../../shared/object-list/duplicate-data/duplicate.model';
+import { SearchDataImpl } from './base/search-data';
+import { FollowLinkConfig } from '../../shared/utils/follow-link-config.model';
 
 /**
  * An abstract service for CRUD operations on Items
@@ -56,6 +59,7 @@ export abstract class BaseItemDataService extends IdentifiableDataService<Item> 
   private createData: CreateData<Item>;
   private patchData: PatchData<Item>;
   private deleteData: DeleteData<Item>;
+  private searchData: SearchDataImpl<Duplicate>;
 
   protected constructor(
     protected linkPath,
@@ -74,6 +78,7 @@ export abstract class BaseItemDataService extends IdentifiableDataService<Item> 
     this.createData = new CreateDataImpl(this.linkPath, requestService, rdbService, objectCache, halService, notificationsService, this.responseMsToLive);
     this.patchData = new PatchDataImpl<Item>(this.linkPath, requestService, rdbService, objectCache, halService, comparator, this.responseMsToLive, this.constructIdEndpoint);
     this.deleteData = new DeleteDataImpl(this.linkPath, requestService, rdbService, objectCache, halService, notificationsService, this.responseMsToLive, this.constructIdEndpoint);
+    this.searchData = new SearchDataImpl(this.linkPath, requestService, rdbService, objectCache, halService, this.responseMsToLive);
   }
 
   /**
@@ -242,24 +247,18 @@ export abstract class BaseItemDataService extends IdentifiableDataService<Item> 
     );
   }
 
-  public getDuplicatesEndpoint(itemId: string): Observable<string> {
-    return this.halService.getEndpoint(this.linkPath).pipe(
-      switchMap((url: string) => this.halService.getEndpoint('duplicates', `${url}/${itemId}`))
-    );
-  }
-
-  public getDuplicates(itemId: string, searchOptions?: PaginatedSearchOptions): Observable<RemoteData<PaginatedList<Item>>> {
-    const hrefObs = this.getDuplicatesEndpoint(itemId).pipe(
-      map((href) => searchOptions ? searchOptions.toRestUrl(href) : href)
-    );
-    hrefObs.pipe(
-      take(1)
-    ).subscribe((href) => {
-      const request = new GetRequest(this.requestService.generateRequestId(), href);
-      this.requestService.send(request);
-    });
-
-    return this.rdbService.buildList<Item>(hrefObs);
+  public findDuplicates(uuid: string, options?: FindListOptions, useCachedVersionIfAvailable = true, reRequestOnStale = true, ...linksToFollow: FollowLinkConfig<Duplicate>[]): Observable<RemoteData<PaginatedList<Duplicate>>> {
+    const searchParams = [new RequestParam('uuid', uuid)];
+    let findListOptions = new FindListOptions();
+    if (options) {
+      findListOptions = Object.assign(new FindListOptions(), options);
+    }
+    if (findListOptions.searchParams) {
+      findListOptions.searchParams = [...findListOptions.searchParams, ...searchParams];
+    } else {
+      findListOptions.searchParams = searchParams;
+    }
+    return this.searchData.searchBy('findDuplicates', findListOptions, useCachedVersionIfAvailable, reRequestOnStale, ...linksToFollow);
   }
 
   /**
--- a/src/app/core/data/item-data.service.ts
+++ b/src/app/core/data/item-data.service.ts
@@ -242,6 +242,26 @@ export abstract class BaseItemDataService extends IdentifiableDataService<Item> 
     );
   }
 
+  public getDuplicatesEndpoint(itemId: string): Observable<string> {
+    return this.halService.getEndpoint(this.linkPath).pipe(
+      switchMap((url: string) => this.halService.getEndpoint('duplicates', `${url}/${itemId}`))
+    );
+  }
+
+  public getDuplicates(itemId: string, searchOptions?: PaginatedSearchOptions): Observable<RemoteData<PaginatedList<Item>>> {
+    const hrefObs = this.getDuplicatesEndpoint(itemId).pipe(
+      map((href) => searchOptions ? searchOptions.toRestUrl(href) : href)
+    );
+    hrefObs.pipe(
+      take(1)
+    ).subscribe((href) => {
+      const request = new GetRequest(this.requestService.generateRequestId(), href);
+      this.requestService.send(request);
+    });
+
+    return this.rdbService.buildList<Item>(hrefObs);
+  }
+
   /**
    * Get the endpoint to move the item
    * @param itemId
--- a/src/app/core/data/item-data.service.ts
+++ b/src/app/core/data/item-data.service.ts
@@ -46,9 +46,6 @@ import { RestRequestMethod } from './rest-request-method';
 import { CreateData, CreateDataImpl } from './base/create-data';
 import { RequestParam } from '../cache/models/request-param.model';
 import { dataService } from './base/data-service.decorator';
-import { Duplicate } from '../../shared/object-list/duplicate-data/duplicate.model';
-import { SearchDataImpl } from './base/search-data';
-import { FollowLinkConfig } from '../../shared/utils/follow-link-config.model';
 
 /**
  * An abstract service for CRUD operations on Items
@@ -59,7 +56,6 @@ export abstract class BaseItemDataService extends IdentifiableDataService<Item> 
   private createData: CreateData<Item>;
   private patchData: PatchData<Item>;
   private deleteData: DeleteData<Item>;
-  private searchData: SearchDataImpl<Duplicate>;
 
   protected constructor(
     protected linkPath,
@@ -78,7 +74,6 @@ export abstract class BaseItemDataService extends IdentifiableDataService<Item> 
     this.createData = new CreateDataImpl(this.linkPath, requestService, rdbService, objectCache, halService, notificationsService, this.responseMsToLive);
     this.patchData = new PatchDataImpl<Item>(this.linkPath, requestService, rdbService, objectCache, halService, comparator, this.responseMsToLive, this.constructIdEndpoint);
     this.deleteData = new DeleteDataImpl(this.linkPath, requestService, rdbService, objectCache, halService, notificationsService, this.responseMsToLive, this.constructIdEndpoint);
-    this.searchData = new SearchDataImpl(this.linkPath, requestService, rdbService, objectCache, halService, this.responseMsToLive);
   }
 
   /**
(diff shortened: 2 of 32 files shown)


src/app/shared/object-list/my-dspace-result-list-element/pool-search-result/pool-search-result-list-element.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, Inject, OnDestroy, OnInit } from '@angular/core';
 
 import { BehaviorSubject, EMPTY, Observable } from 'rxjs';
-import { mergeMap, tap } from 'rxjs/operators';
+import {map, mergeMap, tap} from 'rxjs/operators';
```
philipprumpf, line 4: Very minor coding style issue here :-)

pnbecker: thank you, fixed.

src/app/submission/sections/duplicates/section-duplicates.component.html
```diff
@@ -0,0 +1,20 @@
+<!--
+Template for the detect duplicates submission section component
+@author Kim Shepherd
+-->
+<div class="text-sm-left" *ngVar="(this.getDuplicateData() | async) as data">
+  <ng-container *ngIf="data.potentialDuplicates.length == 0">
+    <p>{{ 'submission.sections.duplicates.none' }}</p>
```
MW3000, line 7: Here it should be `{{ 'submission.sections.duplicates.none' | translate }}`
instead of `{{ 'submission.sections.duplicates.none' }}`. At the moment the key is displayed, not the message.

src/app/shared/object-list/my-dspace-result-list-element/claimed-search-result/claimed-search-result-list-element.component.html
```diff
@@ -4,7 +4,14 @@
                                [showSubmitter]="showSubmitter"
                                [badgeContext]="badgeContext"
                                [workflowItem]="workflowitem$.value"></ds-themed-item-list-preview>
-
+  <!-- Display duplicate alert, if feature enabled and duplicates detected -->
```
MW3000, line 7: Screenshot 2024-02-01 at 14 01 48

Duplicate alert needs to be aligned with other lines.

At his stage, the item already is claimed, so the alert should not tell users to "Claim and edit this item to see details." but just to "Edit this item to see details.".

src/app/shared/object-list/my-dspace-result-list-element/pool-search-result/pool-search-result-list-element.component.html
```diff
@@ -4,6 +4,15 @@
                                [showSubmitter]="showSubmitter"
                                [badgeContext]="badgeContext"
                                [workflowItem]="workflowitem$.value"></ds-themed-item-list-preview>
+  <!-- Display duplicate alert, if feature enabled and duplicates detected -->
```
MW3000, line 7: Screenshot 2024-02-01 at 16 35 54

Duplicate alert needs to be aligned with other lines.

MW3000 (approved):
Works as intended also using the new search endpoint.

src/app/shared/object-list/my-dspace-result-list-element/pool-search-result/pool-search-result-list-element.component.html
kshepherd: Fixed with a 'dummy' col to simulate thumbnail. 
screenshot_2024-02-11_12:18:45_selection

src/app/shared/object-list/my-dspace-result-list-element/claimed-search-result/claimed-search-result-list-element.component.html
kshepherd: Fixed in both claimed and pooled task template

src/app/shared/object-list/duplicate-data/duplicate.resource-type.ts
```diff
@@ -0,0 +1,9 @@
+import { ResourceType } from 'src/app/core/shared/resource-type';
+
+/**
+ * The resource type for Access Status
```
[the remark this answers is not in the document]
kshepherd: good catch, fixed

src/app/submission/sections/duplicates/section-duplicates.component.html
kshepherd: Fixed

kshepherd:
@MW3000 thank you for the further review. I have fixed the issues identified.

atarix83 (changes requested):
Thanks @kshepherd for this PR.

I've reviewed this PR and found out this is not ported as is from implemented on DSpace-CRIS.

I agree to keep it simple than the most comprehensive functionality that is present on DSpace-CRIS, but i've couple of suggestions i'd like to be ported here in this PR: 
- in my opinion is not necessary to show the submission step when no duplicates were detected because I don't think it gives any added value to the submission process
immagine
We hide mandatory section without section data here on DSpace-CRIS https://github.com/4Science/dspace-angular/blob/e4e919307924a2f912bdcc53cf00fd20fc18e8c0/src/app/submission/objects/submission-objects.effects.ts#L87C25-L88C70
- when a duplicate is detected i'd highlight the message with a warning alert, something like 
immagine
In this case i'd also change the section status to warning

Moreover in console there is a repetitive error which occurs when the submission is deposited due to the detect duplicate step



Schermata del 2024-02-15 19-55-21

src/app/shared/object-list/my-dspace-result-list-element/pool-search-result/pool-search-result-list-element.component.html
atarix83: @kshepherd 

as suggested in [my comment](https://github.com/DSpace/dspace-angular/pull/2749#pullrequestreview-1883632104) for the submission also here i'd highlight the message with a warning alert or a text warning class

pnbecker:
@atarix83 thank you for the review! Yes, this isn‘t a port anymore. As the port didn‘t got reviewed for several DSpace versions, we reimplemented this to get much smaller PRs. We hoped to get it reviewed this time by reducing the PR to roughly 20% of the original size.

pnbecker:
@atarix83 we addressed your feedback. Can you please confirm?

kshepherd:
@atarix83 thanks for your review, I have addressed your feedback as below:

1. Warning alert panel for "potential duplicates" message in submission section, claimed task and pooled task

screenshot_2024-02-19_16:49:02_selection

2. Section data is not assumed to exist in the component template so those errors reading from missing section data will not occur
3. I made the default behaviour of the submission section to hide the duplicate section (at form load and after parsing save response, using very similar approach to the CRIS code you linked) when there are no duplicates to display.

However, if the angular env config `submission.duplicateDetection.alwaysShowSection` is set to `true` (default false) then the section will show the "no duplicates" message, in case that fits the user requirements. The message is displayed in an info alert panel:

screenshot_2024-02-19_16:48:52_selection
(ignore that black border, that is my screenshot utility, not the actual web display)

src/app/shared/object-list/my-dspace-result-list-element/pool-search-result/pool-search-result-list-element.component.html
kshepherd: done, thanks

atarix83 (commented):
@pnbecker  @kshepherd 

i'm sorry i didn't realize that the rest part is fully changed and my thought was that only the Angular interface has been simplified.

also based on what @abollini  pointed out in this comment https://github.com/DSpace/RestContract/pull/252#pullrequestreview-1896437014 I'd not approve this PR.

There are any changes the old porting PR is resumed?

kshepherd:
@tdonohue I have moved duplicate search API method to a separate controller and angular data service as per comments on the related PRs, and rebased against main.

src/app/core/data/item-data.service.ts
```diff
@@ -242,6 +247,20 @@ export abstract class BaseItemDataService extends IdentifiableDataService<Item>
...
+    let findListOptions = new FindListOptions();
+    if (options) {
+      findListOptions = Object.assign(new FindListOptions(), options);
+    }
+    if (findListOptions.searchParams) {
+      findListOptions.searchParams = [...findListOptions.searchParams, ...searchParams];
+    } else {
+      findListOptions.searchParams = searchParams;
+    }
+    return this.searchData.searchBy('findDuplicates', findListOptions, useCachedVersionIfAvailable, reRequestOnStale, ...linksToFollow);
```
tdonohue, line 261: This appears to be code using the *old* endpoint under `/items/search/findDuplicates`?

src/app/core/data/duplicate-search.service.ts
```diff
@@ -0,0 +1,101 @@
...
+import { BaseDataService } from './base/base-data.service';
+import { FindListOptions } from './find-list-options.model';
+import { Duplicate } from '../../shared/object-list/duplicate-data/duplicate.model';
+import { PaginatedList } from './paginated-list.model';
+import { RequestParam } from '../cache/models/request-param.model';
+import { ObjectCacheService } from '../cache/object-cache.service';
+
+
+/**
+ * Service that performs all general actions that have to do with the search page
```
tdonohue, line 24: This needs a better description.  It obviously is only used in the submission process to try to detect duplicates.

Optionally, it could even be moved to `/app/core/submission` and named `submission-duplicate-serch.service.ts`.... that'd be similar to how there's a `submission-cc-license-data.service.ts`, etc.

src/app/shared/object-list/duplicate-data/duplicate.model.ts
```diff
@@ -0,0 +1,42 @@
+import {autoserialize, deserialize} from 'cerialize';
+import { MetadataMap } from '../../../core/shared/metadata.models';
+import { HALLink} from '../../../core/shared/hal-link.model';
+import { CacheableObject } from '../../../core/cache/cacheable-object.model';
+import { DUPLICATE } from './duplicate.resource-type';
+import { ResourceType } from '../../../core/shared/resource-type';
+
+export class Duplicate implements CacheableObject {
```
tdonohue, line 8: Please describe this class with a comment. That will help other developers understand where/how it is used

src/app/submission/sections/duplicates/section-duplicates.component.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+        'language': null,
+        'authority': null,
+        'confidence': -1,
+        'place': 0
+      })]
+    },
+    type: DUPLICATE,
+    _links: {
+      self: {
+        href: 'http://localhost:8080/server/api/core/items/search/findDuplicates?uuid=testid'
```
tdonohue, line 79: URL needs updating to match new endpoint

tdonohue (commented):
@kshepherd : Gave this a test/review today.  The basic feature works for a new submission.  But, I'm finding that the "Potential Duplicates" submission step **does not load** if I'm reviewing a submission in a workflow step.  Here's a brief video that shows the behavior...it starts on a new submission & shows that the step appears there. But, if I switch to one in workflow, the step does **not appear**

duplicate-step

Beyond that, the basics appear to work.  I have a few other inline comments below that should be addressed or responded to.

tdonohue:
@kshepherd : I just found a second bug which will need to be fixed

When I disable this feature (`duplicate.enable=false` or commented out), the checks for Duplicates **still occur** on the Workflow pages. To reproduce:
* Disable the feature.  Open up your browser DevTools to watch the "Network" tab
* Go to MyDSpace, select "Workflow tasks"
* In the "Network" Tab of DevTools you'll see that a call to `[dspace.url]/server/api/duplicates/search?uuid=[uuid]` is still made for every Item in workflow.

MW3000:
I tested the feature again with the new changes. For me the feature worked well. I did not experience the bug @tdonohue was describing for the workflow step. The duplicate section did appear in the workflow step as well.

src/app/core/data/item-data.service.ts
kshepherd: yep, removed - missed it on my first run

src/app/core/data/duplicate-search.service.ts
kshepherd: i have moved the service to /app/core/submission and renamed it to submission-duplicate-data-service.ts

src/app/shared/object-list/duplicate-data/duplicate.model.ts
kshepherd: done

src/app/submission/sections/duplicates/section-duplicates.component.spec.ts
kshepherd: done

kshepherd:
@tdonohue regarding the requests being made even if the feature was disabled - i had a check on the backend that always returned an empty response if the feature was disabled (now that will be 404 as discussed, with my other change).
To better achieve what you want here, I have reworked things so that `duplicate.enable` is exposed via REST, and the claimed/pooled task components will only request duplicate searches if that config comes back as enabled. I hope I have done it in the most efficient manner - it is async, and no new subscriptions were introduced.

I moved the api path to /core/submission/ as suggested, and renamed my service accordingly.

I could not reproduce the problem you had, but i haven't got my local env tested in production yet (will do when i get time). The only thing i could thing of is some unexpected submission configuration?

kshepherd:
not sure why these e2e tests are failing - i didn't touch a 'profiles' endpoint or edit-item...

pnbecker:
I restarted the tests and now they passed all.

src/app/shared/object-list/my-dspace-result-list-element/pool-search-result/pool-search-result-list-element.component.ts
```diff
@@ -85,6 +101,14 @@ export class PoolSearchResultListElementComponent extends SearchResultListElemen
       followLink('submitter')
     ), followLink('action'));
 
+    // Get configuration for duplicate detection feature
+    this.enableDetectDuplicates$ = this.configService.findByPropertyName('duplicate.enable').pipe(
```
tdonohue, line 105: This variable doesn't appear to be used anywhere?  I'm confused why you set this variable but then never seem to use it.  (Maybe I'm missing where it is used?)

src/app/submission/objects/submission-objects.reducer.ts
```diff
@@ -856,3 +860,20 @@ function deleteFile(state: SubmissionObjectState, action: DeleteUploadedFileActi
   }
   return state;
 }
+
+function cleanDuplicateDetectionSection(state: SubmissionObjectState, action: CleanDuplicateDetectionAction): SubmissionObjectState {
+  if (isNotEmpty(state[ action.payload.submissionId ])) {
+    return Object.assign({}, state, {
+      [ action.payload.submissionId ]: Object.assign({}, state[ action.payload.submissionId ], {
+        sections: Object.assign({}, state[ action.payload.submissionId ].sections, {
+          [ 'duplicates' ]: Object.assign({}, state[ action.payload.submissionId ].sections.duplicates, {
```
tdonohue, line 869: Is this maybe where the error occurs?  It appears this might be running even if duplicate detection is disabled.

tdonohue (changes requested):
@kshepherd : Thanks!  I'm **nearly** +1, but your fixes resulted in another small bug.

First though, the good news is that I can *no longer* reproduce that Workflow editing bug.  It must've been something on my end.  All your other fixes also look correct to me (just a few minor comments inline below).

However, the **new bug** is that I'm now seeing a Javascript error on the Workflow tasks page when this feature is **disabled**.  Here's how to reproduce it:
1. First, ensure it's disabled (`duplicate.enable=false` or comment out).  Restart Tomcat if you just changed this value as it requires a Tomcat reboot
2. Now, submit a new item that will trigger this duplicate check to a Collection with workflow enabled  (This step might be optional?)
3. Open up your browser's DevTools window.  The error will only be visible here.
4. Login as a user with workflow approval rights & go to MyDSpace -> Workflow Tasks
5. In your DevTools window, you'll see two "errors" (one is expected)
    * First, you may see the 404 response to `/server/api/config/properties/duplicate.enable` . This is **expected behavior** as it's disabled.
    * Second, you'll see a bug in the Console that says:
      ```
       Error:  (Large amount of JSON) doesn't contain the link duplicates
      ```

I think I might have located where in your code this error comes from (see inline below).  I think there's one bit of code that is *not being disabled* when this feature is disabled.  But, the expected behavior here is that we should ONLY see that 404 response to the config check.  No other errors should occur.

src/app/submission/objects/submission-objects.reducer.ts
kshepherd: I found an error that occurs in submission-objects effects (and potentially in reducer as well as you identified), and added fixes so that there is no assumption the 'duplicates' section is present when it tries to clean/remove it.

Regarding the "no duplicates link" - this is similar to the error i saw with profiles, and which the e2e tests in this PR encountered once.
Regarding duplicates, this is sort of a problem we introduced - when we use that annotation to disable the controller when the feature is disabled, it is no longer registered in teh main list of endpoints. That is where the 'duplicates' is missing from, when the data service tries to resolve the endpoint. That said, I'm not sure how or why your prod instance is trying to resolve that endpoint since that only occurs with a findDuplicates call, which itself is conditional on the config property being enabled.

Is it possible you changed the config property so the feature detection came back as 'enabled', but since tomcat wasn't fully restarted, the controller had not re-initialised and therefore was still 404'd / not registered in the endpoint lists? If so, is that a scenario we should anticipate? I could introduce a second config check into the data service itself, perhaps

src/app/shared/object-list/my-dspace-result-list-element/pool-search-result/pool-search-result-list-element.component.ts
kshepherd: removed

tdonohue:
@kshepherd : I'm noticing this is now failing Codecov checks (sigh).  It has dropped just below the "-0.5%" threshold for whether it has enough tests (we allow PRs to only drop overall code coverage by 0.5% at a maximum).

It looks like your latest changes were to `submission-objects.effects.ts` and `submission-objects.reducer.ts`.  Both of these already have related spec files.  Maybe you could add specs for your small updates to bring the code coverage back up slightly?

In the meantime, I'm giving this a quick re-test to see if prior issues are resolved.

src/app/submission/objects/submission-objects.reducer.ts
tdonohue: @kshepherd : I'm able to reproduce this issue with the latest code.  I restarted Tomcat completely with this feature disabled by default.  When I run the UI via `yarn build:prod; yarn serve:ssr`, I can reproduce this issue whenever I visit MyDSpace and select "Workflow tasks" from the dropdown.

I *know* other features are doing this properly...as we have a lot of features which only get enabled via an `enable` boolean.  I'm not exactly sure how they fix this issue, but there must be a workaround.  I'm looking around at other features for examples.

kshepherd: I will try as well. Perhaps it is something as simple as forcing an initial state of 'false' for the config property if that observable somehow is initially evaluated in a state where the endpoint is attempted to be queried. (or if my combineLatest block is otherwise not quite perfect - async angular operations like this are not my area of expertise)

tdonohue: I have to leave for the day...but I admit this isn't my area of expertise either.  I just know that we *don't* have these errors appearing all over the application every time we disable a feature on the backend with `@ConditionalOnProperty`. 

So, there has to be a way to achieve this, but I'm not sure myself how others have done this.  If it comes down to it, I can ask others in tomorrow's DevMtg to provide advice here.....just let me know if you cannot find a better solution.

(I'm *hoping* maybe if you can reproduce this it might become more obvious to you what may be happening...we'll see)

src/app/shared/object-list/my-dspace-result-list-element/claimed-search-result/claimed-search-result-list-element.component.ts
```diff
@@ -90,13 +104,58 @@ export class ClaimedSearchResultListElementComponent extends SearchResultListEle
       tap((itemRD: RemoteData<Item>) => {
         if (isNotEmpty(itemRD) && itemRD.hasSucceeded) {
           this.item$.next(itemRD.payload);
+          this.duplicates$ = this.duplicateDataService.findDuplicates(itemRD.payload.uuid).pipe(
```
tdonohue, line 107: @kshepherd : I believe I located the cause of the error that I was getting.  The "Claimed" search results list appears to load the duplicates here **without** first checking if the feature is enabled.  

That's different behavior from the "Pool" search results (in `pool-search-result-list-element.component.ts`) where the code first checks to see if duplicate detection is enabled before searching for duplicates.

tdonohue:
@kshepherd : If you have time to do a final cleanup, I believe this PR just needs two things:
1. A fix to the error I was getting... I think the problem is [in `claimed-search-result-list-element.component.ts`](https://github.com/DSpace/dspace-angular/pull/2749#discussion_r1507694399)
2. A [few more specs/tests](https://github.com/DSpace/dspace-angular/pull/2749#issuecomment-1970047639) so that it no longer fails our CodeCov checks

Once those are resolved, I can give it a final test & then merge both the backend & frontend (assuming everything works as planned)

kshepherd:
pushed a large rebase (lots of conflicts with modules now that other submission PRs are pushed and this one has so many commits), and includes changes to use searchBy endpoint again as per last reversion/change in backend PR

src/app/shared/object-list/my-dspace-result-list-element/claimed-search-result/claimed-search-result-list-element.component.ts
kshepherd: Yes, you're right -- just a case of staring at the same code for days and missing this! Thanks, removed and fixed.

kshepherd:
@tdonohue although code cov looked ok after i made my last changes, i also added new tests for the data service and the object reducer / actions, to keep things in line with general good practice and get coverage up a bit more, hope that works well

tdonohue:
Flagging as `needs documentation` until Documentation is added for this new feature to the pre-8.0 docs: https://wiki.lyrasis.org/display/DSDOC8x/Submission+User+Interface

tdonohue:
Removing "needs documentation" flag as docs now exist in https://wiki.lyrasis.org/pages/viewpage.action?pageId=328958055

PR #2770, opened 2024-01-25
German translations of feedback page translation keys

saschaszott:
## Description

This PR adds missing German translation of all feedback page translation keys.


tdonohue (approved):
👍 Thanks @saschaszott ! Looks good. Merging immediately & auto-backporting to 7.x

PR #2818, opened 2024-02-19
site search should not adopt existing URL query string parameters

saschaszott:
## Description

A site search (issued by the search box in the navigation bar) should not adopt existing URL query parameters. For example, a site search that is issued in the context of MyDSpace will adopt `configuration=workspace` as a URL query parameter.

The normal user would not expect that a site search considers only a subset of all items in the repository. It is commonly expected that a site search behaves consistent regardless of the context from which the search was started.


alanorth:
Thanks @saschaszott! I confirmed the bug and the fix on DSpace 7.6.1.

PR #2820, opened 2024-02-20
Add translation for DOI multiple parallel import source.

philipprumpf:
## References
* Requires DSpace/DSpace#9132

## Description
Add translation key


tdonohue (approved):
Thanks @philipprumpf !

PR #2827, opened 2024-02-22
use translation service to determine placeholder values of date picker

### LINKED ISSUE #2825: missing i18n of placeholder values in date picker
# Bug Description

Placeholder values (year, month, day) in the date picker form element are hard-coded in 

https://github.com/DSpace/dspace-angular/blob/d5aca7633c666e8827eb6645868d4fffd6b04bb6/src/app/shared/form/builder/ds-dynamic-form-ui/models/date-picker/date-picker.component.ts#L51

image

saschaszott:
## References

Fixes #2825


saschaszott:
@tdonohue , all conflicts were resolved.

tdonohue:
@saschaszott : This has code conflicts. If you wish to have it still considered for 8.0, please get this PR cleaned up and updated to latest `main`. Otherwise, we'll have to reschedule for 9.0. Thanks!

tdonohue:
@saschaszott : This PR currently has lint errors.  You may be able to solve them by running `yarn lint --fix`.

saschaszott:
@tdonohue , lint errors were resolved - PR is now ready for merging ;)

nwoodward:
I tested this PR, and it works well. It's an obvious win to have these strings translated. But I noticed that the 'y' in Day is cut off in the dropdown (screenshot below). It's also cut off on `main`, so it should be fixed with or without this PR. But these changes could make the problem worse if a translation is longer than "day", so it may be worth fixing it here.

Screenshot 2024-11-26 at 5 01 05 PM

src/app/shared/form/builder/ds-dynamic-form-ui/models/date-picker/date-picker.component.ts
alexandrevryghem, line 1: @saschaszott: Thnx for the PR! This works fine, but could you use the `translate` pipe in the HTML template instead. That way we don't need to deal with the `Subscription` returned by `this.translationService.get().subscribe()`. So something like this:
```html
[placeholder]="'form.date-picker.placeholder.year' | translate"
```

saschaszott: @alexandrevryghem , thanks for reviewing this PR and your feedback.

Regarding your suggestion: Is this a general issue? I see that there is some variation in using the `translate` pipe in HTML templates, e.g. in this template https://github.com/DSpace/dspace-angular/blob/main/src/app/shared/form/form.component.html you'll find

```
[attr.aria-label]="'form.remove' | translate"
```

and

```
attr.aria-label="{{'form.add' | translate}}"
```

Which one is "better"? Should we harmonize this in the entire code base?

alexandrevryghem: @saschaszott: Both work, I personally prefer the one with the square brackets, but I don't think that its something that we schould try to "fix" everywhere. The problem I have with this PR is that it doesn't perform the translation in the [HTML template](https://github.com/DSpace/dspace-angular/blob/b47f2cba9b62ef115b7f2e98c8b6dd241c2405fe/src/app/shared/form/builder/ds-dynamic-form-ui/models/date-picker/date-picker.component.html#L17), but in the the typescript code instead. Your solution requires additional [subscribe calls](https://github.com/DSpace/dspace-angular/blob/b47f2cba9b62ef115b7f2e98c8b6dd241c2405fe/src/app/shared/form/builder/ds-dynamic-form-ui/models/date-picker/date-picker.component.ts#L122-L124), from which you manually need to unsubscribe. By using the `translate` pipe you can avoid all those issues, so that's why I would replace [this line](https://github.com/DSpace/dspace-angular/blob/b47f2cba9b62ef115b7f2e98c8b6dd241c2405fe/src/app/shared/form/builder/ds-dynamic-form-ui/models/date-picker/date-picker.component.html#L17) with:
```html
[placeholder]="'form.date-picker.placeholder.year' | translate"
```
(Same applies for the month & day translation)

tdonohue (approved):
👍 Thanks @saschaszott !  I had a chance to test this today and can confirm it's working for English & German.  Code looks correct and all the prior feedback looks to be resolved. 

I'm going to squash & merge this to avoid the messy commit history.  But, will try to add this to 8.1 & 7.6.3 since it's tiny.

PR #2832, opened 2024-02-23
[Port dspace-7_x] Refactored community & collection pages

alexandrevryghem:
### References
- Fixes: #2808
- Related: #2800

### Description
Partial port of #2722 to `dspace-7_x`:
> The browse by dateissued has an infinite load animation when there are 0 items to display

### DIFF
--- a/src/app/browse-by/browse-by-date-page/browse-by-date-page.component.spec.ts
+++ b/src/app/browse-by/browse-by-date-page/browse-by-date-page.component.spec.ts
@@ -23,6 +23,7 @@ import { PaginationServiceStub } from '../../shared/testing/pagination-service.s
 import { APP_CONFIG } from '../../../config/app-config.interface';
 import { environment } from '../../../environments/environment';
 import { SortDirection } from '../../core/cache/models/sort-options.model';
+import { cold } from 'jasmine-marbles';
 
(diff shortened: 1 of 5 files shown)


tdonohue (changes requested):
@alexandrevryghem : Thanks for creating this port.  The good news is this fixes the bug with Browse by Date. The bad news, though, it seems to **break** some of the other Browse By pages.  Here's what I tried:
1. I ran this PR in production mode (`yarn build:prod; yarn serve:ssr`) *against* the demo.dspace.org backend, comparing it to the results at https://demo.dspace.org
    ```
    rest:
      ssl: true
      host: demo.dspace.org
      port: 443
      nameSpace: /server
    ```
2. Browse by Date works perfectly now, even if I delete the `dc.date.issued` field in the first result, or if I update it so that it has an empty value
3. Browse by Title **no longer works**. Page never loads
4. Browse by Author will load the author list. But, if you click on an Author name the next page **no longer works**. Page never loads
5. Browse by Subject shows the same behavior as Browse by Author.

I'm not sure what the exact cause is...but this PR seems to have unexpected side effects.

alexandrevryghem:
@tdonohue: This backport was indeed missing one fix, I cherry-picked it and everything seems to work correctly now

PR #2851, opened 2024-03-05
fix selector value in CommunitySidebarSearchListElementComponent

saschaszott:
## Description

This PR corrects the selector value of `CommunitySidebarSearchListElementComponent`. It was previously discussed in Slack: https://dspace-org.slack.com/archives/C3TTSEB1V/p1709632996100769


artlowel (approved):
Thanks @saschaszott!

PR #2887, opened 2024-04-04
Renaming DSpace to DSpace 8

paulo-graca:
## Description
Just changing the name presented in the homepage to be DSpace 8.


tdonohue (approved):
👍 Thanks @paulo-graca !

PR #2949, opened 2024-04-17
Fix simple item view author link

### LINKED ISSUE #2948: Item View Authors Link Resolving to Incorrect Browse by Query Parameter
**Describe the bug**

The author link on an item view does not work as expected. It has been discovered that the query parameter `value` matching exactly the author string literal works. The `startsWith` query parameter does not.

**To Reproduce**
1. Create community
2. Create collection
3. Submit item
4. View newly submitted item
5. Click the author link

**Actual behavior**

image

image

> Browsing by Author, starting with "Test Author"

**Expected behavior**

image

image

wwelling:
## References
* Fixes #2948

## Description
Fixes the simple item view author links.

## Instructions for Reviewers
See issue above.

List of changes in this PR:
* First, returned `this.value.type`, which is hardcoded "valueList", instead of data type that could never evaluate to the desired query parameter for the URL link to the browse by author.
* Second, update tests to test the static reference to `this.value.type`.

## Checklist

### DIFF
--- a/src/app/core/shared/value-list-browse-definition.model.ts
+++ b/src/app/core/shared/value-list-browse-definition.model.ts
@@ -31,6 +31,32 @@ export class ValueListBrowseDefinition extends NonHierarchicalBrowseDefinition {
   };
 
   getRenderType(): string {
-    return this.dataType;
+    // FIXME
+    // 4/17/2024 9:03 AM CDT https://github.com/DSpace/dspace-angular/blob/dspace-7.6.1/src/app/core/shared/non-hierarchical-browse-definition.ts
+    // dataType: BrowseByDataType;
+
+    // 4/17/2024 9:03 AM CDT https://github.com/DSpace/dspace-angular/blob/dspace-7.6.1/src/app/browse-by/browse-by-switcher/browse-by-decorator.ts#L9
+    // export enum BrowseByDataType {
+    //   Title = 'title',
+    //   Metadata = 'text',
+    //   Date = 'date'
+    // }
+
+    // 4/17/2024 9:03 AM CDT https://github.com/DSpace/dspace-angular/blob/dspace-7.6.1/src/app/core/shared/value-list-browse-definition.resource-type.ts
+    // ...
+    // 4/17/2024 9:05 AM CDT https://github.com/DSpace/dspace-angular/blob/dspace-7.6.1/src/app/shared/object-list/metadata-representation-list-element/browse-link/browse-link-metadata-list-element.component.ts
+    // 4/17/2024 9:06 AM CDT https://github.com/DSpace/dspace-angular/blob/dspace-7.6.1/src/app/shared/object-list/metadata-representation-list-element/browse-link/browse-link-metadata-list-element.component.html#L5
+    // ...
+    // 4/17/2024 9:10 AM CDT https://github.com/DSpace/dspace-angular/blob/dspace-7.6.1/src/app/item-page/simple/metadata-representation-list/themed-metadata-representation-list.component.ts
+    // 4/17/2024 9:10 AM CDT https://github.com/DSpace/dspace-angular/blob/dspace-7.6.1/src/app/item-page/simple/metadata-representation-list/metadata-representation-list.component.html
+    // ...
+    // 4/17/2024 9:11 AM CDT https://github.com/DSpace/dspace-angular/blob/dspace-7.6.1/src/app/shared/metadata-representation/metadata-representation-loader.component.ts
+    // 4/17/2024 9:11 AM CDT https://github.com/DSpace/dspace-angular/blob/dspace-7.6.1/src/app/shared/metadata-representation/metadata-representation-loader.component.html
+
+    // Return `VALUE_LIST_BROWSE_DEFINITION.value` for simple item view to render a link for 'dc.contributor.author' or 'dc.creator' of Person item type to the correct browse by URL with queryParam `value` and not `startsWith`.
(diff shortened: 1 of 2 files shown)


src/app/core/shared/value-list-browse-definition.model.ts
```diff
@@ -31,6 +31,6 @@ export class ValueListBrowseDefinition extends NonHierarchicalBrowseDefinition {
   };
 
   getRenderType(): string {
-    return this.dataType;
+    return this.type.value;
```
ybnd, line 34: I encountered a few similar issues today while resolving merge conflicts on #2865, which included upgrading TSESLint and introducing a new default rule for enum consistency
- There were a few cases these `getRenderTypes` calls being compared with a ResourceType, and I was considering replacing them with `type` instead
- If we change this browse definition we should probably do the same for the other ones
- If we use `this.type` here, do we still need the method at all?

wwelling: https://github.com/DSpace/dspace-angular/blob/fd000cdfc4b4b53074e1fe8b1b6b9a10f383305e/src/app/core/shared/browse-definition.model.ts#L15

The abstract inheritance requires it of implementations. I don't think I would make that many changes to resolve a bug. The component and class structure might require revisiting and that would be a scope creep within this patch. Additionally, as the unit test required breaking apart and it went under the coverage of e2e tests, I would think e2e tests are required to handle an instance of significant refactoring. Then again, I have seen larger changes to fix a bug before. 🥁

wwelling: Perhaps e2e test could be in the forward port to the next release.

ybnd: Good point about the scope creep, should probably be it's own issue/PR 👍

wwelling:
@tdonohue does `port to main` label require another PR into main for this or will there be additional work involved to resolve for acceptance into main?

tdonohue:
@wwelling : This is simply waiting on a reviewer.  Because the PR is not built on `main` it has a lower priority right now as we are finishing up the 8.0 release (and have a lot of outstanding bug fixes we are working on reviewing).  If you wish to create a version on `main` you are welcome to do so.  However, assuming the code you modified exists on both `dspace-7_x` and `main` (i.e. this PR can be merged cleanly into `main` as well), then it will be ported automatically once merged.

If you'd like to speed up the review process for any of your PRs, we have a PR Trading option where you can volunteer to review others PRs in exchange for this one being reviewed more quickly. https://wiki.lyrasis.org/display/DSPACE/Trading+reviews+on+Pull+Requests   

Otherwise, this PR will be reviewed as soon as we find a volunteer who has time to review it.  Or, as soon as someone else finds this useful and just tests it and reports back on this PR.

tdonohue:
@wwelling : I found time to look more closely at this today. Because of the changes in this area of the code between 7.x and 8.x, this PR is not able to be applied to pre-8.0 (`main`).  I'm not certain how to move this forward, as the code on `main` is quite different.  

If you find time to create a secondary PR for 8.0 (against `main`) that could help this PR along. Overall, I'm trying to ensure any bug fixes can be applied to both branches (`dspace-7_x` and `main`), assuming they affect both branches.

wwelling:
@tdonohue this resolves 7x and the PR for 8x as you discovered will be significantly different. How is this patch being blocked with that discovery?

tdonohue:
@wwelling : When a bug is discovered in DSpace, we obviously want to ensure it is patched in both 7.x and 8.x if it affects both.   I've not had time to verify yet whether this affects both, but my assumption was that this bug likely does (unless I've overlooked us patching it in a different manner in 8.x).

In general, we refrain from fixing bugs *only in 7.x* unless they are verified to no longer impact 8.x.  The reason is that we simply don't want a bug to be fixed in an older release while it still appears in a newer release....that can causes institutions upgrading to be frustrated as bugs that were previously fixed will reappear.

I haven't had time to get back to this PR myself to do a more thorough analysis (as I'm busy getting 8.0 finished up).  I'm hoping another developer can test whether this bug is truly 7.x specific and no longer exists on 8.x.   Perhaps one easier way to do so is to look at the differences in behavior on demo.dspace.org (running 7.x) versus sandbox.dspace.org (running 8.x), if it's possible to reproduce the behaviors there.

wwelling:
Is still a bug in 8. This is where the author link on the simple item view is from `dc.contributor.author` and not part of relationship such as [relation.isAuthorOfPublication](https://sandbox.dspace.org/entities/publication/017138d0-9ced-4c49-9be1-5eebe816c528).

https://sandbox.dspace.org/entities/publication/e1ffc1c5-284d-492c-9742-b18f0c4f8a43

image

https://sandbox.dspace.org/browse/author?startsWith=Gillen,%20Martie

tdonohue:
It sounds like this fix needs to also have a `main` PR.  Otherwise, we only fix this issue in 7.x and then 8.0 will revert back to doing a `startsWith=[name]` search instead of a `value=[name]` as [noted above](https://github.com/DSpace/dspace-angular/pull/2949#issuecomment-2129879836).

I'd hesitate to only merge this into 7.x because of that behavior change.  Ideally, we should have this feature working the same way in both 7.x and 8.x.  (And, as noted, the implementation in this PR won't work for 8.x. The modified `ValueListBrowseDefinition.getRenderType()` method now returns a `BrowseByDataType` instead of a simple `string` in 8.x.)

tdonohue:
An alternative fix has been added in #3159 (to potentially replace this PR).  This alternative PR seems more compatible with 8.x, so we may want to verify whether it is a complete replacement for this PR or not.

PR #2968, opened 2024-04-19
[Port dspace-7_x] Fixed pagination issues on item mapper

alexandrevryghem:
Port of #2642 to `dspace-7_x`


tdonohue (approved):
👍 Thanks @alexandrevryghem !  I reviewed the code and tested on the `dspace-7_x` branch using demo.dspace.org as the backend.  This fixes the pagination issues and I've verified that it doesn't have negative effects on other paginated pages (I tested item/collection mapping, eperson/groups, and many other areas)

PR #3096, opened 2024-05-31
Change - Metadata field selector, add infinite scroll for data paginated

### LINKED ISSUE #2730: List of available metadata fields is truncated when adding new metadata in "Edit Item"
**Describe the bug**
In DSpace 7.6, when adding new metadata to an item via the "Metadata" tab of the "Edit Item" interface, the list of available fields is truncated at ten and it is not possible to scroll or see more.

Screenshot 2024-01-08 at 17-14-42 DSpace Repository Item Edit - Metadata-fs8

**To Reproduce**
1. Log into the demo.dspace.org as a user with rights to edit an item
2. Go to "Edit Item" → "Metadata" and click "Add"
3. The list of available metadata fields is limited to ten

**Expected behavior**
The text field has autocomplete, so you can type a few letters to filter, but this **requires you to know what field you are looking for**. For example, I typed "dcterms" and I get a list of the first ten fields matching those characters:

Screenshot 2024-01-08 at 17-15-07 DSpace Repository Item Edit - Metadata-fs8

It would be nice if there was a way to scroll the list, or see all available fields.

VictorHugoDuranS:
Hi @tdonohue , I'm @jtimal partner, I like to share this PR with you:

## References
* Fixes #2730 

## Description
 I changed metadata field selector add infinite scroll for pagination all data from metadata field service, this change load all options in the start

## Instructions for Reviewers

To  see the new implementation, I attch a video with the functionality: https://youtu.be/8D-EjhubABI

List of changes in this PR:
* I changed the template metadata-field-selector.component.html adding the infinite scroll directives
* I changed the search function to page the query detecting change value on input
* I changed the observable list to "behaviorsubject" for update the current list
* I added a function for update the current list options according the query search, only if the user scroll to last item of the list

## Checklist


src/app/dso-shared/dso-edit-metadata/metadata-field-selector/metadata-field-selector.component.ts
```diff
@@ -112,32 +143,33 @@ export class MetadataFieldSelectorComponent implements OnInit, OnDestroy, AfterV
    * Update the mdFieldOptions$ depending on the query$ fired by querying the server
    */
   ngOnInit(): void {
+    this.input.valueChanges.pipe(
+      debounceTime(this.debounceTime),
+      startWith(''),
+      tap( () => this.currentPage$.next(1) )
```
artlowel, line 149: It's not a good idea to "perform work" in a tap. Taps are meant for debugging purposes, and shouldn't affect anything else. The main reason for that is that the code in the tap won't be executed if nothing is subscribed to the pipe in the end. Since you're subscribing immediately underneath this line that's not an issue here, but then there's also no good reason not to move this code to the subscribe as well.

src/app/dso-shared/dso-edit-metadata/metadata-field-selector/metadata-field-selector.component.ts
```diff
@@ -112,32 +143,33 @@ export class MetadataFieldSelectorComponent implements OnInit, OnDestroy, AfterV
...
+    ).subscribe((valueChange) => {
```
artlowel, line 150: Please also add this subscribe to the subs array, so it can be unsubscribed onDestroy

pilasou:
Hi, I tested the PR. It works as described : when adding a metadata field to an archived Item, full list of metadata available is listed in the Field Column with ability to scroll down : 
image

When starting to type metadata name (i.e dc.identifier) the metadata list is filtered accordingly and can still be scrolled.

src/app/dso-shared/dso-edit-metadata/metadata-field-selector/metadata-field-selector.component.html
```diff
@@ -6,13 +6,30 @@
...
+          <button *ngFor="let listEntry of (mdFieldOptions$ | async)"
+                  class="d-block dropdown-item"
+                  dsHoverClass="ds-hover"
+                  (click)="select(listEntry)" #listEntryElement>
+            <span [innerHTML]="listEntry"></span>
+          </button>
+        </ng-container>
+        <button *ngIf="loading"
+                class="list-group-item list-group-item-action border-0 list-entry">
+          <ds-themed-loading [showMessage]="false"></ds-themed-loading>
```
tdonohue, line 32: When this is ported to `main`, I believe this `<ds-themed-loading>` tag will need to be `<ds-loading>` because of the changes in 8.0 applied in #2865. 

So, this is an example of why we need a separate PR for `main`.  This current code is correct for 7.x, but will fail in 8.x or `main`.

tdonohue (changes requested):
@VictorHugoDuranS : We reviewed this as a team in our weekly DevMtg today.  Some of the feedback can be found above.  One other note to add is that we'd need to have a version of this PR built against `main`, so that we can port it to 8.x and 9.0.   If you could create a separate PR for `main`, that'd be appreciated.  (An alternative would be to update this PR to be against `main`, and we can then just apply this fix in 9.0 and 8.x only).  Thanks!

src/app/dso-shared/dso-edit-metadata/metadata-field-selector/metadata-field-selector.component.ts
VictorDuranEscire: @artlowel Hi, I changed this observable in the subscribe function

src/app/dso-shared/dso-edit-metadata/metadata-field-selector/metadata-field-selector.component.ts
VictorDuranEscire: @artlowel  I aded to subs array

src/app/dso-shared/dso-edit-metadata/metadata-field-selector/metadata-field-selector.component.html
VictorDuranEscire: @tdonohue. I merged the main branch with this branch and changed the "ds-loading" import. Thanks for the feedback

artlowel (approved):
Thanks @VictorHugoDuranS!

PR #3149, opened 2024-06-26
missing German translations (500 error)

saschaszott:
## Description

This PR adds 3 missing German translations related to the 500 error page.


pnbecker (approved):
Danke!

PR #3234, opened 2024-08-07
WorkspaceitemsEditPageModule: remove duplicated delete route

saschaszott:
## Description

This PR removes a duplicated `/delete` route in `WorkspaceitemsEditPageModule`.


tdonohue (approved):
👍 Thanks @saschaszott !  This is obviously correct.  This exact route code appears twice in the file.  Porting also to 8.x as the duplication appears there as well.

PR #3236, opened 2024-08-07
#3094 - Update date slider to prevent automatic page reloads

Simone-Ramundi:
## References
* #3094 

## Description
* Updated the slider implementation in `SearchRangeFilterComponent` to comply with WCAG 2.1 Success Criterion 3.2.2 (On Input). This improvement ensures that slider value changes no longer trigger automatic page reloads, enhancing user control over when context changes occur.

## Instructions for Reviewers
List of changes in this PR:
* **Updated Slider Event Handling**: Replaced `(onDebounce)="onSubmit()"` with `(change)="onSliderChange($event)"` to prevent automatic page reloads when slider values change. This ensures users have control over when the page context updates.
* **Refactored Form Submission**: Decoupled slider value updates from form submission, allowing users to submit changes only when they choose, thus preventing unintended context changes.
* **Enhanced User Control**: By separating slider changes from form submission, users can manage when and how context changes occur, aligning with WCAG 2.1 guidelines for user control and context management.

**To Reproduce**

1. Visit the search page on the sandbox, demo, or localhost (e.g., .../search?query=).
2. Expand the date filter to reveal the date slider.
3. Adjust the date slider to filter search results.

* **Expected Behavior**: The page should not reload automatically when the date slider is adjusted. The page should only reload when the submit button is clicked after adjusting the slider.
* **Actual Behavior (before changes)**: The page would reload automatically each time the date slider was adjusted, which interfered with the user's ability to refine search results smoothly.

## Checklist
- [X] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [X] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [X] My PR **passes [ESLint](https://eslint.org/)** validation using `yarn lint`
- [X] My PR **doesn't introduce circular dependencies** (verified via `yarn check-circ-deps`)
- [X] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [X] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [X] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC8x/Accessibility)** if it makes changes to the user interface.
[...]

PR #3245, opened 2024-08-09
[Port dspace-7_x] #3094 - Update date slider to prevent automatic page reloads

alanorth:
Manual port of https://github.com/DSpace/dspace-angular/pull/3236 by @Simone-Ramundi to dspace-7_x.


tdonohue (approved):
👍 Verified via code review that this provides the same changes as #3236.  Thanks for the port, @alanorth !

Issue #3259, opened 2024-08-22
Typo in German translation


dietmarkammerer:
https://github.com/DSpace/dspace-angular/blob/ee7e8358e7e0fd0b69f7c2e308d539670a103026/src/assets/i18n/de.json5#L3453

"Beschreibeung" should be "Beschreibung"

vNovski:
Done 😉

PR #3277, opened 2024-08-29
Typo in German translation

vNovski:
## References
* Fixes https://github.com/DSpace/dspace-angular/issues/3259

## Description
A Typo has been fixed  in the German translation


tdonohue (approved):
👍 Thanks @vNovski !  Looks good, merging immediately

PR #3251, opened 2024-08-13
Minor accessibibity fixes

alexandrevryghem:
## Description
These are minor fixes that were required in order to fix some issues regarding accessibility on the item page & the browse controlled vocabulary page.

## Instructions for Reviewers
In order to detect these issues we used [Silktide](https://silktide.com/)

List of changes in this PR:
* Fixed issues where the search tabs on the item pages didn't have the correct roles (only an issue when multiple tabs are available)
* Removed id `browseDropdown` from `ExpandableNavbarSectionComponent`, since it's not used anywhere an that component is also not meant to be browse specific. This caused a duplicate id issue when you used 2 expandable secitons in your navbar
* Fixed missing label for the search input field on the browse by controlled vocabulary pages

## Checklist
- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `yarn lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `yarn check-circ-deps`)
- [x] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [x] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [x] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC8x/Accessibility)** if it makes changes to the user interface.
- [x] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
- [x] My PR **includes details on how to test it**. I've provided clear instructions to reviewers on how to successfully test this fix or feature.
- [x] If my PR includes new libraries/dependencies (in `package.json`), I've made sure their licenses align with the [DSpace BSD License](https://github.com/DSpace/DSpace/blob/main/LICENSE) based on the [Licensing of Contributions](https://wiki.lyrasis.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-LicensingofContributions) documentation.
- [x] If my PR includes new features or configurations, I've provided basic technical documentation in the PR itself.
[...]

PR #3353, opened 2024-09-24
[Port dspace-8_x] Minor accessibility fixes

alexandrevryghem:
Port of #3251 to `dspace-8_x`.


kshepherd (approved):
+1 already merged to main and 7.x, merging immediately

PR #3368, opened 2024-09-29
[Port dspace-7_x] Fixed Edit Item's Version History crashing

alexandrevryghem:
Port of #3253 to `dspace-7_x`.


tdonohue (approved):
👍 Thanks @alexandrevryghem !  Tested this locally with a 7.x backend and it works well.

PR #3370, opened 2024-09-30
[Port dspace-8_x] Live region

AAwouters:
Port of #3337 to `dspace-8_x`.


tdonohue (approved):
👍 Thanks @AAwouters !  Verified this is identical to #3337.  Merging immediately

PR #3371, opened 2024-09-30
[Port dspace-7_x] Live region

AAwouters:
Port of #3337 to dspace-7_x.


tdonohue (approved):
👍 Thanks @AAwouters !  Verified this is the same code as #3337, but using the older format rules & no standalone components.  Merging immediately for 7.6.x

PR #3375, opened 2024-10-01
Always sanitize HTML passed to dsMarkdown even if markdown rendering is disabled

kshepherd:
## Description
The markdown directive currently sets `this.el.innerHTML` directly to `value` if markdown is disabled in config and forcePreview is not set. This means unsanitized HTML can end up making its way into the `innerHTML` of an element when somebody uses the `[dsMarkdown]=....` directive.

This small pull request ensures the `innerHTML` is always sanitized.

New tests are included to prove this behaviour (the template is initialised with `test<script>alert(1)</script>` and we check that in all cases, the div innerHTML is equal only to `test`.

## Instructions for Reviewers

* Try to inject unsafe HTML in a `[dsMarkdown]` directive (you can do this via metadata value, or as a raw string), with both `environment.markdown.enabled` true, and false. In both cases HTML should be correctly sanitized.
* You should notice that in main branch, disabling markdown allows HTML through unsanitized, and enabling it does sanitize the HTML correctly
* In this PR, the HTML should be sanitized in both cases
* Quick review of `markdown.directive.spec.ts` test coverage is appreciated

## Checklist

### DIFF
--- a/src/app/shared/utils/markdown.directive.spec.ts
+++ b/src/app/shared/utils/markdown.directive.spec.ts
@@ -8,21 +8,20 @@ import {
 } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 
+import { environment } from '../../../environments/environment.test';
 import { MathService } from '../../core/shared/math.service';
 import { MockMathService } from '../../core/shared/math.service.spec';
 import { MarkdownDirective } from './markdown.directive';
 
 @Component({
-  template: `<div dsMarkdown="test"></div>`,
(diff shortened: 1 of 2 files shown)


tdonohue (approved):
👍 Thanks @kshepherd !  This looks correct to me and the tests prove it works.  The Codecov failure in this PR seems to be a false positive...it says this PR has 100% coverage, but says the coverage somehow drops overall (which makes no sense).   In any case, merging immediately and backporting to 8.x (this doesn't appear to impact 7.x which has uses a markdown pipe instead of a directive).

PR #3465, opened 2024-10-17
[Port dspace-8_x] Edit Item, Bitstreams tab: Accessibility improvements

AAwouters:
Port of #3464 to `dspace-8_x`.

### DIFF
--- a/src/app/item-page/edit-item-page/item-bitstreams/item-bitstreams.component.html
+++ b/src/app/item-page/edit-item-page/item-bitstreams/item-bitstreams.component.html
@@ -31,7 +31,7 @@
     </button>
   </div>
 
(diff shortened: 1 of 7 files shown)


kanasznagyzoltan:
We have tested:
- created an intem with 12 bitstreams in 2 separate bundles
- changing the page size can be done separatly for each bundle

image

- dragging an item makes to help tooltip pop up

image

Result: 
- moving the bitstream on the page number correctly moves it to the next page
- moving the bitstream with the arrow keys also works correctly

tdonohue (approved):
👍 Thanks @AAwouters !  I verified this is an accurate port, and gave it a quick test on the `dspace-8_x` branch. It looks to work well.

PR #3466, opened 2024-10-17
[Port dspace-7_x] Edit Item, Bitstreams tab: Accessibility improvements

AAwouters:
Port of #3464 to dspace-7_x.


tdonohue (approved):
👍 Thanks @AAwouters !  Gave this a quick test against `dspace-7_x` and it works the same as the `main` version. Code looks the same too

PR #3540, opened 2024-10-23
Geospatial maps for item pages, search, browse

kshepherd:
## References
* Requires DSpace/DSpace#9918

## Description
This pull request is introduced as part of a wider contribution (GND authority and external data) but is a standalone feature which has a wide application, so is opened as a separate PR.

It allows maps to be drawn on item pages, a new geo-based search results view mode, and a 'browse by geolocation' browse map.

Full usage and technical documentation of this new feature is at https://github.com/kshepherd/dspace-geospatial-maps-doc/blob/main/README.md

demo map

## Instructions for Reviewers

List of changes in this PR:
* New base and item field components (example usage commented out in untyped item and publication component templates) for drawing maps from metadata values or other input
* New configuration properties to support this feature
* A few new i18n keys
* Marker assets for leaflet maps

To prepare test data, populate some item metadata with `dcterms.spatial` WKT point values (as per [documentation](https://github.com/kshepherd/dspace-geospatial-maps-doc/blob/main/README.md)).

For example, `Point ( +008.660000 +047.336111 )` is a WKT text value to set in `dcterms.spatial`

To test item page maps, try out the item page field components that are commented out in untyped-item and publication templates. This can be tested without the [backend PR](DSpace#9918).

To test the search results map, enable it in angular config and look for a new 'map' view mode beside the list and grid view mode icons. This can be tested without the [backend PR](DSpace#9918), though I recommend applying it and reindexing so you can try out the "has geospatial metadata" sidebar facet.

To test the browse-by map, make sure to apply the [backend PR](DSpace#9918) (note- this requires the solr search core to be recreated and reindexed due to a new schema field, if you want to use the "has geospatial metadata" filter) and make sure you have the documented search configuration, filters and facets applied, then do a full-reindex and enable the browse map in angular config. It should appear in the browse menu and clicking markers should open a new search filtered by that point.

Unit tests are written for all new components.

## Checklist


EikLoe:
I tested the PR locally and it worked fine! Thanks @kshepherd !

package.json
```diff
@@ -120,6 +120,9 @@
     "@ngrx/store": "^17.1.1",
     "@ngx-translate/core": "^14.0.0",
     "@nicky-lenaers/ngx-scroll-to": "^14.0.0",
+    "@terraformer/wkt": "^2.2.1",
+    "@types/grecaptcha": "^3.0.4",
+    "angular-idle-preload": "3.0.0",
```
tdonohue, line 125: I think we removed `@types/grecaptcha` and `angular-idle-preload` a while back from `main` as they were unused.  I don't think they are used here either (unless I missed it), so they can be removed from this PR

package.json
```diff
@@ -145,6 +148,10 @@
     "json5": "^2.2.3",
     "jsonschema": "1.5.0",
     "jwt-decode": "^3.1.2",
+    "klaro": "^0.7.18",
```
tdonohue, line 151: Looks like you are also accidentally adding Klaro back in.  This has been replaced by Orejime on `main`.  So, this dependency can be removed.

src/app/item-page/simple/item-types/publication/publication.component.html
```diff
@@ -100,6 +100,15 @@
...
     </ds-item-page-uri-field>
+    <!-- Below is an example of how to render one or more lat/lng points and/or bounding box rectangles
+         in a tiled map viewer. Set 'cluster' to true for marker clustering -->
+<!--    <ds-geospatial-item-page-field [item]="object"-->
+<!--                                   [label]="'item.page.places'"-->
+<!--                                   [pointFields]="['dcterms.spatial']"-->
+<!--                                   [bboxFields]="['external.spatial.bbox']"-->
+<!--                                   [cluster]="true"-->
+<!--    >-->
+<!--    </ds-geospatial-item-page-field>-->
```
tdonohue, line 111: Should this commented out code be removed?  Or are you supposed to comment it out in order to use it?

If the expectation is you have to uncomment it, then I'd prefer if we had a way to enable this dynamically via a configuration in your `config.yml`.  That would allow you to more easily turn this feature "on" or "off" in the same way that we do for other features on the Item pages (e.g. mediaviewer, iiif, etc)

src/app/item-page/simple/item-types/untyped-item/untyped-item.component.html
```diff
@@ -65,6 +65,15 @@
...
     </ds-generic-item-page-field>
+    <!-- Below is an example of how to render one or more lat/lng points and/or bounding box rectangles
+         in a tiled map viewer. Set 'cluster' to true for marker clustering -->
+<!--    <ds-geospatial-item-page-field [item]="object"-->
+<!--                                   [label]="'item.page.places'"-->
+<!--                                   [pointFields]="['dcterms.spatial']"-->
+<!--                                   [bboxFields]="['external.spatial.bbox']"-->
+<!--                                   [cluster]="true"-->
+<!--                                >-->
+<!--    </ds-geospatial-item-page-field>-->
```
tdonohue, line 76: Same comment as above.  I'd prefer if this was more dynamic, so you don't require people to modify the HTML code in order to enable it.  But, maybe I'm misunderstanding this?

src/app/item-page/simple/metadata-representation-list/metadata-representation-list.component.ts
```diff
@@ -97,14 +98,17 @@ export class MetadataRepresentationListComponent extends AbstractIncrementalList
...
     return observableZip(
       ...metadata
         .slice((this.objects.length * this.incrementBy), (this.objects.length * this.incrementBy) + this.incrementBy)
         .map((metadatum: any) => Object.assign(new MetadataValue(), metadatum))
         .map((metadatum: MetadataValue) => {
-          if (this.metadataService.isVirtual(metadatum)) {
+          if (this.metadataService.isVirtual(metadatum) && !metadatum.authority.includes('http')) {
             return this.relationshipService.resolveMetadataRepresentation(metadatum, this.parentItem, this.itemType);
+          } else if (this.metadataService.isVirtual(metadatum) && metadatum.authority.includes('http')) {
+            // TODO: we could do authority virtual handling here?
```
tdonohue, line 110: I'm not sure I understand the meaning of this TODO.  Is it necessary, or should we instead add details on why this is returning an empty result when `http` is included?

src/app/shared/geospatial-map/geospatial-map.component.scss
```diff
@@ -0,0 +1,61 @@
...
+}
+
+.map-frame {
+  border: 2px solid black;
+  height: 100%;
+}
+
+//.map {
+//  height: 100%;
+//}
```
tdonohue, line 43: We should likely just remove any commented out code.

src/app/shared/object-geospatial-map/object-geospatial-map.component.ts
```diff
@@ -0,0 +1,118 @@
...
+  encapsulation: ViewEncapsulation.Emulated,
+  selector: 'ds-object-geospatial-map',
+  styleUrls: ['./object-geospatial-map.component.scss'],
+  templateUrl: './object-geospatial-map.component.html',
+  animations: [fadeIn],
+  standalone: true,
+  imports: [ GeospatialMapComponent, NgIf ],
+})
+
+export class ObjectGeospatialMapComponent {
```
tdonohue, line 39: Please add a description to this component in a comment or TypeDocs.

tdonohue (changes requested):
@kshepherd : Thanks for this! I gave this a really quick initial code review today, just looking for anything obvious.  Overall, the code looks good, but there's some flaws in your `package.json` (which might be why this PR is failing in the `npm install` step -- I cannot seem to test it).  Other than that, a few more things that I noticed inline below.

I also have a general question (which might be clarified when I test it)...how "accessible" is this feature?  We may want to run an accessibility scanner (e.g. WAVE or axe DevTools in Chrome) against it to make sure no errors result.  If there are errors. we'd want to either try to address them, or perhaps add some warnings to the documentation if there are known accessibility issues (and create tickets to address them later).

tdonohue:

src/app/item-page/simple/item-types/untyped-item/untyped-item.component.html
kshepherd: (answered above, tbc)

src/app/shared/geospatial-map/geospatial-map.component.scss
kshepherd: done

src/app/item-page/simple/metadata-representation-list/metadata-representation-list.component.ts
kshepherd: i have reverted the changes to this file for now as I believe this is related to virtual authority metadata / other GND functionality, not pure geospatial data rendering, good spotting

src/app/shared/object-geospatial-map/object-geospatial-map.component.ts
kshepherd: done

package.json
kshepherd: done

package.json
kshepherd: done

src/app/browse-by/browse-by-geospatial-data/browse-by-geospatial-data.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      getFirstSucceededRemoteDataPayload(),
+      filter((searchFilterConfigs) => hasValue(searchFilterConfigs)),
+      take(1),
+      map((searchFilterConfigs) => searchFilterConfigs[0]),
+      filter((searchFilterConfig) => hasValue(searchFilterConfig))),
+    ],
+    ).pipe(
+      switchMap(([scope, searchFilterConfig]) => {
+        // Get all points in one page, if possible
+        searchFilterConfig.pageSize = 99999;
```
tdonohue, line 97: This seems like a potential performance issue.  Do we really want to plot everything on a single map?

src/app/browse-by/browse-by-geospatial-data/browse-by-geospatial-data.component.ts
```diff
@@ -0,0 +1,110 @@
...
+        const searchOptions: PaginatedSearchOptions = Object.assign({
+          'configuration': environment.geospatialMapViewer.spatialFacetDiscoveryConfiguration,
+          'scope': scope,
+          'facetLimit': 99999,
```
tdonohue, line 101: Same here, isn't this a possible performance issue? We just recently tried to remove these sort of `99999` references from components in #2513

src/app/item-page/simple/item-types/publication/publication.component.html
tdonohue: Keep in mind, if there's no way to "enable" this via configuration then it won't be enabled on our Demo site (demo.dspace.org) and will therefore be *difficult* to support / debug bug tickets/issues (as developers will have to manually enable it to test that the component is working properly). 

It also means that we can never create e2e tests related to this feature, as we cannot modify the HTML in an e2e test.  While we don't require e2e tests at this time, there is a goal to build more of them in #3978

I still find this approach problematic for those reasons.  But, that said, I won't block this PR because of this approach.  It just seems like it's not easy to setup.

src/app/item-page/simple/field-components/specific-field/geospatial/geospatial-item-page-field.component.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * List of fields to parse for WKT points
+   */
+  @Input() pointFields = ['dcterms.spatial'];
+
+  /**
+   * List of fields to parse for bounding box GeoJSON
+   */
+  @Input() bboxFields = ['gnd.coordinates.bbox'];
```
tdonohue, line 54: This field doesn't exist in DSpace?  Is it a metadata field, or something else?

src/app/item-page/simple/item-types/publication/publication.component.html
```diff
@@ -106,6 +106,15 @@
       [fields]="['datacite.relation.isReferencedBy']"
       [label]="'item.page.referenced'">
     </ds-item-page-uri-field>
+    <!-- Below is an example of how to render one or more lat/lng points and/or bounding box rectangles
+         in a tiled map viewer. Set 'cluster' to true for marker clustering -->
+<!--    <ds-geospatial-item-page-field [item]="object"-->
+<!--                                   [label]="'item.page.places'"-->
+<!--                                   [pointFields]="['dcterms.spatial']"-->
+<!--                                   [bboxFields]="['external.spatial.bbox']"-->
```
tdonohue, line 114: Is this a typo?  In the component you have this `bboxFields` set to `gnd.coordinates.bbox` by default.  But, as noted above, it doesn't exist in DSpace

src/app/shared/geospatial-map/models/geospatial-map-detail.model.ts
```diff
@@ -0,0 +1,7 @@
+import { GeospatialMarkerDetail } from './geospatial-marker-detail.model';
+
+export class GeospatialMapDetail {
```
tdonohue, line 3: Please describe this class & its properties in code comments

src/app/shared/geospatial-map/models/geospatial-marker-detail.model.ts
```diff
@@ -0,0 +1,6 @@
+export class GeospatialMarkerDetail {
```
tdonohue, line 1: Please describe this class & its properties in code comments

tdonohue (changes requested):
@kshepherd : I gave this a code review today, as I'm not sure I understand how to test this.  I honestly don't understand how to set `dcterms.spatial` (I'd prefer some examples, as I have no sample geospatial data to work with). The [documentation you've created](https://github.com/kshepherd/dspace-geospatial-maps-doc/blob/main/README.md) shows a lot of nice visuals, but with no hints on how that visual was created

I'm also slightly concerned about the performance of parts of this feature (it disables all pagination for some requests to the backend), and I have a few other notes inline below.

Overall, I'm not against this feature, but I'm really struggling to understand how to use it.  I don't have a ton of time available to figure it out, so I'm hoping either you can give me some hints or someone else (maybe @vins01-4science if he finds time) can help test it.

src/app/browse-by/browse-by-geospatial-data/browse-by-geospatial-data.component.ts
pnbecker: Well, you cannot paginate a map. So if you want to browse by a map, we will have to show all points. Also have in mind, that it only loads the geo coordinates, not the whole items... I don't think performance would be better, if we do pagination in the REST API here and have a for loop running over all pages in angular.

pnbecker:
@tdonohue As an example, you can set dcterms.spatial to `Point ( +008.660000 +047.336111 )`. So basically use the geocordinates.

src/app/browse-by/browse-by-geospatial-data/browse-by-geospatial-data.component.ts
tdonohue: @pnbecker & @kshepherd : I see, I guess I had assumed this would be similar to the "search" function that I see a screenshot of here: https://github.com/kshepherd/dspace-geospatial-maps-doc/blob/main/search-results-map.png

Because I wasn't able to test this, my expectation was that you'd still see the list of Items *under* the map.  In other words, I was expecting that the map should only plot the *current page* of items.  Then, if you change to the next page, then the map would update.

If this browse by geolocation **requires** no pagination, then this code can be left as-is.  However, I feel strongly that we MUST put a warning on this feature.  If the response it sends to the backend will return thousands of Items, then there's a risk that this feature will be very slow to load (simply because the backend will be very slow in returning thousands of items).

This is one advantage to actually doing the pagination in Angular.  If you did the pagination in Angular, you would be able to plot the first 10 items, then send the next request, etc.  But, again, the performance could eventually get bad if you are plotting thousands of items on a single map.

Another option is to only plot the first 50-100 items at a time and make it clear to the user that they must zoom in on an area of the map to get better results.

tdonohue: I meant to add, if you've tested this with thousands of geolocations/items at once and not found issues, then I'm OK with this as-is.  The current approach just *looks very similar to code we've seen in the past which had major performance issues*

kshepherd: Totally understand the concern, and I think a warning is well warranted. In this case, I haven't seen performance issues with 1000s, but I may with 99,999 I suppose. (probably more about drawing/clustering markers than handling the discovery response)

If it helps, bear in mind these are facet values, not full rows of search object results, so it's a much smaller Solr / Discovery payload than an actual *search* - this is the *browse* map just to be clear. That means that not only are we just dealing with simple strings and counts, we are already at least clustering by point.

The *search* map view does paginate along with the items displayed on that page, to be clear.

In the future I'd love to scope the search with some geospatial-aware Solr functionality but it requires a bit more complexity, and still would suffer from a "search the entire world" query when a user is fully zoomed out.
We could also look into the future at clustering by *region* to lower the facet result count from a zoomed out search. But so far there hasn't been a need for this.

pnbecker:
@tdonohue you can see the browse map in production here: https://zop.zb.uzh.ch/browse/map

src/app/item-page/simple/item-types/publication/publication.component.html
kshepherd: i've changed this field default to dcterms.spatial

src/app/item-page/simple/field-components/specific-field/geospatial/geospatial-item-page-field.component.ts
kshepherd: updated to dcterms.spatial (and ensured the rendering can support bbox and points without conflict) since there is no other appropriate field to use as a default. We could also remove a default and force it to be determined only by user input if there is concern about `dcterms.spatial` usage, though the DCMI documentation appears to support this usage which is why I picked it.

src/app/browse-by/browse-by-geospatial-data/browse-by-geospatial-data.component.ts
kshepherd: see previous comment - a warning is a good idea, but in order to make a browse map usable without a lot of complex solr spatial features, we do need to be able to make large queries.
This is a facet query, the values are lists of facet string values and counts, not full indexable search objects as with a regular discovery search.
So I hope this is good enough for an initial implementation

kshepherd:
@tdonohue this should now be ready, all tests are passing. I also updated my documentation at https://github.com/kshepherd/dspace-geospatial-maps-doc/blob/main/README.md to include example formats for bounding box and point data, and the new config properties

src/app/shared/view-mode-switch/view-mode-switch.component.html
```diff
@@ -40,5 +40,20 @@
...
+         [queryParams]="{view: 'geospatialMap'}"
+         queryParamsHandling="merge"
+         (click)="switchViewTo(viewModeEnum.GeospatialMap)"
+         routerLinkActive="active"
+         [class.active]="currentMode === viewModeEnum.GeospatialMap"
+         class="btn btn-secondary"
+         [attr.data-test]="'geospatial-view' | dsBrowserOnly">
+         <span class="fas fa-map"></span><span class="sr-only">{{'search.view-switch.show-geospatialMap' | translate}}</span>
+      </button>
+    }
```
toniprieto, line 57: The map icon is not being displayed. The condition is inside "`@if (isToShow(viewModeEnum.DetailedListElement)) {`", but it should be after the` }` on line 58.

src/app/item-page/simple/item-types/shared/item.component.ts
```diff
@@ -75,9 +75,16 @@ export class ItemComponent implements OnInit {
...
+  /**
+   * Enables display of geospatial item page fields
+   */
+  geospatialItemPageFieldsEnabled = false;
+
   constructor(protected routeService: RouteService,
               protected router: Router) {
     this.mediaViewer = environment.mediaViewer;
+    this.geospatialItemPageFieldsEnabled = environment.geospatialMapViewer.enableItemPageFields;
+    console.log(this.geospatialItemPageFieldsEnabled);
```
toniprieto, line 87: This could be removed.

toniprieto (commented):
Hi @kshepherd , thanks for this contribution. In our institution, we had a similar functionality in previous versions, and it will be fantastic to have it in new versions! 
I am testing the functionality with several records and hope to send a review soon.

I’m also sharing a CSV with 1,000 fake records containing geolocated points in case it might be useful for other tests: https://github.com/toniprieto/test-import-metadata-dspace/blob/main/metadata-import-test-geospatial-1000-new.csv

For now, I’m adding a couple of comments.

toniprieto:
I have tested all the functionalities: displaying the map in search (after fixing the previously mentioned error), the new facet, displaying the map in an item, and the browse-by-map functionality. Everything has worked well for me. I'm sharing an example of a map with multiple points and a bounding box.

image

The only issue I found is that the browse-by-map functionality only displays 100 points. From what I’ve seen, the number of results returned by a facet is limited to 100. More results can be added by configuring the `rest.search.max.results `option (but at most 1000, as I believe the API pagination is limited to 1000 items).
This configuration key is used here: [DiscoverQueryBuilder.java](https://github.com/DSpace/DSpace/blob/main/dspace-api/src/main/java/org/dspace/discovery/utils/DiscoverQueryBuilder.java#L62).

Since the browse-by-map functionality can be disabled, I think documenting this limitation should be sufficient.

kshepherd:
Thanks @toniprieto ! I have addressed the view mode button control flow as you noted, and removed the console output.
I agree with you that we should document the implication of `rest.search.max.results` on the browse map, but leave it up to the user to change (since the feature is disabled by default anyway)

tdonohue (commented):
@kshepherd : Thanks for all your hard work on this.  This is looking *much better* to me.  Also a huge thanks to @toniprieto for the [test data set](https://github.com/toniprieto/test-import-metadata-dspace/blob/main/metadata-import-test-geospatial-1000-new.csv), which I used in order to see how this performs with larger amounts of data.

Overall, I've successfully tested the following:
* `enableItemPageFields: true` - Verified this can now be used to more easily display the "Related Places" map on an individual Item or Publication page.  Works well!
* `enableSearchViewMode: true` - Verified this shows a new "Map display" on the Search Results page (and Admin Search Results).  That display looks to work well. However one small issue I noticed:
    * **Issue?** If no geolocational data is available on the current page, then it defaults to a map of London (I expect this may just be the defaults of the library you are using?).  This isn't a bug though, it just looked odd to me because it's zoomed in on London as if that's what's important on the current page.  That said, I'm OK with this. Just pointing it out in case it isn't expected behavior.
* `enableBrowseMap: true` - Verified this works at a basic level, but I'm hitting a small bug (at least when running in Production mode (`npm run:build && npm run serve:ssr)
    * Click the "By Geolocation (Map)" option in the global browse by menu. Brings you to the large map (http://localhost:4000/browse/map)
    * Zoom in on a section of the map where there's at least one point displayed (it all works well)
    * Now, click on the point. (Here's where the bug occurs)
    * I'm sent to an empty search page, with that "POINT" as a facet.  The URL looks like this `http://localhost:4000/search?spc.page=1&f.point=POINT(-5.995402002455609%2037.388434867709165),equals`
    * However, the Search Results page remains empty and I see this error on the backend:
````
2025-03-28 22:05:05,301 ERROR 0d1c1077-2146-4e96-9306-0f503e20ce35 76d18306-dcab-4560-872e-1e4dd0860379 org.dspace.app.rest.exception.DSpaceApiExceptionControllerAdvice @ An exception has occurred (status:400)
org.dspace.app.rest.exception.DSpaceBadRequestException: point is not a valid search filter
   at org.dspace.app.rest.utils.RestDiscoverQueryBuilder.buildFacetQuery(RestDiscoverQueryBuilder.java:169) ~[dspace-server-webapp-9.0-SNAPSHOT.jar!/:9.0-SNAPSHOT]
   at org.dspace.app.rest.repository.DiscoveryRestRepository.getFacetObjects(DiscoveryRestRepository.java:143) ~[dspace-server-webapp-9.0-SNAPSHOT.jar!/:9.0-SNAPSHOT]
   at org.dspace.app.rest.DiscoveryRestController.getFacetValues(DiscoveryRestController.java:215) ~[dspace-server-webapp-9.0-SNAPSHOT.jar!/:9.0-SNAPSHOT]
````

Am I missing something on the backend?  I installed the backend PR.  I also ran a complete reindex (`./dspace index-discovery -b`).  Any ideas of what might be going on?

Overall, I'm nearly a +1.  I'm just trying to verify the behavior of the Browse Map, and I'm failing to do so.  Everything else is working well though.

kshepherd:
thanks @tdonohue for the extra testing. I'll look into the issues here, but regarding the search map, I think the discovery rebuild might not be quite enough, as we add a new field to the Solr schema. So you might need to make sure you copy the config over and perhaps even delete the core and allow it to be recreated with the new field, to avoid that "point is not a valid search filter" error - that is something we'll have to document as well

kshepherd:
If a map is displayed showing London for an item with no dcterms.spatial metadata then that seems like a bug to me - yes, it's a default point, but we should not display a map at all if no relevant data is present

toniprieto:
@tdonohue About the error "point is not a valid search filter," I also encountered it during the initial tests, but it was resolved by enabling the facet `<ref bean="searchFilterGeospatialPoint"/>` in the `discovery.xml` of the backend. It fails because the facet used by the search is not enabled.

toniprieto (approved):
Thanks @kshepherd ! Looks good to me! 

Just pointing out a minor issue I've noticed in the latest tests, the translation key is missing for: `search.filters.applied.f.point`

imagen

pnbecker:
@kshepherd if I understand @tdonohue correctly, London is shown if you switch on the map view on a list of search results without any geospatial item. It does not show a wrong marker, its just the default position of the map without markers. @tdonohue  did I understand you correctly?

tdonohue:
@pnbecker is correct. There’s no point displayed. It’s just that the default view of the map is a view of London. It’s not necessarily a bug, it just seemed odd to me. But if there’s no way to fix that, we can leave it as is.

________________________________

kshepherd:
I see. Here are some changes:

1. New "DEFAULT_CENTRE_POINT" configuration, documented in dfault-app-config.ts and config.example.yml - it uses Istanbul as a default value since it shows more of the (old) world nicely and isn't so "anglo centric default" ;)
2. New behaviour in browse and search map layer logic, where instead of using the default zoom level, it instead zooms out to level 1 and shows a tool-tip, with some i18n-controlled text that indicates there are no points to display
3. As per @toniprieto 's comment, i18n added for `search.filters.applied.f.point` (displayed if someone clicks a browse marker icon)
4. ^^^ I will also push some backend changes so that the geospatial point and "has point" filters are enabled in discovery.xml by default, which will further help enable the feature - since they are not included as facets by default, it won't affect the facets displayed "out of the box" but it will ensure that clicking browse markers use a valid filter without any other user action

I believe the combination of zoom level and tooltip should give better UX now, when there are no results. If you click next page on a search page and the new page has results, the map will behave normally, zoom in to the bounds of the furthest markers, etc.

screenshot_2025-03-30_16:31:17_selection
screenshot_2025-03-30_16:35:39_selection

tdonohue (approved):
👍 Thanks @kshepherd ! I finally got to retesting this today, and all my prior feedback is resolved.  I also like the new "no results on this page" display that you've added.  This all looks good to me now

tdonohue:
Merging with +2 approval.  This obviously needs the documentation for this feature moved to our DSpace 9.0 docs: https://wiki.lyrasis.org/display/DSDOC9x  So, flagging this as "needs documentation" until that is completed.  Thanks again @kshepherd !

PR #3614, opened 2024-11-06
[Port dspace-8_x] Accessibility settings page

AAwouters:
Port of #3613 to `dspace-8_x`


AAwouters:
@tdonohue I've fixed the SSR errors, both in this PR and in a separate, tiny, PR for 7_x: #4519

PR #3620, opened 2024-11-07
class `Metadata`: rename parameter names in JSDoc from `mapOrMaps` to `mdMapOrMaps`

saschaszott:
## Description

This trivial PR fixes incorrect parameter names in JSDoc.


tdonohue (approved):
👍 Thanks @saschaszott !  This is an obvious correction to TypeDocs only.  Merging immediately.

PR #3660, opened 2024-11-16
[Port dspace-7_x] Added support for caching embedded objects without a self link & null responses

alexandrevryghem:
Port of #3415 to `dspace-7_x`.

Also restored [these changes](https://github.com/DSpace/dspace-angular/commit/bd8659ac116b4780223c2fea833616990a59b1a3) that were accidentally dropped in #3405 causing the `item-edit.cy.ts` e2e tests to be more flaky again

### DIFF
--- a/cypress/e2e/item-edit.cy.ts
+++ b/cypress/e2e/item-edit.cy.ts
@@ -15,13 +15,11 @@ describe('Edit Item > Edit Metadata tab', () => {
   it('should pass accessibility tests', () => {
     cy.get('a[data-test="metadata"]').click();
 
+    // Our selected tab should be active
+    cy.get('a[data-test="metadata"]').should('have.class', 'active');
+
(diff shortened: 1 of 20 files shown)


alanorth:
Thanks @alexandrevryghem. I tested this on a DSpace 7.6.3-SNAPSHOT instance and found:

- Before: 66 requests, 210KiB transferred
- After: 56 requests, 201KiB transferred

This was an extreme case where none of the ten items on the search page had thumbnails. Total time taken was ~300ms less. I didn't notice any side effects in my brief testing of the search page.

tdonohue (approved):
👍 Thanks @alexandrevryghem !  Gave this a quick test and seems to be working for 7.x.  Thanks also for seemingly fixing the random failures in `item-edit.cy.ts` (as those were annoying me at times as well).

PR #3661, opened 2024-11-17
[Port dspace-8_x] Added support for caching embedded objects without a self link & null responses

alexandrevryghem:
Port of #3415 to `dspace-8_x`.


tdonohue (approved):
👍 Thanks @alexandrevryghem !  I've tested this manually and found no issues.  Also verified the code changes in this PR are identical to those in #3415

PR #3716, opened 2024-12-05
[Port dspace-7_x] Fix infinite loading on item pages and optimize menu resolver usage

alexandrevryghem:
Manual port of #3677 to `dspace-7_x` (the `BaseDataService` fixes are actually not required, but at least this way all supported DSpace versions work in the same way).

The differences with the `dspace-8_x` & `main` PRs are:
- The `menu` resolver is still present on the browse pages. This is because the browse pages are still used to represent the comcol pages on `dspace-7_x` to render the browse tabs (fixed in #2722 for DSpace 8.0)
- The correction type improvement is not present, since this is also a new feature since `dspace-8_x`

### DIFF
--- a/src/app/collection-page/collection-page-routing.module.ts
+++ b/src/app/collection-page/collection-page-routing.module.ts
@@ -55,7 +55,6 @@ import { CommunityBreadcrumbResolver } from '../core/breadcrumbs/community-bread
         resolve: {
           dso: CollectionPageResolver,
           breadcrumb: CollectionBreadcrumbResolver,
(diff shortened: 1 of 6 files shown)


alanorth:
Thanks @alexandrevryghem. It seems GitHub CI got stuck and the Node.js 18 / 20 jobs never started, so I'm going to close and re-open the pull request to trigger those.

tdonohue (approved):
👍 Thanks @alexandrevryghem !  Finally got back to testing this backport. It looks good and works well.

PR #3718, opened 2024-12-06
fix selector in OrgUnitSearchResultListSubmissionElementComponent compontent

saschaszott:
## Description

This minor PR fixes the value of the selector in the `Component` annotation.


nwoodward (approved):
👍 This fixes an obvious bug in the selector.

alanorth:
Thanks @saschaszott! Merging ASAP as this is an obviously correct and simple patch and is https://github.com/DSpace/dspace-angular/labels/1%20APPROVAL.

PR #3719, opened 2024-12-06
Remove nested span elements with duplicate content of parent element

saschaszott:
## Description

This PR removes redundant nested span elements in various template files. These span elements contained identical content with respect to their parent elements.


artlowel (approved):
Thanks @saschaszott!

While this works, and can be merged as is, would you mind also removing those text-muted classes inside the tooltips? That would be another a11y improvement

PR #3720, opened 2024-12-06
Embed the community/collection on the edit EPerson page

alexandrevryghem:
## References
* Fixes #3571
* Continuation of #3585
* `dspace-7_x` backport: https://github.com/DSpace/dspace-angular/pull/3721

## Description
Fixed issue where the edit EPerson page didn't show the comcol names when they weren't already cached. Also removed a template observable call from `ItemDetailPreviewComponent`.

## Instructions for Reviewers
List of changes in this PR:
* Embedded the `Group`'s linked `object` in the group overview of edit `EPerson`. Otherwise only already cached communities and collections would have a visible name

## Checklist
- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [x] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [x] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [x] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC8x/Accessibility)** if it makes changes to the user interface.
- [x] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
- [x] My PR **includes details on how to test it**. I've provided clear instructions to reviewers on how to successfully test this fix or feature.
- [x] If my PR includes new libraries/dependencies (in `package.json`), I've made sure their licenses align with the [DSpace BSD License](https://github.com/DSpace/DSpace/blob/main/LICENSE) based on the [Licensing of Contributions](https://wiki.lyrasis.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-LicensingofContributions) documentation.
- [x] If my PR includes new features or configurations, I've provided basic technical documentation in the PR itself.
- [x] If my PR fixes an issue ticket, I've [linked them together](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

PR #3721, opened 2024-12-06
[Port dspace-7_x] Removed observable function calls from template (part 1)

alexandrevryghem:
Manual port of #3585 & #3720 to `dspace-7_x`


tdonohue (approved):
👍 Thanks @alexandrevryghem !  Gave this a quick test against the `dspace-7_x` branch and it looks to work well.  Bugs are all fixed in 7.x as well.

PR #3753, opened 2024-12-16
Get rid of unnecessary and failing REST requests when navigating between different browse indexes 

jensvannerum:
## References
* Fixes #573 

## Description
- Destroys and reconstructs the browse component when the route changes so that any route change calls `ngOnInit` in the browse-by component, this way we can be sure that we only care about the first value of the route (necessary for our `take(1)` in subscription)
   - in the  `browse-by-page-component`the type is wrapped in an object so a change from one browse index to another browse index of the same type still causes a lifecycle trigger (since objects are compared by identity).
 
You could argue that performance wise we don't want to re-build the component every time but any performance gains that this would offer is lost by the multiple useless requests I think 

## Instructions for Reviewers
- Set this angular branch up against latest main / demo instance
- Open any browse index and switch to other browse indexes while having the 'Network' tab open in developer tools
- Notice no unnecessary requests are being sent and no failing requests are present either (compare to eg current demo.dspace.org)
- Verify browse indexes still work as before including pagination, filtering, ordering thumbnails, etc...

## Checklist
_This checklist provides a reminder of what we are going to look for when reviewing your PR. You do not need to complete this checklist prior creating your PR (draft PRs are always welcome).
However, reviewers may request that you complete any actions in this list if you have not done so. If you are unsure about an item in the checklist, don't hesitate to ask. We're here to help!_

- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [x] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [x] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [x] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC8x/Accessibility)** if it makes changes to the user interface.
- [x] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
[...]

PR #3788, opened 2024-12-30
[Port dspace-7_x] Get rid of unnecessary and failing REST requests when navigating between different browse indexes

jensvannerum:
This is a backport of [#3753](https://github.com/DSpace/dspace-angular/pull/3753) to 7.x


alanorth:
Thanks @jensvannerum. This is failing due to unused imports. Could be something that was added in DSpace 8+ that we don't need here.

alanorth (approved):

alanorth:
Thanks @jensvannerum. I have tested the patch on DSpace 7.6 and it fixes the bug. Seeing as this is a simple port of #3753 and it's passing all tests in CI and my manual test, I will merge this. I will squash the commits into one though.

PR #3792, opened 2025-01-03
README.md: update Node.js and NPM versions

alanorth:
Angular 17.x needs Node.js 18.x or 20.x, which come with NPM v10.

See: https://angular.dev/reference/versions
See: https://nodejs.org/en/about/previous-releases#looking-for-latest-release-of-a-version-branch


tdonohue (approved):
👍 Thanks @alanorth .  This is obviously correct, merging immediately.

PR #3815, opened 2025-01-09
[Port dspace-8_x] Update spanish translation

toniprieto:
Manual port of #3814 for dspace-8x


tdonohue (approved):
👍 Thanks @toniprieto !

PR #3816, opened 2025-01-09
[Port dspace-8_x] Update catalan translation

toniprieto:
Manual port of #3797 for dspace-8x


tdonohue (approved):
👍 Thanks @toniprieto !

PR #3819, opened 2025-01-10
Vocabulary preloadlevel fix

AAwouters:
## References
* Fixes #3784 

#### Backports
#3820
#3821 

## Description
Two changes are made:
Before initializing a treeview, an attempt is made to retrieve the `preloadLevel` value for the opened vocabulary.
If a value is found the preloadLevel is set to that value, otherwise 1 is used.

Whenever a node is initialized and that node is on a level lower than the preloadLevel, that node is immediately expanded.

## Instructions for Reviewers
- Set config for preLoadLevel, e.g. vocabulary.plugin.srsc.hierarchy.preloadLevel = 3 (values greater than 3 are not recommended for 'srsc')
- Open Browse 'By Subject Category' page (browse/srsc)
- The tree is immediately expanded up to 3 levels

## Checklist
- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [x] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [x] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [x] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC8x/Accessibility)** if it makes changes to the user interface.
- [x] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
- [x] My PR **includes details on how to test it**. I've provided clear instructions to reviewers on how to successfully test this fix or feature.
- [x] If my PR includes new libraries/dependencies (in `package.json`), I've made sure their licenses align with the [DSpace BSD License](https://github.com/DSpace/DSpace/blob/main/LICENSE) based on the [Licensing of Contributions](https://wiki.lyrasis.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-LicensingofContributions) documentation.
- [x] If my PR includes new features or configurations, I've provided basic technical documentation in the PR itself.
[...]

PR #3820, opened 2025-01-10
[Port dspace-8_x] Vocabulary preloadlevel fix

AAwouters:
Port of #3819 to dspace-8_x.


kanasznagyzoltan:
We have tested:
Set the vocabulary.plugin.srsc.hierarchy.preloadLevel to 3, opened /browse/srsc and looks like it works as intended
image

For comparison, this is what it looks like with preLoadLevel 1 and 2
preLoadLevel = 1
image

preLoadLevel = 2
image

tdonohue (approved):
👍 Thanks @AAwouters !  I tested this today and it fixes the bug.

I also found that this "preloadLevel" configuration is undocumented (along with a few other configs in "authority.cfg"), so I've added docs to 7.x/8.x/9.x.  For example: https://wiki.lyrasis.org/display/DSDOC9x/Authority+Control+of+Metadata+Values#AuthorityControlofMetadataValues-AdditionalconfigurationsforHierarchicalControlledVocabularies

PR #3821, opened 2025-01-10
[Port dspace-7_x] Vocabulary preloadlevel fix

AAwouters:
Port of #3819 to dspace-7_x.


tdonohue (approved):
👍 Thanks @AAwouters !  Tested this manually on a 7.x instance and it works like the others.

PR #3879, opened 2025-01-22
[Post dspace-7_x] Restrict Angular SSR to paths in the sitemap

alanorth:
Manual port of #3682 to DSpace 7.6.x.

There are minor differences due to a refactoring of the Universal preboot configuration to "SSR" in DSpace 8 and main.


tdonohue (approved):
👍 Thanks @alanorth .  Code looks correct & this passes all automated tests (including the test that verifies SSR is still working on `/home`)

PR #3889, opened 2025-01-24
Add new orejime cookie for correlation-id header

AndreaBarbasso:
## References
* Fixes part of #3335 

## Description
I added a new cookie to orejime's settings. This cookie relates to the X-CORRELATION-ID request header. If the cookie is not accepted, no X-CORRELATION-ID header will be set on requests.  
The cookie title and description are not final - I'm open to suggestions on how to make them more clear for non-technical users.

image

## Instructions for Reviewers
You can check any request done while the cookie has not been accepted, and it won't have the X-CORRELATION-ID header.  

image

If you then accept the cookie, you will see the X-CORRELATION-ID cookie back again.  

image

List of changes in this PR:
* Added a new cookie to orejime's settings;
* Added a check for that cookie before setting up the X-CORRELATION-ID header in the `LogInterceptor`.

## Checklist


AndreaBarbasso:
I'm fixing merge conflicts, then this issue should be ready (it was more or less blocked by #3963 - the functionality was there, but it could not be tested correctly because of the issues with the `optOut` property.).

AndreaBarbasso:
I also added a little piece of code to disable orejime during e2e testing:
```
        if (this._window?.nativeWindow?.Cypress) {
          this.orejimeConfig.apps = [];
        } else {
          this.orejimeConfig.apps = this.filterConfigApps(appsToHide);
        }
 ```
This removes the need to update the cookie object in cypress' `e2e.ts` file, and also makes it work with cookies that are not yet approved for the test user (the correlation-id is one of them).

File: src/app/shared/cookies/browser-orejime.service.ts [hunk 1]
```diff
@@ -191,7 +191,11 @@ export class BrowserOrejimeService extends OrejimeService {
          */
         this.translateConfiguration();
 
-        this.orejimeConfig.apps = this.filterConfigApps(appsToHide);
+        if (this._window?.nativeWindow?.Cypress) {
```
tdonohue, line 194: I'm not a fan of adding Cypress specific code into our main codebase (i.e. the `./src/` folder).  Is there no way to do this in the Cypress code under `./cypress/`?

I just feel this makes our code messy if we start to check whether Cypress is active or not in our codebase -- it also complicates our ability to move to a different e2e testing system in the future (if we ever need to do so), because we'll have to find everywhere that Cypress is now referenced in our code.  In other words, I'd hate to see this sort of `if (Cypress)` check repeated all over our codebase.

An alternative might be to find a way to just disable Orejime altogether in Cypress (maybe via a passed in param or a configuration that we could set from Cypress).

src/assets/i18n/en.json5
```diff
@@ -1640,6 +1640,10 @@
 
   "cookies.consent.app.description.authentication": "Required for signing you in",
 
+  "cookies.consent.app.title.correlation-id": "Correlation ID",
+
+  "cookies.consent.app.description.correlation-id": "Allow us to track your session for debugging purposes",
```
tdonohue, line 1645: I'd recommend a slightly difference wording for this text in order to make it clear that this is for backend logs.  Maybe something like this:
```
"Allow us to track your session in backend logs for support/debugging purposes"
```

tdonohue (changes requested):
@AndreaBarbasso : I tested this and reviewed this today. Overall, it looks good, but I noticed a few things that need fixing:

* First, if I change my Cookie Settings to **disable** the Correlation ID, the `CORRELATION-ID` cookie is still created.  I expected the cookie to be gone if I don't allow for Correlation ID tracking.  I *can* verify though that the value in the `CORRELATION-ID` cookie is correctly not sent to the backend. 
* Second, I'm not a fan of adding checks into our main codebase for whether Cypress is running.  I feel this is "messy code".   In the backend codebase we don't embed checks for whether code is running as part of an integration test or unit test.  Similarly, I don't think we should embed checks to see if Cypress is running.  A few possible ideas inline below.

Overall, this looks good though & works at the basic level (other than the cookie wrongly being created)

src/app/shared/cookies/browser-orejime.service.ts
tdonohue: @AndreaBarbasso and @atarix83 : I've just realized this same issue with Orejime in e2e tests may be appearing in #3613, where some solutions are offered.  So, we may want to consider merging #3613 first (if one of you could help review/test it that'd help), as it may allow for a better way to solve this issue with Cypress e2e tests not working well with Orejime.

AndreaBarbasso: @tdonohue, I think the suggestion made on #3613 (adding all necessary accepted cookies to the test user in the e2e docker db) is the best solution, even if it will need continuous maintenance (i.e. adding each new non-mandatory cookie to the ones the user accepted). This way, there is no need for the action created in #3613 to accept orejime cookies, and  we can fix #4106 with no backside.  
If then we want to test the orejime cookies banner, we might want to create a new test user that has no cookies accepted.

tdonohue: @AndreaBarbasso : As clarified in today's Developers meeting, adding information to the e2e docker database is not a simple task (it requires updating the database dump in the [demo-entites-data test data set](https://github.com/DSpace-Labs/AIP-Files/releases/tag/demo-entities-data)) While it's possible to do so, I feel that approach is painful for the long term, as it would require pre-updating the e2e docker database anytime a new PR needs to add information to the orejime cookie.

I'd recommend we look at potentially merging #3613 **first** as it is currently passing e2e tests without having to update the e2e docker database.  If we are lucky, maybe we'll find the approach taken in #3613 will also work for this PR... if so, that would mean there's no need to update the e2e docker database.

AndreaBarbasso: @tdonohue, the problem with merging #3613 first is that, as soon as #4106 is fixed, e2e tests will no longer work as expected. The fix used in #3613 to make tests pass is actually a workaround that relies on the fact that we have the #4106 bug.

What we have now is:
- #4106, yet to be fixed, which is (IMHO) a critical bug, even more critical as soon as we introduce non-required cookies;
- #3613, which implemented a workaround for another issue (i.e. we have cookies that are not accepted on the Test User profile, and so the Orejime popup shows up after logging in) by clicking on the Orejime consent banner after each login. This is working now because #4106 is yet to be fixed, but as soon as it is fixed it will result in e2e tests failing, because after the first click on Orejime's banner we are going to save the cookie preferences on the User profile, and so the banner will no longer appear. This will make the `cy.get` of the banner fail consistently. @artlowel suggested to implement a way to check if the banner is rendered and - only if it is - to click on it, but this is not easy to do properly and, above all, [strongly not recommended](https://docs.cypress.io/app/guides/conditional-testing#Element-existence) by Cypress;
- #3889 (this issue), that tried another approach on solving the Orejime banner issue, which I see why is not considered as the best possible approach.

The problem I see in merging #3613 first is that we are merging a workaround that will fail as soon as #4106 is fixed.
So, I think that the correct line of fixes and merges is:
1. Fix and merge #4106 first (since it does not change how tests should be handled);
2. Then, find a way to solve the banner issue here, in #3889, and merge it;
3. Finally, revert the workaround on #3613 and merge it.

---

Now, about "how do we solve the banner issue", I can see a few options:
- We keep the check for `window.Cypress`. I see that this is kinda getting the code dirtier, but it's an approach [suggested by Cypress](https://docs.cypress.io/app/faq#Is-there-any-way-to-detect-if-my-app-is-running-under-Cypress) and I think it will not be used a lot around the code;
- We can mimic the solution above and keep it cleaner by checking for an environment variable (e.g. `environment.isTestEnvironment`) instead of `window.Cypress`. The problem is that we should either use the current test environment configuration file or create a new one, and make sure that current e2e tests will not fail, since they are now performed after a `yarn run serve:ssr`, which uses the default environment configuration file;
- Finally, we can change the Test User configuration on the e2e docker db, but this is time-consuming and it is not a solution that can be easily maintained.

@tdonohue, @artlowel, what are your thoughts on this?

tdonohue: @AndreaBarbasso : thanks for providing more information here.  I'm not against your solutions to solve the banner issues, but I'd prefer the *second one* (checking for an environment variable) rather than hardcoding references to `window.Cypress` in our code.

Regarding the order of merger, do we have a solution for #4106 yet?  It would be good to move that forward quickly if you feel that is the highest priority.. as I would hate to "block" the merger of these other related PRs for very long

AndreaBarbasso: @tdonohue, I opened a PR for #4106. If we manage to move that forward, I can make the requested changes on this PR and move it forward too!

tdonohue:
@AndreaBarbasso : When you have a chance, please rebase this PR now that #4225 is merged.  Thanks!

tdonohue:
@AndreaBarbasso : Reminder that I need your help getting this PR ready for review again.  Currently, I was waiting on this PR before I review #3613...but, if we cannot more it forward, then I may need to move forward #3613 *first*.  Please get this rebased & conflicts resolved as soon as you are able.

AndreaBarbasso:
@tdonohue, I added a commit that changes the way we check for running e2e tests.

pilasou:
HI @AndreaBarbasso, hi @tdonohue, thanks. I have tested the new cookie for Correlation ID using Docker, it worked as discribed. The new set up page for tracking contains a section to enable/disable Correlation ID. 
image

If enable, the Correlation ID is generated
image


If disable, it is not: 
image

src/environments/environment.production-e2e.ts
```diff
@@ -0,0 +1,10 @@
+import { BuildConfig } from '../config/build-config.interface';
+import { environment as prodEnvironment } from './environment.production';
+
+export const environment: Partial<BuildConfig> = Object.assign(
+  {},
+  prodEnvironment,
+  {
+    isE2E: true,
```
tdonohue, line 8: I could be wrong, but I feel like we could remove all of this new `environment.production-e2e.ts` stuff and simply set this flag via an environment variable in `package.json` like this:
```
"e2e" : "cross-env NODE_ENV=production DSPACE_E2ETEST=true ng e2e"
```

Then, check for `if (environment.e2eTest) {` in the `browser-orejime.service.ts`.  It should be "true" if that environment variable is set.  This should work because of how [environment variables can be used to override our settings](https://github.com/DSpace/dspace-angular/?tab=readme-ov-file#configuring).

Assuming that works, it'd have a few distinct advantages.  We'd no longer need to duplicate all the production settings in `package.json` and `angular.json`.  That means it'd also be *much easier to maintain* because we would not have to remember to keep the settings for production & production-e2e "in sync" in `angular.json`.

tdonohue (commented):
@AndreaBarbasso : Thanks for the updates.  Overall, this looks good & it works as described.  

But, in looking at the way you've set the "isE2E" variable/flag (to disable orejime for e2e tests), I cannot help but feel there may be a much easier solution.  I'm not *against* your approach, but I wanted to suggest something that may be much easier to maintain.  See the note inline below.

tdonohue:
@AndreaBarbasso : Just a reminder to take a look at my feedback from Friday, when you find the chance.  I *think* it would really simplify this PR, and let us get this finally merged.  (But, please make sure to try it on your end to verify)

src/environments/environment.production-e2e.ts
AndreaBarbasso: @tdonohue, I tried to do as you suggested, but it does not seem to work the way it should.

I applied that strategy to the usual `build:prod` in order to test it easily:
`"build:prod": "cross-env NODE_ENV=production DSPACE_E2ETEST=true npm run build:ssr",`

but then. in the `config.json` file generated in the `dist` folder I can't find the environment variable set up via the script in the `package.json` file. This is confirmed after starting it up in SSR - that configuration is nowhere to be found.

I also tried to set a default value of `false` and override it via the script, to no avail. Lastly, I changed its name to `DSPACE_CYPRESSTEST` (maybe the number in `E2ETEST` could spell trouble), and again, nothing happens.

I might be missing something, though...

tdonohue: @AndreaBarbasso : I was looking around for how we've done this before, and I think I've stumbled on something that will work.  We already override the environment variables used in e2e tests in our `build.yml` here: https://github.com/DSpace/dspace-angular/blob/main/.github/workflows/build.yml#L17-L29

So, I'd propose we create a new boolean configuration called `test` (or `e2e`), at the same level as the boolean `production` configuration here: https://github.com/DSpace/dspace-angular/blob/main/src/config/app-config.interface.ts#L43

We *might* need to default it to `false` in all our `src/environments/environment.*.ts` files (not sure if this is required).

Then, in `browser-orejime.service.ts`, you should be able to add a simple check for `if (environment.test`) {`

Finally, to set this to true for our e2e tests, we just need to modify that `build.yml` list of environment variables to include:
```
env:
   # Enable "test" mode, which sets "environment.test" to true and disables some features (e.g. orejime).
   DSPACE_TEST: true
```

I believe this should work, because (as you can see in that build.yml) we already override some DSpace UI environment variables in order to ensure tests have other features disabled by default.

Finally, I should add, we could even make this setting *specific* to Orejime, and just call it `enableCookieConsent` under [the `info` section](https://github.com/DSpace/dspace-angular/blob/main/config/config.example.yml#L466-L469) like this:
```
info:
  enableCookieConsent: true
```
In that scenario, we'd default it to *true* though and set `DSPACE_INFO_ENABLECOOKIECONSENT: false` in the build.yml.

artlowel (changes requested):
Thanks @AndreaBarbasso!

While the correlation headers aren't sent when you disagree to the cookie, currently the cookie itself isn't removed. It looks like the reason is that the [cookie name configured for orejime](https://github.com/4Science/dspace-angular/blob/9d3b298f1a475b4a29be05008de8fa16afc68d8e/src/app/correlation-id/correlation-id.service.ts#L15) is not [the one we're actually using](https://github.com/4Science/dspace-angular/blob/9d3b298f1a475b4a29be05008de8fa16afc68d8e/src/app/correlation-id/correlation-id.service.ts#L37).  I expect that if we configure the correct cookie name, orejime will take care of removing the cookie. 

However, before [setting it again](https://github.com/4Science/dspace-angular/blob/9d3b298f1a475b4a29be05008de8fa16afc68d8e/src/app/correlation-id/correlation-id.service.ts#L51), it would be good to check if the user has agreed to it, otherwise I expect you'll see the cookie appear, and quickly being removed by orejime again every time you go to a new page. 

In general it would also be a good idea to use the constant for the cookie name everywhere it's needed.

As for the e2e issue. I prefer @tdonohue's last suggestion to add a `info.enableCookieConsentPopup` config property that defaults to true. That’s not only useful for our tests, but people who aren’t beholden to the GDPR might want to use it too

src/environments/environment.production.ts
```diff
@@ -14,4 +14,7 @@ export const environment: Partial<BuildConfig> = {
     enableSearchComponent: false,
     enableBrowseComponent: false,
   },
+  info: {
+    enableCookieConsentPopup: true,
+  },
```
tdonohue, line 19: @AndreaBarbasso : I see the e2e tests are still failing from the Orejime popup getting in the way.  I think it might be that these `environment.*.ts` settings should be **removed**.  

I think the *default* value of `enableCookieConsentPopup` should be added to the `default-app-config.ts` alongside all the other default values.  Otherwise, I *think* the `environment.*.ts` settings might *override* the settings elsewhere based on how our configs are all merged together here: https://github.com/DSpace/dspace-angular/blob/main/src/config/config.util.ts#L18-L21

src/app/shared/cookies/browser-orejime.service.ts
```diff
@@ -191,11 +191,13 @@ export class BrowserOrejimeService extends OrejimeService {
...
+        } else {
+          this.orejimeConfig.apps = this.filterConfigApps(appsToHide);
+        }
         this.applyUpdateSettingsCallbackToApps(user);
-
-        void this.lazyOrejime.then(({ init }) => {
+        this.lazyOrejime.then(({ init }) => {
           this.orejimeInstance = init(this.orejimeConfig);
         });
       });
```
artlowel, line 203: If we set the apps to `[]` you also can't open the cookie settings manually anymore when `enableCookieConsentPopup` is false, so we'd have to remove that broken link from the footer as well.

Maybe instead, you can leave the apps array intact, but immediately agree to them all if `enableCookieConsentPopup` is false:
```typescript
        this.orejimeConfig.apps = this.filterConfigApps(appsToHide);
        this.applyUpdateSettingsCallbackToApps(user);
        this.lazyOrejime.then(({ init }) => {
          this.orejimeInstance = init(this.orejimeConfig);
          if (environment.info?.enableCookieConsentPopup === false) {
            this.orejimeInstance.internals.react.saveAndHideAll();
          }
        });
  ```      
  
That would still allow you to open the cookie options from the footer link, though I'll admit it doesn't feel the most robust to have to use a function on inside `internals`. But I couldn't find another way to do so.
  
Either way I don't think this should block this PR

Comment (tdonohue) [reply src/app/shared/cookies/browser-orejime.service.ts:203]: If possible/easy to do so, I think it would make sense to just remove the "Cookie settings" link from the footer if `enableCookieConsentPopup` is `false`.  That'd provide similar behavior to other settings like `enablePrivacyStatement` and `enableCOARNotifySupport` which also remove links from the footer when disabled,

Comment (AndreaBarbasso) [reply src/app/shared/cookies/browser-orejime.service.ts:203]: @tdonohue, @artlowel, I removed the footer link if the cookie consent popup is disabled in the settings. I also moved the default value in the right place. Let's see if the pipelines work now!

**Comment by artlowel** [comment]
Thanks @AndreaBarbasso. That works, and the e2e tests pass! 🎉 

I found one more issue though. If you enable the popup, and open it for the first time with cleared cookies (or in an incognito window) you get an error in the console. I added an inline comment where it happens:

File: src/app/correlation-id/correlation-id.service.ts [hunk 6]
```diff
@@ -24,15 +36,32 @@ export class CorrelationIdService {
...
+    }
   }
 
   /**
-   * Initialize the correlation id based on the cookie or the ngrx store
+   * Check if the correlation id is allowed to be set, then set it
    */
   initCorrelationId(): void {
+    this.orejimeService?.getSavedPreferences().subscribe(preferences => {
+      if (preferences[CORRELATION_ID_OREJIME_KEY]) {
```
artlowel, line 52: If preferences is undefined, you get an error in the console:

```suggestion
      if (hasValue(preferences) && preferences[CORRELATION_ID_OREJIME_KEY]) {
```

artlowel (approved):
Thanks @AndreaBarbasso!

tdonohue (approved):
👍 Thanks @AndreaBarbasso !  This looks great to me now too, and all my testing is successful.  I've again verified that the `X-CORRELATION-ID` is not stored when the user turns it off.  I've also verified it works when turned on.  Finally, I verified that the new `enableCookieConsentPopup` can be turned off to disable the popup altogether (and remove the footer link)

One note: After today's Developer Meeting, I did realize we *might* still want to backport the ability to manage the `X-CORRELATION-ID` to 8.x and 7.x.  That would involve a partial backport of this PR though, because it doesn't require the ability to turn off orejime (which didn't exist in 8.x and 7.x anyways).  But, there's not a rush to backport this immediately, and I'd consider it optional at this time.

PR #3913, opened 2025-01-28
[Port dspace-7_x] Browse performance fixes

alexandrevryghem:
Port of #3846 to `dspace-7_x`.


tdonohue (approved):
👍 Thanks @alexandrevryghem !  Reviewed the code, and I also gave this a manual test (since there are some necessary differences in the code for 7.x).  It all works well.

PR #3914, opened 2025-01-28
[Port dspace-8_x] Browse performance fixes

alexandrevryghem:
Port of #3846 to `dspace-8_x`.


tdonohue (approved):
👍 Thanks @alexandrevryghem !  I verified the code changes here are identical to #3846.  Looks good to me.

PR #3924, opened 2025-01-29
[Port dspace-7_x] Made expandable navbar section more keyboard accessible

alexandrevryghem:
Port of #3581 to `dspace-7_x`.


tdonohue (approved):
👍 Thanks @alexandrevryghem !  Tested this manually and verified it's working properly in 7.x

PR #3925, opened 2025-01-30
[Port dspace-7_x] Prevent request with page size of 9999

FrancescoMolinaro:
## References

Backporting of https://github.com/DSpace/dspace-angular/pull/3694 to `dspace-7_x`


tdonohue (approved):
👍 Thanks @FrancescoMolinaro !  I tested this today and it seems to be an accurate backport.  I'm seeing all of the new behaviors on the modified components & no obvious errors.

PR #3957, opened 2025-02-05
fixed deprecated import of TransferState

saschaszott:
## Description

This PR fixes the ESLint deprecation warning

`'TransferState' is deprecated. TransferState has moved, please import TransferState from @angular/core`

This affects all DSpace version which use Angular 17 or above.


alanorth (approved):

alanorth:
Thanks @saschaszott! This is obviously correct. After applying this patch I no longer see the warning during lint.

PR #3973, opened 2025-02-11
submission-forms mergeMap issues fix

### LINKED ISSUE #3972: Submission errors not showing
## Describe the bug
In very specific custom submission setups, it is possible errors shown on top of the submission sections are not displayed at all, when they should be present. In those cases, the section is also marked as valid when it shouldn't.

## To Reproduce
First, set up the project with the following conditions in the backend:
- In `item-submission.xml`, configure a submission definition containing two steps with type "submission-form" (e.g. "traditionalpageone" and "traditionalpagetwo")
- Modify (or configure a different) `MetadataValidation` to add section-wide errors (`/sections/{sectionId}`). By default, this validation will only allow metadata field errors, but this bug requires section-wide errors.

Next, start a submission in the configured form and:
- In one of the two "submission-form" sections, fill in any value of any of the fields
- In the other section, perform an action that would result in the custom section-wide error

If the setup is correct, this should result in NO section-wide error (at the top of the section) being displayed for the relevant section.
If the other section is left empty, the error _does_ display.

## Expected behavior
The error should always display when the conditions are met, regardless of the other sections within the submission.

Atmire-Kristof:
## References
* Fixes #3972 

## Description
This PR fixes a bug caused by a `mergeMap` in submission by replacing it with a `switchMap` which is a more suitable map for the specific situation the map is used in. (more elaborate technical details below)

## Instructions for Reviewers
How to reproduce: Follow the steps from issue #3972, as this requires a very specific custom setup for the error to occur (or at least one setup I could find, there are likely others)

## Technical details
What this `mergeMap` is trying to achieve:
For "submission-form" sections, retrieve the "section-wide" errors (to display on top of the section) from the section's state and combine them with metadata field errors retrieved from the form's state (requires the form ID from the section state, thus inside a map operator to combine observables).

Why `mergeMap` causes issues:
This type of map is known for the inconsistent order of execution from the internal observable it's trying to merge with. It is mainly used when we shouldn't care about the order, but when we want all of the observable emissions to complete within.
A use-case would be when several requests are sent in the inner observable, say two patch requests, it would be of importance that all of the observables complete.
In our case, this sometimes causes the emission for an empty array of errors to fire AFTER the emission containing the error we require to be displayed. Because the empty array fires last, it will make whatever code using this map to think there are no errors, when there in fact are some, but they got drowned by the later executing observable.

Why `switchMap` is a better alternative here:
This type of map will ONLY accept the latest emission of the inner observable, meaning it will cancel any pending observable emission (e.g. empty array of errors). This should cause the errors to always emit correctly, because the latest emission is the correct one, regardless of how long each emission takes to complete.

I know that was a lot to go through and try to explain, but hopefully that paints a picture why `switchMap` is a safer option in this (and frankly many other) case, avoiding potentially very hard to spot issues.

## Checklist

### DIFF
--- a/src/app/submission/sections/sections.service.ts
+++ b/src/app/submission/sections/sections.service.ts
@@ -12,12 +12,12 @@ import isEqual from 'lodash/isEqual';
 import {
   combineLatest,
   Observable,
+  switchMap,
 } from 'rxjs';
 import {
   distinctUntilChanged,
   filter,
   map,
-  mergeMap,
   take,
 } from 'rxjs/operators';
 
@@ -206,7 +206,7 @@ export class SectionsService {
       errorsState$ = this.getSectionErrors(submissionId, sectionId);
     } else {
       errorsState$ = this.getSectionState(submissionId, sectionId, sectionType).pipe(
-        mergeMap((state: SubmissionSectionObject) => this.formService.getFormErrors(state.formId).pipe(
+        switchMap((state: SubmissionSectionObject) => this.formService.getFormErrors(state.formId).pipe(
           map((formErrors: FormError[]) => {
             const pathCombiner = new JsonPatchOperationPathCombiner('sections', sectionId);
             const sectionErrors = formErrors


nwoodward:
Hi @Atmire-Kristof. This PR looks good, and the explanation for switching from `mergeMap` to `switchMap` makes sense. But I'm having difficulty modifying `MetadataValidation` to add section-wide errors. If you could provide some sample code that would help me to test it.

pzlakowski:
Checklist:

- [x] I have checked PR against current main
  - [x] with mergeMap
  - [x] with switchMap  
  - [x] simulated response times 
- [x] I have attempted to reproduce the issue given instructions

**Reproduction:**

I attempted to try to reproduce the problem but given multiple scenarios including given in the issue, nothing worked. 

The only way to trigger section wide errors for me was to run "Deposit" action and then error would be displayed always. When we fill value in section A and then move to section B it triggers request to patch data but the error for section B would not get displayed. It is stored in the store as `serverValidationError` but it would never propagate to set generic error. 

@nwoodward 

This is a way to add section-wide errors. Add this validate method just before returning errors, then type test and click deposit. Section wide errors are added once action e.g. like Deposit is called since it reinitializes the form. 

`space-server-webapp/src/main/java/org/dspace/app/rest/submit/step/validation/MetadataValidation.java`
```java
if (config.getId().equals("traditionalpagetwo")) {
    List<MetadataValue> abstractMetadata = itemService.getMetadataByMetadataString(obj.getItem(), "dc.description.abstract");
    String abstractValue = "";

    if (abstractMetadata.size() > 0) {
        abstractValue = abstractMetadata.get(0).getValue();
    }

    if (abstractValue.equals("test")) {
        addError(errors, ERROR_VALIDATION_REQUIRED, "/sections/" + config.getId());
    }
}
```

My `item-submission.xml` looks like this:

```xml
<item-submission>
  <submission-map>
    <name-map collection-handle="default" submission-name="prtest"/>

    <!-- ... -->
  </submission-map>

  <submission-definitions>
    <submission-process name="prtest">
      <step id="traditionalpageone" />
      <step id="traditionalpagetwo" />

      <!-- ... -->
    </submission-process>
  </submission-definitions>
</item-submission>
```

Also to make the configuration of `traditionalpageone` & `traditionalpagetwo` you must fix `getCollectionVisibility` function since it attempted to take `visibility` property from undefined. Perhaps it should be noted as the other bug. 

`dspace-angular/src/app/submission/sections/form/section-form.component.ts`
```typescript
  private getCollectionVisibility(): SectionVisibility {
    const submissionSectionModel: SubmissionSectionModel =
      this.submissionDefinition.sections.page.find(
        (section) => isEqual(section.sectionType, SectionsType.Collection),
      );

    return isNotUndefined(submissionSectionModel?.visibility) ? submissionSectionModel.visibility : null;
  }
```

PR #3981, opened 2025-02-13
No platform specific code in abstract services main

AAwouters:
## References
- Fixes #3974 
- 8_x backport: #3982 
- 7_x backport: #3983  

## Description
Moves platform-specific imports out of the abstract [Cookie-](https://github.com/DSpace/dspace-angular/blob/446f6810984a11fb865399aa4a2d659ea59985e8/src/app/core/services/cookie.service.ts) and [AuthService](https://github.com/DSpace/dspace-angular/blob/446f6810984a11fb865399aa4a2d659ea59985e8/src/app/core/auth/auth.service.ts) and moves them to [ServerCookie-](https://github.com/DSpace/dspace-angular/blob/446f6810984a11fb865399aa4a2d659ea59985e8/src/app/core/services/server-cookie.service.ts) and [ServerAuthService](https://github.com/DSpace/dspace-angular/blob/446f6810984a11fb865399aa4a2d659ea59985e8/src/app/core/auth/server-auth.service.ts).

## Instructions for Reviewers
This PR does not directly produce an observable difference.
It does solve the problem mentioned in #3974.

## Checklist
- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [x] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [x] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [x] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC8x/Accessibility)** if it makes changes to the user interface.
- [x] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
- [x] My PR **includes details on how to test it**. I've provided clear instructions to reviewers on how to successfully test this fix or feature.
[...]

PR #3982, opened 2025-02-13
No platform specific code in abstract services 8_x

AAwouters:
DSpace-8_x Backport of #3981.


tdonohue (approved):
👍 Thanks @AAwouters !  Tested this via #3981.  This code is identical.

PR #3983, opened 2025-02-13
No platform specific code in abstract services 7_x

AAwouters:
DSpace-7_x backport of #3981.


alanorth:
Thanks @AAwouters. I tested this on DSpace 7.6.4-SNAPSHOT by logging in, making an item submission, searching, and browsing a bit. I didn't see any change in functionality.

So +1 from me on functionality unless there is something else I should test.

tdonohue (approved):
👍 Thanks @AAwouters . I reviewed/tested this via #3981 (which I just merged).  This code looks the same and has been tested by @alanorth

Issue #3961, opened 2025-02-06
Reduce request chaining and associated overhead


ybnd:
As described in https://github.com/DSpace/dspace-angular/issues/3110#issuecomment-2377716959 and https://github.com/DSpace/dspace-angular/issues/3110#issuecomment-2379902378, many DSpace REST requests show multiple other requests as "initiators".

We're currently investigating the following avenues:
- Reducing the amount of chained requests
    - The example in the comments above mentions search filters, which depend on a chain of at least the following requests:
        - `/server/api` → HAL endpoint map
        - `/server/api/discover` → HAL endpoint map for discovery
        - `/server/api/discover/search` → HAL endpoint map / default search configuration
        - `/server/api/discover/objects` and/or `/server/api/discover/facets` → determine which facets are visible
        - `/server/api/discover/facets/<:id>` → finally retrieve the actual filter
    - These can be direct (within the same RxJS pipe) or indirect (components rendering after a response is received and sending their own requests)
- Assessing our total NgRx/RxJS overhead
    - We may want to try and tackle request chains by embedding more or pre-caching, but we'd still go through a long stack of NgRx selectors and RxJS operators
    - Intuitively we think that this shouldn't have _too_ much impact, but it's not immediately clear how to measure this

PR #3993, opened 2025-02-14
[Port dspace-7_x] Cache-bust dynamic configuration files and theme CSS

### LINKED ISSUE #1961: Add hash/fingerprint to the custom-theme CSS files
Add hash/fingerprint to the custom-theme CSS files as we do with all other JS and CSS files. This way one could set a Cache-Control HTTP header for all CSS files with a long expiration date and still have the new CSS file loaded after it changed in an update, because the hash/fingerprint will have changed.

That's how all other JS and CSS files are treated:
https://demo7.dspace.org/styles.e382c7dd35ca1222.css
https://demo7.dspace.org/main.971071c37a2addc9.js
https://demo7.dspace.org/polyfills.279549ba67e21ba8.js

That's how the custom-theme CSS files are treated. The hash/fingerprint part in the filename is missing:
https://demo7.dspace.org/dspace-theme.css

### LINKED ISSUE #3992: Angular runs with stale configuration when SSR is bypassed
## Describe the bug
The production server for DSpace Angular generates a `config.json` file every time it starts up, based on:
- YAML configuration
- `DSPACE_...` Environment variables

This JSON file is the only source of configuration when SSR is turned off, but it is also cached by the browser. Configuration changes cannot be reliably pushed in this case: users can either
- Wait for their cache to invalidate by itself (default: 7 days) → not immediate
- Invalidate their cache manually (delete site data, developer tools, ...) → not something the average user will think of doing

## To Reproduce
1. Start a local DSpace Angular server in production mode
2. Open a page in your browser, with the network tab open in the developer tools
    - You will see a request for `config.json`
3. Stop the production server, make a change to the configuration via YAML or an environment variable and restart the server
4. Refresh the page in your browser ("regular" refresh)
   - You'll see that `config.json` is server from the cache, and doesn't include the change
5. Invalidate the cache manually and refresh again ("hard refresh)
   - Now `config.json` will be fetched again, including the change

## Proposed solution
We want to introduce a cache busting mechanism for files that our Angular server generates dynamically, outside of the build.
This would allow us to invalidate coniguration when it changes, in the same way as we already do for compiled JS and CSS.

ybnd:
## References
* Fixes #3992
* Fixes #1961

## Description
DSpace Angular's production server can modify the configuration it serves to CSR-only clients through YAML or environment variables. However, these files can remain cached in the browser and leave users with out-of-date configuration until the TTL runs out or the user does a "hard refresh". On build time this sort of problem is solved by saving the content hash as part of the path of each file (JS, CSS, ...)

We introduce HashedFileMapping to bridge the same gap for dynamic content generated _after_ the server is built:
- Files added to this mapping on the server are copied to a hashed path
- A copy is injected into index.html, where clients can read it out and resolve the hashed paths
- If a given path is not found in the mapping, the client will fall back to the original version (to prevent errors in development mode)

With this mechanism we can ensure updates to config.json and similar files take immediate effect without losing the performance benefit of client-side caching.

Compression support is included in anticipation of a pending PR for https://github.com/DSpace/dspace-angular/issues/3961, where the initial size of `config.json` can become large enough where compressing it would be a plus.

## Discussion
- It's not clear how/if we can unit test this at the time of writing
- Should we inlude a configuration flag to toggle this feature on/off? We may want to roll it out as opt-in so downstream users can wait it out until it has been battle-tested a bit.

## Instructions for Reviewers

To test:
1. Run DSpace Angular locally, in production mode
2. Confirm that you see the following files:
    - `[dist]/assets/config.<hash>.json` → hashed configuration file
    - `[dist]/dspace-theme.<hash>.css` → hashed theme CSS file
    - `[dist]/index.html` → index file with an extra `script#hashed-file-mapping` element containing the mapping of plain paths to hashed paths
3. Confirm that `config.<hash>.json` and `dspace-theme.<hash>.css` are correctly served to the browser, and that the app runs/looks as before the PR
5. Confirm that https://github.com/DSpace/dspace-angular/issues/3992 is resolved

## Checklist


ybnd:
@tdonohue adjusted the labels since this PR was opened against 7.x

ybnd:
Note: when running multiple DSpace Angular instances in PM2 cluster mode this would cause the same filesystem operations in parallel, could cause problems.

Could be addressed with Redis similar to https://github.com/DSpace/dspace-angular/issues/3110#issuecomment-2188477157

pcg-kk (commented):
Work for me fine, tested with SSR & without SSR

kshepherd:
@ybnd could you rebase + repush to force the build tasks to re-run ? the unit test was so long ago that the logs aren't available so it's not clear why it failed

ybnd:
@kshepherd if possible I'd like avoid rebasing to keep the history consistent with the "clean 7.6" version of the branch (already using that elsewhere)

Looking back now I did find a few small improvements to make (mainly with the theme prefetching part), so that will revive the CI just as well.

MW3000 (approved):
> Run DSpace Angular locally, in production mode
> Confirm that you see the following files:
> `[dist]/assets/config.<hash>.json` → hashed configuration file

Yes, `config.<hash>.json` files are correctly generated ✅
The `config.<hash>.json` is correctly referenced ✅

> `[dist]/dspace-theme.<hash>.css` → hashed theme CSS file

Yes, custom theme `theme.<hash>.css` files are correctly generated and referenced ✅

> `[dist]/index.<hash>.html` → index file with an extra script#hashed-file-mapping element containing the mapping of plain paths to hashed paths

There is no `index.<hash>.html` file generated, just the normal `index.html` file! But I guess, that is just an error in this description? As I understand the code, there should not be an `index.<hash>.html` file, as the normal `index.html` is being excluded from cashes? ❓

The `index.html` contains the correct hashed-file-mappings ✅

> Confirm that `config.<hash>.json` and `dspace-theme.<hash>.css` are correctly served to the browser, and that the app runs/looks as before the PR

Confirmed ✅

The feature works as described. I read the code and see what it does but I am not competent enough to decide if this is the right solution architecturally.

ybnd:
@MW3000 

> There is no `index.<hash>.html` file generated, just the normal `index.html` file! But I guess, that is just an error in this description? As I understand the code, there should not be an `index.<hash>.html` file, as the normal `index.html` is being excluded from cashes?

Good eye, that bit stayed behind in the PR description from an earlier iteration.

Initially I tried to use the same hashing mechanism for the index file, but that lead to tricky edge cases depending on _how exactly_ it was requested → better to rely on `Cache-Control` headers

Amended the description to reflect this ✔️

tdonohue:
@ybnd : I took a quick look at the code in this PR today.  Overall, it looks useful.  But, I'm wondering why we are starting with the `dspace-7_x` branch instead of testing this on `main`?  Do we have a stable version of this PR against `main` that we can also review?

I'm not against also potentially backporting this all the way to 7.x.  I just often find that *forward ports are more difficult* (because of new lint rules and similar), so I think it's important to get a `main` version of this PR ready in order to move this change forward.

ybnd:
@tdonohue just out of preference to keep the same commit history across the supported versions; at the time this was opened the 9.x draft PR was pointed at `main`, probably should have opened that one instead of 7.x

@AbhinavS96 prepared a PR here: https://github.com/DSpace/dspace-angular/pull/4936

tdonohue (approved):
👍 Thanks @ybnd !  Verified this is the same as #4936.  I also see it's already been tested by others. Merging immediately.

This still needs to be ported to 9.x and 8.x but I see PRs for those already exist in draft form (see #4030 and #4031)

PR #4020, opened 2025-02-21
Changes to fr.json5 following version 8.1

pilasou:
Adding and changing parameters after upgrading to 8.1.


nimabehforouz (approved):
All French labels reviewed and verified.

tdonohue (approved):
👍 Thanks @pilasou !

PR #4035, opened 2025-02-26
[Port main] Add citation_doi tag to head meta

alanorth:
Port of #4034 to `main`.


tdonohue (approved):
👍 Looks good. Tested via https://github.com/DSpace/dspace-angular/pull/4058#issuecomment-2704034985 and approved via #4034

PR #4036, opened 2025-02-26
[Port dspace-8_x] Add citation_doi tag to head meta

alanorth:
Port of #4034 to `dspace-8_x`.


tdonohue (approved):
👍 Approved via #4034.  I've verified this code is identical

PR #4038, opened 2025-02-26
Translations for access status filter

paulo-graca:
## References
* Requires DSpace/DSpace#10434

## Description
Translations required by https://github.com/DSpace/DSpace/pull/10434


tdonohue (approved):
👍 Thanks @paulo-graca !  Tested this alongside the backend PR and it looks good to me

PR #4057, opened 2025-03-06
apply spellCheck configuration to input model

saschaszott:
## Description

With this PR, the `spellCheck` attribute is also applied to simple input fields based on the application configuration `form.spellCheck`. Currently, the `spellCheck` attribute is only set for textarea fields (see #2003).


tdonohue (approved):
👍 Thanks @saschaszott !  Stumbled on this today, and it's a tiny useful improvement.  I tested it and verified that input boxes (onebox) now have spell check capabilities in the submission form.

PR #4168, opened 2025-04-11
Fix export button enabled in bulk access management without selecting step 2

AbhinavS96:
## References
- Fixes #3893
- Backport PR #4169

## Description
The export button in Bulk Access Management was not considering the state of step 2 in the form. This has been fixed.

## Instructions for Reviewers
List of changes in this PR:
* Export button in Bulk Access Management has an extra check in `canExport` now.
* Tests for the component are modified. This includes tests for the new case and some refactoring to handle the dynamically loaded components.

This can be tested by logging in as Admin and navigating to Bulk Access Management. The button should activate only when there is some selection in both steps.

## Checklist

- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [x] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [x] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [x] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC8x/Accessibility)** if it makes changes to the user interface.
- [x] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
- [x] My PR **includes details on how to test it**. I've provided clear instructions to reviewers on how to successfully test this fix or feature.
- [x] If my PR includes new libraries/dependencies (in `package.json`), I've made sure their licenses align with the [DSpace BSD License](https://github.com/DSpace/DSpace/blob/main/LICENSE) based on the [Licensing of Contributions](https://wiki.lyrasis.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-LicensingofContributions) documentation.
- [x] If my PR includes new features or configurations, I've provided basic technical documentation in the PR itself.
- [x] If my PR fixes an issue ticket, I've [linked them together](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

PR #4169, opened 2025-04-11
[Port dspace-7_x] Fix export button enabled in bulk access management without selecting step 2

AbhinavS96:
Port of #4168 to `dspace-7_x`.


tdonohue (approved):
👍 Thanks again @AbhinavS96 !  Verified this is the same code as #4168, except adapted to 7.x

PR #4340, opened 2025-05-15
Fix unqualified metadata field validation in Edit Metadata tab by sorting fields in validation request

alexandrevryghem:
## References
* Fixes #4339

## Description
Fixed a validation issue that prevents you from submitting unqualified metadata fields, if the unqualified metadata field was added to the registry when 20 qualified fields with the same schema & element were already present in the DB

image

## Instructions for Reviewers
List of changes in this PR:
* Updated the `MetadataFieldSelectorComponent` to sort the field names during the `validate()` call. This ensures that the unqualified field is always present on the first page. I also reused the other `pageOptions` as well, this way if you type out the whole metadata field name correctly, the validate call won't send out an additional network request

**Guidance for how to test or review this PR:** See #4339

## Checklist
- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [x] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [x] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [x] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC8x/Accessibility)** if it makes changes to the user interface.
- [x] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
- [x] My PR **includes details on how to test it**. I've provided clear instructions to reviewers on how to successfully test this fix or feature.
- [x] If my PR includes new libraries/dependencies (in `package.json`), I've made sure their licenses align with the [DSpace BSD License](https://github.com/DSpace/DSpace/blob/main/LICENSE) based on the [Licensing of Contributions](https://wiki.lyrasis.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-LicensingofContributions) documentation.
- [x] If my PR includes new features or configurations, I've provided basic technical documentation in the PR itself.
[...]

PR #2856, opened 2024-03-06
fix 'undefined' error on anonymous item page landing

frabacche:
## References
_Related issue:_
* Fixes #2842

## Description
when configuration ldn.enabled=false, anonymous item page landing shows an 'undefined' error on browser console.

## Instructions for Reviewers
Error is not present landing on item page as anonymous or logged user, with the ldn.enabled configuration evaluated as true or false.

PR #4370, opened 2025-05-23
[Port dspace-8_x] Fix unqualified metadata field validation in Edit Metadata tab by sorting fields in validation request

AbhinavS96:
Port of #4340 by @alexandrevryghem

PR #4383, opened 2025-05-26
[Port dspace-7_x] Fix unqualified metadata field validation in Edit Metadata tab by sorting fields in validation request

AbhinavS96:
Port of #4340 by @alexandrevryghem

Also backported https://github.com/DSpace/dspace-angular/commit/e6389a73df10dea2ca9c492d8ca4b322000f3731 from #2856 to fix the `item-edit.cy.ts` test


tdonohue (approved):
👍 Thanks @AbhinavS96 !  Tested this via #4370.  This is identical (except for the extra commit, which is identical to code already existing on dspace-8_x branch)

PR #4448, opened 2025-06-11
Lazy-load geospatial map code

ybnd:
## References
Port of #4447 for current main

### DIFF
--- a/src/app/browse-by/browse-by-geospatial-data/browse-by-geospatial-data.component.html
+++ b/src/app/browse-by/browse-by-geospatial-data/browse-by-geospatial-data.component.html
@@ -1,13 +1,11 @@
 <div class="container">
   <h1>{{ 'browse.metadata.map' | translate }}</h1>
-  @if (isPlatformBrowser(platformId)) {
-    @defer {
-      <ds-geospatial-map [facetValues]="facetValues$"
-                         [currentScope]="this.scope$|async"
-                         [layout]="'browse'"
-                         style="width: 100%;">
-      </ds-geospatial-map>
-    }
+  @defer {
+    <ds-geospatial-map [facetValues]="facetValues$"
+                       [currentScope]="this.scope$|async"
+                       [layout]="'browse'"
+                       style="width: 100%;">
+    </ds-geospatial-map>
   }
 </div>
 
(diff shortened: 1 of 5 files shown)


kshepherd:
@ybnd thanks for looking at this - let me know when it is ready for review (tests etc) and i'll help get it moving

ybnd:
@kshepherd should be ready now 🤞

Let me know if you think we should spend more time to try and get the Item page field unit test working more as it did before.

kshepherd (approved):
+1 thanks @ybnd , this is an important change and something I should have been more careful about when first implementing. The test change is fine by me.

src/app/browse-by/browse-by-geospatial-data/browse-by-geospatial-data.component.html
```diff
@@ -1,11 +1,13 @@
 <div class="container">
   <h1>{{ 'browse.metadata.map' | translate }}</h1>
   @if (isPlatformBrowser(platformId)) {
```
atarix83: @ybnd 
as mentioned in the documentation https://angular.dev/guide/templates/defer#how-does-defer-work-with-server-side-rendering-ssr-and-static-site-generation-ssg

using the defer the block is non renderend during SSR, so we can simplify the code removing the if statment

src/app/shared/object-geospatial-map/object-geospatial-map.component.html
```diff
@@ -1,3 +1,3 @@
 @if (isPlatformBrowser(platformId)) {
```
atarix83: also here we can use defer block instead if statement

src/app/browse-by/browse-by-geospatial-data/browse-by-geospatial-data.component.html
ybnd: Good point, thanks!

src/app/shared/object-geospatial-map/object-geospatial-map.component.html
ybnd: Putting a `@defer` here actually splits Leaflet & Terraformer into two separate JS files that are loaded sequentially, which is not ideal (see https://angular.dev/guide/templates/defer#avoid-cascading-loads-with-nested-defer-blocks).

IMO it's fine to ensure that `ds-object-geospatial-map` is always used within a `@defer` itself, so I removed the check altogether.
Also added some hints so we remember this later if/when these components are used elsewhere.

atarix83 (approved):
thanks @ybnd LGTM

PR #4519, opened 2025-07-01
[Port dspace-7_x] Accessibility settings page SSR Fix

AAwouters:
SSR fix for issue mentioned in #3615.


tdonohue (approved):
👍 Thanks @AAwouters !  I've verified that this is identical to the fix applied to #3614, which I've verified works.

PR #4522, opened 2025-07-02
[Port dspace-9_x] Vocabulary preloadlevel fix

AAwouters:
Dspace-9_x backport of #3819


tdonohue (approved):
👍 Thanks @AAwouters !  Verified this is the same as #3819 (which I tested yesterday)

PR #4446, opened 2025-06-11
Removed theme modules

alexandrevryghem:
## Description
Currently all the themed components in `src/themes` have been migrated to standalone declaration, but they don't work unlesss they are referenced inside either the `LazyThemeModule` or `EagerThemeModule`. I updated the config in order to remove this constraint, this way we won't have to fix merge conflicts in those files anymore.

The reason why those modules were still necesarry is because the `tconfig.*.json` files only compile the `*.module.ts` files inside the `src/themes` folder. Since those modules contained imports to all the components inside the theme, they were automatically comiled as well. I updated the config to compile all the typescript files inside the `src/themes` folder, this way we can drop the modules.

## Instructions for Reviewers
List of changes in this PR:
- Updated the `tsconfig.*.json` files to compile all the typescript files inside the `src/themes` folder instead & removed the unused `angularCompilerOptions.entryModule` property that isn't being used anymore since the switch to standalone components
- Removed all the `LazyThemeModule`s
- Removed all the `EagerThemeModule`s and replaced them with 2 arrays:
  - `lazy-listable-components.ts#LISTABLE_COMPONENTS`: This is similar to the old `ENTRY_COMPONENTS` array, but it only contains imports for components that use `@listableObjectComponent` decorators. Other dynamically loaded components that don't use custom decorators like `StartsWithDateComponent` should be declared in the decorator files instead (e.g. `starts-with-decorator.ts`).
  - `eager-theme-components.ts#COMPONENTS`: This is similar to the `DECLARATIONS` array from the old `EagerThemeModule`, except that it doesn't contain the `ENTRY_COMPONENTS` entries

**Guidance for how to test or review this PR:**
- Verify that everything still works as expected (you can test it with the custom & eager theme)
- Verify the bundle size didn't increase (you can use `build:stats` for this)

## Checklist
- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [x] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
[...]

PR #4528, opened 2025-07-03
Restore dynamic component decorators

### LINKED ISSUE #2899: Size regression in `main.js` bundle caused by standalone component migration
**Describe the bug**
Looks like migrating to standalone components in #2750 increased the size of our `main.js` bundle by about 30kB.

While this is not a significant increase, it's a bit counter-intuitive since the claim is that standalone components should _improve_ the bundle structure.

Here are some improvements we can make:
- Move route-specific components out of `ListableModule` and reduce the number of imports
  - Mainly: Standalone Item pages, MyDSpace components
  - This can easily save us ~60kB (gz)
  - A little fragile though: we've already seen that overwriting these components in themes causes regressions. We can propose a workaround to mitigate this.
- Make dynamic form code lazy again
  - Dynamic forms are only used on a select number of routes, many of which are not even accessible by anonymous users, so it would be good to keep them out of the main bundle
  - Looked into this for a short while, and it's not as easy but could save us another ~40kB (gz)

alexandrevryghem:
## References
* Fixes #2899

## Description
When DSpace was migrated to standalone components the decorators on all the [dynamic components were removed](https://wiki.lyrasis.org/display/DSPACE/Migration+to+Standalone+Components#MigrationtoStandaloneComponents-HandlingCustomDecoratorsforMapsofInjectableComponents). The way dynamic components used to work was by mapping the different types of supported components with their component declaration. These mappings were automatically populated by the decorators themselves once the class was loaded, so that's why we always loaded those components eagerly. During the switch to standalone components the logic to populate the maps automatically was replaced with hardcoded maps. This PR now generates these hardcoded decorator maps automatically during every rebuild using the new `npm run generate:decorator:registries` script, and we also lazy loaded all the component imports to reduce the main bundle size.

## Instructions for Reviewers
List of changes in this PR:
* A new script `scripts/generate-decorator-registries.ts` has been created that goes over the whole code and finds all the decorators defined inside the `DECORATORS` array. These decorators are then all bundled together in a registry file that will be used to retrieve the matching component.
* A new custom webpack plugin has been created to run `npm run generate:decorator:registries` on every rebuild
* All the existing decorators that were not themable yet, have been made themable
* Update the `listable-object.decorator.ts#getMatch` not always being able to fall back to a default component
* Used the `listable-object.decorator.ts#getMatch` logic in all the decorators to find the best match

- Go over all the decorators listed in `generate-decorator-registries.ts#DECORATORS` and verify that they are still all rendering correctly in prod & dev mode
- Verify the bundle size is now smaller (757.6KB → 631.27KB)

## Checklist


FrancescoMolinaro (commented):
Hello @alexandrevryghem , @artlowel ,

Many thanks for these changes. I’ve taken an initial look at the PR and performed some testing, and I wanted to share some feedback with you.

Overall, I believe this PR is an improvement over the previous static map approach, both in terms of bundle size and maintainability. It also brings back the decorator functionality, which is definitely a nice feature to have.

During my testing, however, I noticed that the script is triggered multiple times during the build process. This could be related to how files are currently being watched by the Webpack plugin, as the script appears to fire on any type of file change. This behavior could lead to unnecessary resource usage and longer build times, especially in development mode.

Additionally, the way the decorators are currently generated raises some concerns regarding the upcoming migration to Nx. As it stands, this kind of independent build process would be difficult to align with Nx’s principles—particularly the project isolation model—and might even break it.

I think it would be best to tailor this solution within the context of an Nx workspace, rethinking how decorators are built. For example, we could isolate each decorator in its own library and ensure the script rebuilds only the files within those libraries when changes occur.

In any case, I believe it would be valuable to discuss this in one of the upcoming community calls, so we can gather feedback from other developers as well.

alexandrevryghem:
@FrancescoMolinaro Thnx for the review & sorry for the delayed response.

I tested [the new NX PR](https://github.com/DSpace/dspace-angular/pull/4019), alongside this script and verified that both PRs are compatible. The only thing we would need to change is that the current implementation only searches for decorators in the [`src` folder](https://github.com/DSpace/dspace-angular/blob/540bf4dcc834c3b9362ce93abb55a3fdb1073a64/scripts/generate-decorator-registries.ts#L39), but now we would need to include the modules folder as well.

Regarding the isolation of the decorators in the modules, I would not recommend doing that. This would mean that if we wanted to create a new plugin for dspace that adds a `@dataservice` for example, that we would always have to create the service in the core module. Whereas the idea is that such a plugin should have its own core module that depends the dspace core module.

To prevent services from disabled modules to be available at runtime, we could update the script in the future though to only scan for decorators inside enabled modules.

Regarding performance, the script takes less than half a second to generate all the registries so I don't think this is a problem. If people would however want to disable this auto rebuild it would be easy though, you would only have to remove the plugin from webpack and run `npm run generate:decorator:registries` manually. But running the script manually takes more time and would complicate the build process for developers so I would leave it enabled by default.

tdonohue:
@alexandrevryghem : When you have a chance, I'd appreciate it if you can rebase this on latest `main`.  Now that #4446 is merged, it seems like this can provide more improvements to themes in 10.x.

I realize there was some discussion above around overlap with Nx (with @FrancescoMolinaro above). But, since the Nx work isn't likely to occur for 10.x, I'd like us all to consider whether this PR is still a _useful step in the right direction_ for 10.0.

FrancescoMolinaro:
Hi @alexandrevryghem , sorry for the delay on this and thanks for your works, I tried this again today and the functionality seems working as expected, the only thing is that gitHub is now signaling 260 files changed and seems that includes a lot of files/changes already on main.
I am not sure if there was an issue in merging the main branch or is just gitHub being "stuck", as the files I have checked look identical.
Could you please have a look so that is clear which files changed in this PR? Maybe just triggering a new comparison from gitHub will be enough to fix it.

alexandrevryghem:
@FrancescoMolinaro thnx for the review! I took a quick look at the current diff on GitHub and all the changes that I can see on my end do look like intentional modifications that are not present on `main` yet. Could you maybe give some examples of files that shouldn't be included in the diff but are present in yours?

NOTE: we started the development of this right after dspace 9.0 got released so you can look at [the original branch this PR started from here](https://github.com/DSpace/dspace-angular/compare/main...atmire:dspace-angular:0f4fce32076f51eea7987c6318de58e04012e0d1), and it has 255 file changes. So having 260 files changes after the merge with `main` does seem like a reasonable amount of changes

FrancescoMolinaro:
Hi @alexandrevryghem , sorry I think it was an issue only on my side, I was seeing code changes to the yml files for docker and the pipelines, cleaning the browser cache did solve the issue and now I can see the correct files, I thought the number of files changed was high because of that but looking at the correct code changes it is a reasonable number as this PR thouches all the components that previously had a decorator.

I am going to do a code review asap, thanks and sorry again for the false alarm.

src/app/shared/menu/menu-section/abstract-menu-section.component.ts
```diff
@@ -11,40 +16,61 @@ import {
...
+import { ThemeService } from '../../theme-support/theme.service';
 import { MenuService } from '../menu.service';
 import { MenuID } from '../menu-id.model';
 import { getComponentForMenuItemType } from '../menu-item.decorator';
+import { LinkMenuItemModel } from '../menu-item/models/link.model';
 import { MenuItemModel } from '../menu-item/models/menu-item.model';
 import { MenuItemType } from '../menu-item-type.model';
 import { MenuSection } from '../menu-section.model';
 
+export interface MenuSectionDTO {
```
FrancescoMolinaro: Hi @alexandrevryghem , this interface seems to be defined and exported with the same name but different structure in menu.component.ts, if both are needed could we change the name of one of the two so that the difference is clear?

src/app/shared/abstract-component-loader/abstract-component-loader.component.ts
```diff
@@ -79,7 +79,7 @@ export abstract class AbstractComponentLoaderComponent<T> implements OnInit, OnC
    * Set up the dynamic child component
    */
   ngOnInit(): void {
-    this.instantiateComponent();
+    void this.instantiateComponent();
```
FrancescoMolinaro, line 84: @alexandrevryghem I have seen that this method is now not a void method anymore, could we remove the void operator and add a catch for any possible error from the returned promise? I think otherwise errors in the component instantiation could be swallowed by the void operator resulting just in an empty component.

scripts/generate-decorator-registries.ts
```diff
@@ -0,0 +1,413 @@
+import {
```
FrancescoMolinaro, line 1: @alexandrevryghem since this file is the core logic of the decorator plugin and it's logic is quite complex, would be possible to add some unit tests so that it would be easier to find issues on future code changes? I think some tests to enforce the functions logic should be enough.

FrancescoMolinaro (commented):
Hi @alexandrevryghem , thanks again for your work on this, I did check again the code and generally it looks ok, I have found just a couple of things that I would like to address, you can find an inline comment for each of those.
Otherwise is looking good for me, thanks again for the effort.

src/app/shared/menu/menu-section/abstract-menu-section.component.ts
alexandrevryghem: 👍 renamed it to `MenuSectionComponentDTO`

src/app/shared/abstract-component-loader/abstract-component-loader.component.ts
alexandrevryghem: Since the `instantiateComponent` method is called on multiple places I added a try catch inside that method to deal with any potential issues

scripts/generate-decorator-registries.ts
alexandrevryghem: @FrancescoMolinaro: If we start creating tests for scripts we should ideally create a new testing suite similar to the e2e tests for them. The other scripts like the translation scripts currently also don't have them, so I think this is out of scope for this PR, but we can create a follow up issue for this

FrancescoMolinaro: Hi @alexandrevryghem I agree that is not in scope and is a wider topic of discussion, nevertheless I think we should eventually add tests also for the scripts as it would prevent some issue on updates and migrations as already happened for instance to the translation sync script. As you suggested I will create a separate issue for that.

FrancescoMolinaro:
Hi @alexandrevryghem, thanks for the update and your work on this, I can see all my previous feedbacks have been addressed, it looks good to me, could you please just have a look at the conflicts?. Thanks in advance!

FrancescoMolinaro (approved):

tdonohue:
@alexandrevryghem and @FrancescoMolinaro : While I think this PR is great, I'm hesitant to merge this at the "last minute" for 10.0 because of its size.  I don't feel I can thoroughly test this today prior to our code freeze for 10.0.  

However, I feel we should work to fully test & merge this quickly after 10.0 so that it can be added to a 10.1 and also get into 11.0.

alexandrevryghem:
@tdonohue Now that the DSpace 10 release is behind us, would it be possible to take another look at this PR?

tdonohue:
@alexandrevryghem : I'd gladly help to move this forward now.  However, I'm finding that this PR doesn't build for me now (it has a number of compilation errors if I pull it down and run `npm run build:prod`).  I suspect this just needs to be rebased on the latest `main` in order to incorporate changes that happened there at the end of the 10.0 release.

Once it's ready for a final test though, let me know and I'll add it to my list.

alexandrevryghem:
@tdonohue: Thxn! I merged the latest changes from `main` in this PR and the code compiles in prod mode, so it's ready for testing

tdonohue:
@alexandrevryghem : I finally got back to this today, and I'm still unable to build this PR on Windows.  It might be something Windows specific, but all I'm doing is this:
* `npm run clean`
* `npm install`
* `npm run build:prod` -> This is what fails.

The overall `build:prod` works up to the point that it tries to build the code in `./src/decorator-registries/`...and then all of that code fails to build with errors like this:
```
./src/decorator-registries/data-service-registry.ts - Error: Module build failed (from ./node_modules/@angular-devkit/build-angular/src/tools/babel/webpack-loader.js):
SyntaxError: C:\dspace-angular\src\decorator-registries\data-service-registry.ts: Bad character escape sequence. (34:61)

  32 | import { VERSION_HISTORY } from '..\app\core\shared\version-history.resource-type';
  33 | import { VERSION } from '..\app\core\shared\version.resource-type';
> 34 | import { USAGE_REPORT } from '..\app\core\statistics\models\usage-report.resource-type';
     |                                                              ^
  35 | import { CORRECTION_TYPE } from '..\app\core\submission\models\correctiontype.resource-type';
  36 | import { SUBMISSION_CC_LICENSE_URL } from '..\app\core\submission\models\submission-cc-licence-link.resource-type';
  37 | import { SUBMISSION_CC_LICENSE } from '..\app\core\submission\models\submission-cc-licence.resource-type';
```
And like this..
```
./src/decorator-registries/metadata-representation-component-registry.ts:15:99-227 - Error: Module not found: Error: Can't resolve '../appsharedobject-listmetadata-representation-list-elemenrowse-linrowse-link-metadata-list-element.component' in 'C:\dspace-angular\src\decorator-registries'
```


If I open the failing file (in this case `src\decorator-registries\data-service-registry.ts`) in Visual Studio Code, I see a TON of warnings for every `import` statement that uses the `..\app\core\` syntax instead of the `@dspace/core` syntax for paths.

For instance:
```
// This shows an error
import { EPERSON } from '..\app\core\eperson\models\eperson.resource-type';

// This doesn't show an error
import { WORKFLOWITEM } from '@dspace/core/eperson/models/workflowitem.resource-type';
```

I suspect the issue is that these generated files are not using paths like `@dspace/core`.  But, I'm not sure why the same errors are not occurring in this build process of this PR.  So, it's possible this is a Windows-specific problem, and the `../app/core` paths work fine on Linux systems.

alexandrevryghem:
@tdonohue: Thnx for testing it again, on my Mac I can run it without any issues, so I'll try running it on a Windows computer now

alexandrevryghem:
@tdonohue: I've fixed all the issues, there were indeed some issues with the paths. I've fixed & tested these bugs on my Windows machine & verified that they still work on Linux environments

tdonohue (commented):
@alexandrevryghem ;  Thanks for the updates.  I can now verify this compiles properly on Windows. Also, everything appears to work well (haven't found any obvious bugs in the UI changes).

A few comments though on minor things I'd like addressed:
* Could we please update `npm run clean` to **delete** the `src/decorator-registries` if it exists?  One issue that I've run into  is that the auto-created `src/decorator-registries` directory appears to be "permanent" (in that there's nothing that ever removes it).  This becomes annoying is you switch to a different branch (like `dspace-10_x`) as then your next commit will attempt to commit the `src/decorator-registries` folder.  We need a way to clean this directory up if you plan to switch branches.
* Do we need to write some basic docs on how to manage dynamic components going forward?  I'm trying to determine if there's developer documentation we need to be writing here to help us avoid a growing "main.js" in the future.  We do have a new [`CODE_CONVENTIONS.md`](https://github.com/DSpace/dspace-angular/blob/main/CODE_CONVENTIONS.md) doc we could add guidelines to.

Beyond that, this looks good & I verified that `main.js` decreases from about 770KB to 660KB with this change.

@FrancescoMolinaro and @atarix83 : Any concerns with merging this prior to upcoming DSpace-CRIS merger PRs?  I suspect it will create some code conflicts, but hopefully they'd be small and relatively easy to resolve.

alexandrevryghem:
@tdonohue: Thanks for the review! I updated the `npm run clean` command to delete that folder as well, and I also updated the `AbstractComponentLoaderComponent` by replacing the explanation about the manual decorator maps with the new auto-generated ones.

Regarding the permanent `src/decorator-registries` folder, I think we have two options:

- The simplest option is to add `/decorator-registries` to your global `.gitignore` file (this is what I currently do in my local setup)
- The other option would be to create a `post-checkout` Git hook. We can commit that Git hook folder to the repository, but users would still need to enable the hooks in their projects. It might also be a bit more tricky if they already have custom Git hooks

Regarding `CODE_CONVENTIONS.md`, I’m not sure whether this is something we need to document there as well or only on the wiki. If we do add it there, we should probably also include a section about lazy loading dependencies, since that is the other most commonly used way to reduce the size of `main.js`. I also noticed that the file is still referencing version 4 of Bootstrap.

FrancescoMolinaro:
@tdonohue this PR shouldn't have a big impact on the open/upcoming PRs if not for some minor conflicts and the reintroduction of the decorators, which shouldn't be a problem.

tdonohue (approved):
👍 Thanks @alexandrevryghem !  This looks good to me now.  I think updating `npm run clean` to delete the `src/decorator-registries` folder is a "good enough" solution for people working on different branches.  They can always just run `npm run clean` before switching branches if they want that folder removed.

I *am* going to label this as "needs documentation", as I think we should likely document this change in the [DSpace 11 docs](https://wiki.lyrasis.org/spaces/DSDOC11x/overview).  It does seem like a potential "breaking change" in v11 as sites may need to update their theme if they override any of the updated components.  Could you add some basic docs for this change to our v11 docs?

PR #4576, opened 2025-07-24
Task/main/duracom 317

FrancescoMolinaro:
## References
Fixes https://github.com/DSpace/DSpace/issues/8824
Based on https://github.com/DSpace/DSpace/pull/10684 to improve and better manage the audit information through the Event's details
Related to Rest Contract https://github.com/DSpace/RestContract/pull/311
Related to Rest Request https://github.com/DSpace/DSpace/pull/11072

## Description
Ui Implementation for Audits Overview, menu accessible in admin sidebar:

image

Implementation of context menu and related detail page for item:

image

## Instructions for Reviewers

For the backend setup please refer to https://github.com/DSpace/DSpace/pull/11072.

To configure the Audit Overview menu we added a new property enableAuditLogsOverview in app config, currently set to true as default.

The context menu on item page is configurable via rest property context-menu-entry.audit.enabled.

List of changes in this PR:

Implemented new page components for Audits overview
Implemented new page components for Audits deatail for item
Implemented presentational table component
Added e2e te

To test the PR, once both front and rest are installed, you need to do some change on an item, for instance add a metadata, the changes to the item will appear in the audit pages, containing detail of the changes made.

## Checklist


tdonohue:
General feedback from the[ DSpace Developers Meeting on Sept 25](https://wiki.lyrasis.org/display/DSPACE/2025-09-25+DSpace+Developers+Meeting): 
* Need to better understand how to "backup" Audit Trail if it's stored in Solr (only).  Also what is the process for upgrading Solr (which sometimes requires a full re-index of all content in Solr)?  Is it the same process as Solr Statistics, where you have to dump all the data out and reindex it? Can it use the same script as Solr Statistics, or do we need something new created?
* It'd be nice to have a search / filter option in the global audit page.  Otherwise, it's very difficult to find anything on that page.  This is especially useful if you need to locate the audit trail for a deleted object.
* Additional notes at https://wiki.lyrasis.org/display/DSPACE/2025-09-25+DSpace+Developers+Meeting

mwoodiupui:
Why are we using Solr as the primary store?  **Audit** data should be logged to the simplest, most bullet-proof format possible, one that is easy to replicate to offline storage, like a simple sequential file.  They can be indexed into a cache in batches by a back-end process.  The cache *could* be Solr, but why -- I see nothing here that uses text analysis.

tdonohue:
@FrancescoMolinaro : If you can find time, it'd be good to resolve the merge conflicts on this PR.  The backend PR is already at +1 approval, so it may be possible to move both forward together quickly if we can solve the conflicts here. Thanks!

config/config.example.yml
```diff
@@ -626,3 +626,7 @@ geospatialMapViewer:
 accessibility:
   # The duration in days after which the accessibility settings cookie expires
   cookieExpirationDuration: 7
+
+# Option to enable the menu entry to see audit logs on a site level.
+# Only site administrators will be able to use this functionality
+enableAuditLogsOverview: true
```
tdonohue: Could we consider removing this configuration and instead read the `audit.enabled` setting from the backend?  It seems like you'd want the menu entry to always exist whenever `audit.enabled` is true.  So, it'd be nice to avoid having two configurations.

cypress/e2e/audit-overview-page.cy.ts
```diff
@@ -0,0 +1,16 @@
...
+    cy.visit('/auditlogs');
+    cy.loginViaForm(Cypress.env('DSPACE_TEST_ADMIN_USER'), Cypress.env('DSPACE_TEST_ADMIN_PASSWORD'));
+  });
+
+  it('should pass accessibility tests', () => {
+    // Page must first be visible
+    cy.get('ds-audit-overview').should('be.visible');
+    // Analyze <ds-audit-overview> for accessibility issues
+    testA11y('ds-audit-overview');
+  });
```
tdonohue, line 25: I would appreciate it if we could add some more basic e2e tests for this page... even just check that the basic page structure exists.

src/app/shared/menu/providers/audit-item.menu.ts
```diff
@@ -0,0 +1,62 @@
...
+      ),
+    ]).pipe(
+      map(([isAdmin, isAuditEnabled]: [boolean, boolean]) => {
+        return [{
+          model: {
+            type: MenuItemType.LINK,
+            text: 'context-menu.actions.audit-item.btn',
+            link: '/auditlogs/object/' + dso.uuid,
+          } as LinkMenuItemModel,
+          icon: 'key',
```
tdonohue: I'm not a fan of reusing the `key` icon for the audit trail just because we already use it for the "Access Control" menu.  Could we look for a new icon?  Maybe [`fa-clipboard-check`](https://fontawesome.com/icons/clipboard-check?f=classic&s=solid), or something similar that is like a "checklist"...as that better represents an audit trail to me.

src/app/shared/menu/providers/audit-overview.menu.ts
```diff
@@ -0,0 +1,62 @@
...
+      map(([isSiteAdmin]) => {
+        return [
+          {
+            visible: isSiteAdmin && this.appConfig.enableAuditLogsOverview,
+            model: {
+              type: MenuItemType.LINK,
+              text: 'menu.section.audit',
+              link: '/auditlogs',
+            },
+            icon: 'key',
```
tdonohue: Same here, let's not use `key` and use something new.

src/app/audit-page/object-audit-overview/object-audit-overview.component.html
```diff
@@ -0,0 +1,25 @@
...
+    </div>
+    @if ((auditsRD$ | async)?.payload; as audits) {
+      <ds-audit-table
+        [audits]="audits"
+        [pageConfig]="pageConfig"
+        [object]="object"
+      ></ds-audit-table>
+    }
+    <button class="btn btn-light mt-3" (click)="goBack()">
+      <i class="fas fa-arrow-left"></i>{{ 'audit.object.back' | translate }}
```
tdonohue: We need to improve the spacing of the back arrow and the text "Back".  It's very squished right now (see my screenshot above).

src/assets/i18n/en.json5
```diff
@@ -3439,6 +3507,8 @@
 
   "menu.section.access_control_people": "People",
 
+  "menu.section.audit": "Audit Overview",
```
tdonohue: I think we might want to just call this "Audit Logs", as that's a better representation of what it is.  It's a list of all audit logs.

src/assets/i18n/en.json5
```diff
@@ -899,6 +899,72 @@
...
+
+  "audit.overview.table.timestamp": "Time",
+
+  "audit.overview.breadcrumbs": "Audit Logs Overview",
+
+  "audit.object.back": "Back",
+
+  "audit.object.breadcrumbs": "Subject Audit Logs",
+
+  "audit.object.overview.title": "Subject Audit Logs Overview",
```
tdonohue: As noted above, this could just say "Subject Audit Logs"

src/assets/i18n/en.json5
```diff
@@ -899,6 +899,72 @@
...
+
+  "audit.detail.metadata.confidence": "Confidence:",
+
+  "audit.detail.metadata.place": "Place:",
+
+  "audit.detail.metadata.action": "Action:",
+
+  "audit.detail.metadata.checksum": "Checksum:",
+
+  "audit.overview.title": "Audit Logs Overview",
```
tdonohue: I think this could just say "All Audit Logs"

src/assets/i18n/en.json5
```diff
@@ -899,6 +899,72 @@
...
+
+  "audit.overview.table.entityType": "Audit Type",
+
+  "audit.overview.table.eperson": "EPerson",
+
+  "audit.overview.table.other": "Other Object",
+
+  "audit.overview.table.timestamp": "Time",
+
+  "audit.overview.breadcrumbs": "Audit Logs Overview",
```
tdonohue: I think this could just say "Audit Logs"

src/assets/i18n/en.json5
```diff
@@ -899,6 +899,72 @@
...
+
+  "audit.detail.eventType": "Audit Type",
+
+  "audit.detail.eperson": "EPerson",
+
+  "audit.detail.timeStamp": "Time",
+
+  "audit.detail.back": "All Audit Logs",
+
+  "audit.detail.back.subject": "Subject Audit Logs",
```
tdonohue: I'm not sure all these `audit.detail.*` keys are used in this PR.  Could you double check?  I'm not seeing where `audit.detail.back` and `audit.detail.back.subject` are used, but maybe I'm missing it.

tdonohue (changes requested):
@FrancescoMolinaro : Thanks for this PR!  I finally had a chance to test this today.  Overall, it works well.  But, I did find a few things we likely should fix.  Some are described inline below, but I've also marked up this screenshot:

empty-audit-trail

The above is a screenshot of the audit trail for  a specific item.  A few improvements I'd like to see:
1. The breadcrumbs should include the Item page because this is a specific item (and we know its UUID).  So, I think we need to enhance these breadcrumbs
2. The title of the Item is displayed with no information about what it means.  I think we need to say something like "Logs for object: [name of object]" rather than just putting the item name with no explanation.  (Should we also consider renaming this page to be "Object Audit Logs" instead of "Subject Audit Logs"??)
3. The Back button is "squished".  The arrow overlaps the text.
4. In this case, this Item has no audit trail.  We should have an empty table displayed or some sort of message that says "No audit logs exist" (or similar).
5. In general, I also think you can remove the word "Overview" from all these pages.  I think it's better to just say "Subject Audit Logs" as the word "Overview" doesn't add anything. This is the same for the global page, it could just say "All Audit Logs" instead of "Audit Log Overview"

I have not had a chance to do a detailed code review yet. But, I've noted a few things I found in my testing inline below.  I'll try to do a more detailed code review next week.

src/app/shared/menu/providers/audit-item.menu.ts
mwoodiupui: A quick look turned up a somewhat common image:  a magnifying glass with a checkmark centered in the lens, perhaps overlaid on a stylized document.  Auditing seems to combine searching (the maginfier) and verification (the checkmark).  I searched for "icon meaning audit".

I also find the key a puzzling image here.

tdonohue: @mwoodiupui : Yes, that'd work too...however, I don't believe that icon is available in the free version of the fontawesome.com icons we're using.  That's why I suggested the clipboard with a checkmark, as it's in those free icons.

That said, either icon works for me.  I just was looking for something that we already have in our fontawesome icon pack.

src/app/audit-page/audit-table/audit-table.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    NgTemplateOutlet,
+    PaginationComponent,
+    RouterLink,
+    StringReplacePipe,
+    TranslateModule,
+    VarDirective,
+  ],
+  standalone: true,
+})
+export class AuditTableComponent {
```
tdonohue, line 52: Please add a code comment which describes where this component is used.  All components should have a basic description in either a comment or TypeDocs.

cypress/e2e/audit-overview-page.cy.ts
tdonohue: It still would be good to add more basic e2e tests here, testing that the basic page structure exists.

src/assets/i18n/en.json5
```diff
@@ -899,6 +899,54 @@
...
+
+  "audit.overview.table.subjectType": "Subject Type",
+
+  "audit.overview.table.entityType": "Audit Type",
+
+  "audit.overview.table.eperson": "EPerson",
+
+  "audit.overview.table.other": "Other Object",
+
+  "audit.overview.table.timestamp": "Time",
```
tdonohue: Should this say "Time (UTC)"?  It might make sense to clarify that these times all appear to be in UTC.  Either that, or change the format of the time to include the timezone.

tdonohue (changes requested):
@FrancescoMolinaro and @steph-ieffam :  Thanks for the updates here. I'm including both of you on this message as I've found some slightly odd behavior in more detailed testing today.  I'm not sure if it's coming from the frontend or backend.

First, thanks @FrancescoMolinaro for addressing my prior feedback.  This is looking good overall and the code looks good too, other than a _few minor inline comments noted below_.

In testing today, this feature is working well, but I've found a few bugs or odd behaviors:

1.  First, when `audit.enabled=false` on the backend, the Audit menu on items is wrongly displayed.  This should not appear if the feature is disabled. (The main audit admin sidebar menu is working properly though and doesn't appear when disabled)

audit-menu

2. When I create a new submission by importing a publication from CrossRef (not sure if the import matters here), the audit logs for that one action are **massive**.  I end up with 56 entries in the audit log, all of them `MODIFY_METADATA` actions.  Some are also very odd as I'll see the same field listed twice, both for `REMOVE` and `ADD` (even though I never removed the field).  Here's what it looks like:

submission

In this scenario I literally just imported metadata from CrossRef, added any missing required fields, uploaded a file and submitted it.  I didn't make any metadata modifications.  Oddly, too, I don't see information about the file I uploaded in the audit trail for this newly submitted item.  It's just 56 entries for `MODIFY_METADATA`.

My expectation is that Item *creation/submission* doesn't need this level of detail.  I was expecting to see a much smaller list of audit actions.  Perhaps just an Item `CREATE` action and a Bitstream `CREATE` action.  I'm not sure this detail about each metadata edit is necessary during submission.  

3. I also tested uploading a *new* bistream to an existing archived Item.   In this case, I also found the audit trail to be overly detailed and duplicative: 

bitstreamupload

Again, I'm a little surprised that we have `METADATA_MODIFY` audit entries here as the file was just added.  However, what's more surprising is we have **two of the same** `METADATA_MODIFY` entries followed by **two of the same** `CREATE` entries.


Overall, again, I think this is working.  I'm just worried that if we log **every single** metadata modification during submission it makes the audit logs extremely long/detailed. In my opinion the audit logs are more useful if they track `CREATE` for new objects, and then track modifications **after creation** in more detail. 

If we need to bring this up for discussion in a future meeting, let me know.  Overall though, I do want to clarify that I like this feature.  Just trying to understand why there are so many unexpected entries being generated.

steph-ieffam:
Hey @tdonohue thanks for spotting these issues. Most of what you have reported was actually a REST issue, in the meantime @FrancescoMolinaro has addressed all the issues on the Angular side.
Getting back to the issues identified let me reply to each point:

> When I create a new submission by importing a publication from CrossRef (not sure if the import matters here), the audit logs for that one action are massive. I end up with 56 entries in the audit log, all of them MODIFY_METADATA actions. Some are also very odd as I'll see the same field listed twice, both for REMOVE and ADD (even though I never removed the field). 

We spotted an issue on the dispatching of the events so the same set of modified metadata was being dispatched multiple times even though these changes were already stored and shouldn't have been processed twice. This has been introduced by mistake during the refactoring for the review so we didn't notice it at fist. Now the set of information which is tracked is way smaller and doesn't contain duplicate information
Also in your comment you mentioned about a metadata removal which you didn't perform that was being tracked. We managed to discover that there was a pre-existing logic which removed **each** single metadata for the **dc.date.issued** field just to replace a "today" value. Now the logic has been improved to just to modify the "today" values if present instead of dropping and re-adding all the values for **dc.date.issued**. 

> UPDATE: I should add that I have audit.item.in-workspace set to the default of false

I can confirm the above issue was happening right after the item was moved to the archived state. Plenty of duplicated **dc.description.provenance**, **dc.identifier.uri** and **dc.date.accessioned** was being tracked. You will still see this information in the audit trail since these metadata are currently added right after the item is set to archived but now the information is only displayed once for each modification.

> I also tested uploading a new bitstream to an existing archived Item. In this case, I also found the audit trail to be overly detailed and duplicative

We indeed noticed the creation Event was being tracked twice and also some information set on the newely created bitstream/bundle was being tracked. We just introduced a fix to track the creation of the objects once and skip the tracking of information about the related objects (bundles & bitstreams) while the item is still in the workspace.

During our tests we managed to reduce the number of tracked events from 56 events (clearly something odd was happening) to just 11 event (plenty of bitstreams were also added in our test case).

Please let us know if anything odd is still left behind and needs to be changed.
Thanks for your review

cypress/e2e/audit-overview-page.cy.ts
tdonohue: Again, it'd be good to add more basic e2e tests here, if possible. I'd like to see checks that the basic page structure exists.  Or, you could potentially perform an action (e.g. add a new metadata field) and verify it shows up in the audit trail.

FrancescoMolinaro:
Hi @tdonohue , I have addressed the label changes as requested, regarding the e2e I think git is playing some trick, the page structure should have been added with this commit: https://github.com/DSpace/dspace-angular/pull/4576/commits/8c5bb1d7ff9907e0adc72188bf2047b7f26813a3.

For the test you mentioned, changing a metadata and checking the UI for the entry, I believe we would need the Audit functionality for the REST container used in the e2e, so i would suggest it as a later improvement of the e2e test once the whole functionality has been merged.

If this is not the case then I will proceed to add to the test the logic you mentioned.

steph-ieffam:
Hey @tdonohue, I apologize for the late reply.
We have pushed some fixes for the reported bug and now it should be fixed correctly.
Regarding the duplication of the dc.identifier.uri this is not occurring in our tests and I have all the properties set to true while testing. Could you kindly check again and let us know if the bug is solved?
We tested with both "direct" submission and import from Crossref and it seems there's no issue of duplication in the audit.

As far as I can see the current situation is ok

[Screencast from 2025-12-11 15-17-55.webm](https://github.com/user-attachments/assets/fceb13dc-c88b-4ef5-93d9-46967ba8b9ae)

**the next one is another test**

screencapture-localhost-4000-entities-publication-2fb16b49-329d-412b-8f2c-19b6a15e7252-auditlogs-2025-12-15-11_32_31
screencapture-localhost-4000-entities-publication-2fb16b49-329d-412b-8f2c-19b6a15e7252-auditlogs-2025-12-15-11_32_51

paulo-graca:
I'm seeing a lot of MODIFY_METADATA entries, one for each field and at the same time? Could it be possible to group them in a single entry (grouped by timestamp, for instance)?

steph-ieffam:
> I'm seeing a lot of MODIFY_METADATA entries, one for each field and at the same time? Could it be possible to group them in a single entry (grouped by timestamp, for instance)?

Hey @paulo-graca thanks for the feedback. The changes are not easy/quick to implement and moreover this would actually change hugely the UI from what currently it is and it's something that in my opinion would require a discussion.

Furthermore, this would need REST changes to properly address the issue without impacting and slowing down the Angular application

tdonohue (approved):
👍 Thanks @FrancescoMolinaro !  I retested this today and all my prior feedback has been addressed.  I'll be creating one small follow-up ticket to fix the bug with `dc.identifier.uri` (as it still exists but only if `identifiers.submission.register = true`).  But, this PR is good enough to merge.  Thanks again!

PR #4732, opened 2025-09-22
Preserve scroll position when opening/closing the relationship modal

### LINKED ISSUE #3303: Opening a Lookup Relation Modal always brings you to the top of the Submission Form
## Describe the bug
Opening a lookup relation modal like of the configured author field (person entity) let jump the submission form at the top of the page.

In most cases, especially at the DSpace Demo Site, only the author relation field is configured at the top and therefore does not interfere. But if relation fields to other entities are configured further down, then it is very annoying for the user to scroll back to the last edited submission field.

Affects all versions of DSpace 7 and 8. At least the browsers Firefox and Chrome.

## To Reproduce
1. Start a Submission with any user role at the DSpace Demo Site (https://demo.dspace.org/)
2. Scroll down to just above the author field like in this Screenshot and open the Lookup Modal with the magnifier icon:
image

3. See the result of the main submission form behind:
image

4. Close the Modal...same result.

## Expected behavior
The position of the modal should be at the height of the field and not at the top edge of the submission.

## Related work
Could not find any DSpace issues dealing with this or similar topics

Wout-atmire:
## References

- Fixes #3303

## Description

Fixes the issue of being redirected to the top of the page after opening a modal, by adding a fragment to the URI of the original position were the modal was opened.

## Instructions for Reviewers

List of changes in this PR:
* Components that may call `this.router.navigate` when opening a modal are given `@Input() retainScrollPosition = false;`
    - For example starting from `dynamic-lookup-relation-search-tab.component.html`
    - This is passed down to more general child-components eg. `ds-search-sidebar`
    - The idea is to make it so that the child-components know if it is fine to reload to the top of the page or not.
* Introduced `scroll.service.ts` allows to scroll to a specific fragment using JS.
    - On navigation angular automatically redirects to the fragment defined in the URI, but Angular doesn't keep the `scroll-margin-top` which prevents the sticky header from overlapping our fragment in mind.
    - To select a target fragment `ScrollService.setFragment` is used.
* If no custom fragment has been set, a fragment of a non-existing id is added: "prevent-scroll"
    - This is to prevent any scrolling to be performed.

#### Testing
1. As submitter navigate to the submission forms on a "Publication" collection.
2. Scroll down and open a modal (eg. for an Author)
    image (1)
3. When openning the modal the scroll position should not be reset to the top of the page.
4. Inside the modal, changing pagination options that edit the URL shouldn't remove `#prevent-scroll` fragment from the URL. And the scroll position should not be reset to the top of the page.
    - Adding search filters
    - Reseting search filters
    - Changing page to another page
    - Changing search order
    - Changing amount of page results

## Checklist


jlipka:
I just tested the behaviour, and the scroll position  seems to be preserved while using Chrome and Edge browser, but using Firefox the ui still jumps to the top of the page. (Linux / ubuntu)

kanasznagyzoltan:
For me is working: #prevent-scroll is appearing in the background and the mainform behind the modal is not jumping back to the top.

https://github.com/user-attachments/assets/1bc074d9-9902-4499-bc22-905edca83da0

nathanmlf (commented):
Hey @Wout-atmire! I've tested it on Chrome just like @kanasznagyzoltan and on Firefox like @jlipka . I had the same problem in Firefox, but it works as expected in Chrome.

Just to add to the investigation, my tests were on Windows 11 (I don't think the Firefox issue is specific to Linux). I've attached both videos below.

Firefox (Version 150.0.3 - 64 bits)

https://github.com/user-attachments/assets/32469921-bbcd-4f90-83f8-009e3119b03a

Google Chrome (Version 148.0.7778.97 - 64 bits)

https://github.com/user-attachments/assets/aa6af9d0-ee6c-4566-8ee0-5fad5f86bbe1

Wout-atmire:
Given [#6108](https://github.com/DSpace/dspace-angular/pull/6108), scroll position retention in submission may be unnecessary, so I’m putting this PR on hold until that is merged and we can confirm whether anything is still needed here.

Credits to @alexandrevryghem for pointing this out

PR #4840, opened 2025-11-19
Update js-cookie to v3 and refactor cookie services for compatibility (#4833)

### LINKED ISSUE #4833: Update js-cookie to version 3
We need to move to the next major version of js-cookie:

- `js-cookie` version 3.0.5
- `@types/js-cookie` version 3.0.6

Version 3.0.0 was released over four years ago and requires some code changes for us to integrate. See the [v3.0.0 release notes](https://github.com/js-cookie/js-cookie/releases/tag/v3.0.0) for more information.

Roshan-Singh-07:
This PR updates `js-cookie` to version 3 as requested in #4833.
Fixes #4833

### Summary of changes
- Updated `js-cookie` to v3 in `package.json`.
- Adjusted `ClientCookieService` to use the v3 API (`Cookies.get`, `set`, `remove`).
- Added safe JSON parsing to keep the previous behavior.
- Updated `ServerCookieService` to match the new cookie format.

### Notes
The previous `getJSON` helper was removed in v3, so the parsing logic is now handled internally.

Tested basic flows (login, logout, redirect cookie, impersonation cookie) and everything continues to work as expected.

Let me know if anything else needs adjusting.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -493,7 +493,6 @@
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
       "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
       "dev": true,
-      "peer": true,
       "dependencies": {
         "fast-deep-equal": "^3.1.1",
         "fast-json-stable-stringify": "^2.0.0",
@@ -739,7 +738,6 @@
           "url": "https://github.com/sponsors/ai"
         }
       ],
-      "peer": true,
       "dependencies": {
         "nanoid": "^3.3.7",
         "picocolors": "^1.0.1",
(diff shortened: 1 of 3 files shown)


tdonohue:
@Roshan-Singh-07 : Thanks for this PR.  However, I wanted to note that this PR is currently failing our `lint` checks.  Could you do the necessary cleanup that is required?  You should be able to run `npm lint` locally to see the necessary changes. Thanks!

Roshan-Singh-07:
Screenshot 2025-11-20 180333
Applied the necessary linting fixes to the cookie service files. Everything now passes local lint checks.

alanorth:
Thanks @Roshan-Singh-07. I tested this locally and I'm not able to log in anymore.

@tdonohue the `peer: true` changes to `package-lock.json` are probably because @Roshan-Singh-07 used a newer NPM than what our lockfile was generated with (likely from Node 22+). Installing this on an older NPM still works from what I can see, but this does create some churn in the changeset. And on that note, I would request that either @Roshan-Singh-07 or we squash the changes into one commit before merging so we get a clean commit history.

Roshan-Singh-07:
> Thanks @Roshan-Singh-07. I tested this locally and I'm not able to log in anymore.
> 
> @tdonohue the `peer: true` changes to `package-lock.json` are probably because @Roshan-Singh-07 used a newer NPM than what our lockfile was generated with (likely from Node 22+). Installing this on an older NPM still works from what I can see, but this does create some churn in the changeset. And on that note, I would request that either @Roshan-Singh-07 or we squash the changes into one commit before merging so we get a clean commit history.

Thanks for checking this. Sure, I'm fine with squashing the commits. I'll also look into the login issue and update the PR accordingly.

tdonohue:
Replaced by #4849

PR #4879, opened 2025-11-27
Feature/make log in password component themeable

lrshd:
## Description
The log-in-password.component within the log-in.component is not yet themeable. I updated/added the necessary files to add this functionality. 

## Instructions for Reviewers
List of changes in this PR:
* Updated the log-in-password.component.ts within the src/app/ directory to be a base component
* Created the themed-log-in-password.component.ts in the same directory with the necessary functions
* Updated auth-methods.type.ts and log-in.methods-decorator.ts to reference the new themed component
* Created the necessary log-in-password.component.ts, log-in-password.component.scss and log-in-password.component.html in the src/themes/custom folder
* Updated the custom themes eager-theme.module.ts to include the new themed component
* Updated the log-in-container.component.spec.ts in src/app to include the necessary imports for themed components

This PR can be tested by creating a local theme based on the custom theme folder and customizing the styling of the log-in-password.component

### DIFF
--- a/src/app/shared/log-in/methods/password/themed-log-in-password.component.ts
+++ b/src/app/shared/log-in/methods/password/themed-log-in-password.component.ts
@@ -10,7 +10,6 @@ import { LogInPasswordComponent } from './log-in-password.component';
   selector: 'ds-log-in-password',
   styleUrls: [],
   templateUrl: '../../../theme-support/themed.component.html',
-  standalone: true,
   imports: [
     LogInPasswordComponent,
   ],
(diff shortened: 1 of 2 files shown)


alexandrevryghem:
The `LoginComponent` is a dynamic component that is rendered using decorators (see `log-in.methods-decorator.ts`). So the correct way to theme it is by adding you new component inside that map. The decorator itself is not themeable yet, so you will have to replace the import from the base component with the one from your custom theme. I already opened a PR that made it themeable like other themeable decorators here, but it hasn't been merged yet: #4528

PR #4898, opened 2025-12-01
Integrate @ng-dynamic-forms in code base and upgrade ng-bootstrap version.

FrancescoMolinaro:
## References

Fixes https://github.com/DSpace/dspace-angular/issues/1272

## Description

This PR brings the code from the @ng-dynamic-forms library directly into the DSpace codebase (under src/external-libraries). The upstream library is no longer maintained, and its outdated dependencies were blocking important framework upgrades (e.g., ng-bootstrap).

By decoupling from the external library, we can now upgrade those dependencies and continue improving the components and services originally provided by @ng-dynamic-forms. As part of this migration, the code has already been optimized for newer Angular versions, and all templates have been migrated to use Angular’s control flow syntax.

A notable improvement resulting from the upgrade is the resolution of the accessibility issue related to ngbTypehead. The previously reported aria-multiline violation detected by axe DevTools is no longer present.

Old version:

image

After upgrade:

image

The code is under ISC license so all the files will contain a copy of the license text specified in the repo: https://github.com/udos86/ng-dynamic-forms?tab=ISC-1-ov-file

## Instructions for Reviewers

This PR does not introduce new features. All library components should behave and appear the same as before. The focus is strictly on migrating and updating the code.

List of changes in this PR:

Integrated the relevant portions of @ng-dynamic-forms (core and ui-ng-bootstrap) into the repository under src/external-libraries/ng-dynamic-forms/.

Updated tsconfig paths and imports across the application:

"@ng-dynamic-forms/core/*": ["src/external-libraries/ng-dynamic-forms/core/*"]
"@ng-dynamic-forms/ui-ng-bootstrap/*": ["src/external-libraries/ng-dynamic-forms/ui-ng-bootstrap/*"]

Upgraded ng-bootstrap to v15.1.2 and updated imports where needed (module → standalone).

Migrated all library templates to Angular's control flow syntax.

Applied ESLint fixes where possible within the migrated library code.

Test:

To verify the fix for the aria-multiline accessibility issue:

1) Use axe DevTools on any form containing:
     a dynamic-onebox component, or
    any component using the ngbTypehead directive.

2) In the current main branch, axe should report the aria-multiline violation.

3) When testing the same form with this PR, the violation should no longer appear.

## Checklist

### DIFF
--- a/src/assets/i18n/en.json5
+++ b/src/assets/i18n/en.json5
@@ -2133,54 +2133,6 @@
 
   "form.other-information.orcid": "ORCID",
 
-  "form.other-information.oairecerif_author_affiliation": "Affiliation",
-
-  "form.other-information.oairecerif_editor_affiliation": "Affiliation",
-
-  "form.other-information.person_identifier_orcid": "ORCID iD",
-
-  "form.other-information.institution-affiliation-name": "Affiliation(s)",
-
-  "form.other-information.dc_relation_grantno": "Grant Number",
-
-  "form.other-information.not-available": "Not available",
-
-  "form.other-information.ror_orgunit_id": "ROR ID",
-
-  "form.other-information.ror_orgunit_type": "ROR type",
-
-  "form.other-information.ror_orgunit_acronym": "ROR acronym",
-
-  "form.other-information.ror_orgunit_countryName": "ROR country",
-
-  "form.other-information.oairecerif-author-affiliation": "Affiliation",
-
-  "form.other-information.oairecerif-editor-affiliation": "Affiliation",
-
-  "form.other-information.dc.publisher": "Publisher",
-
-  "form.other-information.dc.identifier.issn": "ISSN",
-
-  "form.other-information.alternative-names": "Alternative names",
-
-  "form.other-information.selection.data-oairecerif_author_affiliation": "Select alternative affiliation",
-
-  "form.other-information.data-oairecerif_author_affiliation": "Affiliation",
-
-  "form.other-information.selection.alternative-names": "Select alternative name",
-
-  "form.entry.source.local": "",
-
-  "form.entry.source.orcid": "",
-
-  "form.entry.source.ror": "",
-
-  "form.entry.source.openaire": "",
-
-  "form.entry.source.zdb": "",
-
-  "form.entry.source.sherpa": "- Sherpa Romeo",
-
   "form.remove": "Remove",
 
   "form.save": "Save",
--- a/src/assets/i18n/ar.json5
+++ b/src/assets/i18n/ar.json5
@@ -1349,106 +1349,6 @@
   // "admin.batch-import.page.remove": "remove",
   "admin.batch-import.page.remove": "إزالة",
 
-  // "audit.data.not-found": "No audits found.",
-  // TODO New key - Add a translation
-  "audit.data.not-found": "No audits found.",
-
-  // "audit.data.self": "self",
-  // TODO New key - Add a translation
-  "audit.data.self": "self",
-
-  // "audit.detail.metadata.field": "Metadata field:",
-  // TODO New key - Add a translation
-  "audit.detail.metadata.field": "Metadata field:",
-
-  // "audit.detail.metadata.value": "Value:",
-  // TODO New key - Add a translation
-  "audit.detail.metadata.value": "Value:",
-
-  // "audit.detail.metadata.authority": "Authority:",
-  // TODO New key - Add a translation
-  "audit.detail.metadata.authority": "Authority:",
-
-  // "audit.detail.metadata.confidence": "Confidence:",
-  // TODO New key - Add a translation
-  "audit.detail.metadata.confidence": "Confidence:",
-
-  // "audit.detail.metadata.place": "Place:",
-  // TODO New key - Add a translation
-  "audit.detail.metadata.place": "Place:",
-
-  // "audit.detail.metadata.action": "Action:",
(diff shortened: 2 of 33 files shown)


FrancescoMolinaro:
Hi @artlowel , thank you for the feedback. I agree with the proposed approach, and as you noted, my initial thought was also to upgrade the library within its own repository.

I raised this topic in the last meeting, and we aligned on proceeding in this direction.  We will then create a temporary npm package that will be maintained until the migration away from this library is completed.

I’ll follow up with a new PR as soon as possible. 
Thanks again for the review and the helpful input.

FrancescoMolinaro:
Hi @artlowel , i have updated the PR as we discussed by publishing the updated library.
The packages can be found here:

https://www.npmjs.com/package/@4science_ng-dynamic-forms/core
https://www.npmjs.com/package/@4science_ng-dynamic-forms/ui-ng-bootstrap

The import paths have been configured as the old ones in the tsconfig, so to avoid changes in the imports.
Accordingly to the scope of the PR I have then updated ng-bootstrap and ngx-mask, as it depended on the first one.

I would really appreciate if you could have another look to the PR.

package.json
```diff
@@ -66,16 +66,11 @@
       "@angular/forms": "^20.3.12",
       "@angular/localize": "^20.3.12"
     },
-    "@ng-dynamic-forms/core": {
+    "@4science_ng-dynamic-forms/core": {
       "@angular/common": "^20.3.12",
       "@angular/core": "^20.3.12",
       "@angular/forms": "^20.3.12"
     },
```
artlowel, line 73: Are these overrides still necessary now?

FrancescoMolinaro: Hi @artlowel , they are still necessary as I didn't upgrade the Angular version of the library, wich is still on v16.
Since we plan anyway to move away from the library I don't know if it makes sense to spend time to upgrade it.
In case necessary we could handle that in a separate task as I believe is not trivial, what do you think?

artlowel: No, you're right it's not worth the effort to upgrade it

artlowel (approved):
Thanks @FrancescoMolinaro!

tdonohue (commented):
@FrancescoMolinaro : I gave this a test & review today.  Overall it looks good and it works well.  However, I've found it only **partially fixes** #1272.

There are two parts to #1272:
1. The `aria-multiline` attribute is invalid on `combobox` role (it currently appears for `dc.subject` field). **(FIXED by this PR)**
2. The `<select>` for the `dc.identifier` field doesn't have a `<label>` or an `aria-label` attribute.  (**NOT FIXED by this PR**)
    * In fact you can see this error still in your screenshot above where axe DevTools still returns a "Select element must have an accessible name" error.
    
I suspect the additional fix here actually needs to be applied in the new `@4science_ng-dynamic-forms` library.  But, I'm assuming that will require a new version which may need updating here.

Beyond that, this approach looks good and the code looks good.

tdonohue:
NOTE: I also checked the versions of `ng-dynamic-forms` and `ng-bootstrap` on all maintenance branches.   It looks like we may only be able to easily backport this to 9.x because we upgraded to Boostrap 5.x in that release.  8.x and 7.6.x ports don't seem to be easily possible.

So, I'm only flagging this for backport to 9.x.

FrancescoMolinaro:
Hi @tdonohue , you are right I overlooked the label issue, I upgraded the library so that the select element will have an aria-label if a label or a placeholder is configured in the model, so that the accessibilty issue will be solved:

image

src/assets/i18n/en.json5
```diff
@@ -2133,6 +2133,54 @@
 
   "form.other-information.orcid": "ORCID",
 
+  "form.other-information.oairecerif_author_affiliation": "Affiliation",
+
+  "form.other-information.oairecerif_editor_affiliation": "Affiliation",
+
+  "form.other-information.person_identifier_orcid": "ORCID iD",
```
tdonohue: @FrancescoMolinaro : There are now a large number of unrelated labels added to this PR.  Could we remove them?

tdonohue (changes requested):
@FrancescoMolinaro : I've verified the accessibility bugs are all fixed, but I'd like to ask you to remove the final commit in this PR.  It's adding a large number of *unrelated* labels to the `en.json5` file (see comment inline below) along with other files. 

In general, in a PR, you should ONLY add new labels that pertain to the PR (and they are only required in `en.json5`).  Even though the other `*.json5` files are outdated, we resync them with every release...so there's no need to resync them in this PR.

(Resyncing all the i18n files will also make this PR more difficult to backport to 9.x, as those will result in merge conflicts on that branch)

So, if you remove the final commit, it will make this PR much smaller (currently it is >20,000 lines) & easier to backport.  Then only add labels that are required by this PR.

Beyond that, this looks great!  It works and fixes the issues.

FrancescoMolinaro:
Hi @tdonohue, you are absolutely right,  my mistake, I committed on the wrong branch, I have reverted the changes of the commit so should be ok now.

Many thanks for the catch!

strawburster:
@FrancescoMolinaro please update the git url and other package.json metadata for the npm packages.

FrancescoMolinaro:
Hi @strawburster , @tdonohue  I have created a separate issue for the update of the library, you can find the related work at the following links:

https://github.com/DSpace/dspace-angular/issues/5083
https://github.com/DSpace/dspace-angular/pull/5084

The relevant metadata should now all point to the correct repo: https://github.com/4Science/ng-dynamic-forms

PR #4904, opened 2025-12-03
Submission - Replace plaintext with relationship on lookup - 7.6

Atmire-Kristof:
## References
* Fixes #4843
* main equivalent: #4911

## Description
This PR changes the submission relationship lookup to make the (first) selected relationship replace the plain text value the lookup started from, if it did.

## Instructions for Reviewers
Technical details:
* If the lookup is started from a plain-text metadata value, its "place" and "metadataField" is stored on the lookup modal
* When a relationship is selected for the first time in the modal and those "replace" properties are present, a `ReplaceRelationshipAction` will be created instead of the usual `AddRelationshipAction`. This only happens for the first selected relationship, the rest sends a `AddRelationshipAction` for each selection.
* Relationship-effects takes the following actions in order, when it recieves a `ReplaceRelationshipAction`:
  * Remove the plain-text metadata value from the item
  * Create the new relationship
  * Update the new relationship's place to be on the deleted value's spot

How to test:
* Start a submission and for a relationship-controlled field (e.g. author), create multiple plain-text values
* Start a lookup from one of the plain-text values
* Select a couple new relationships in the lookup modal (at least two)
* Close the lookup
* Verify the plain-text value you started the lookup from is not present anymore
* Verify the first selected relationship is placed where the removed plain-text value was
* Verify all other selected relationships are placed at the end of the list

## Checklist
_This checklist provides a reminder of what we are going to look for when reviewing your PR. You do not need to complete this checklist prior creating your PR (draft PRs are always welcome).
However, reviewers may request that you complete any actions in this list if you have not done so. If you are unsure about an item in the checklist, don't hesitate to ask. We're here to help!_

- [ ] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [x] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [x] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
[...]

PR #4911, opened 2025-12-03
Submission - Replace plaintext with relationship on lookup - main

### LINKED ISSUE #4843: Plaintext author is not cleared after selecting entity through lookup
## Describe the bug
The relationship field does not remove the old plaintext Author value when you replace it with an entity in the `DsDynamicLookupRelationModalComponent`.

## To Reproduce
1. Set `orcid.api-url = https://pub.sandbox.orcid.org/v3.0` in your `local.cfg` (this way you can easily import new authors from ORCID)
2. Import/create a workspace item with some authors in a collection with entity type `Publication` (this way the author field will already be correctly set up)
3. Click on the lookup button next to one of the plain text authors
4. Select an author from the Local Authors tab or from the ORCID tab if you don't have any locally
5. Close the modal

https://github.com/user-attachments/assets/60ee9182-fa91-44d2-8287-17ff7f54e176

## Expected behavior
The new entity relationship should be created AND the old plain text value should be removed

Atmire-Kristof:
## References
* Fixes #4843
* 7.6 equivalent: #4904

## Description
This PR changes the submission relationship lookup to make the (first) selected relationship replace the plain text value the lookup started from, if it did.

## Instructions for Reviewers
Technical details:
* If the lookup is started from a plain-text metadata value, its "place" and "metadataField" is stored on the lookup modal
* When a relationship is selected for the first time in the modal and those "replace" properties are present, a `ReplaceRelationshipAction` will be created instead of the usual `AddRelationshipAction`. This only happens for the first selected relationship, the rest sends a `AddRelationshipAction` for each selection.
* Relationship-effects takes the following actions in order, when it recieves a `ReplaceRelationshipAction`:
  * Remove the plain-text metadata value from the item
  * Create the new relationship
  * Update the new relationship's place to be on the deleted value's spot

How to test:
* Start a submission and for a relationship-controlled field (e.g. author), create multiple plain-text values
* Start a lookup from one of the plain-text values
* Select a couple new relationships in the lookup modal (at least two)
* Close the lookup
* Verify the plain-text value you started the lookup from is not present anymore
* Verify the first selected relationship is placed where the removed plain-text value was
* Verify all other selected relationships are placed at the end of the list

## Checklist


jrcwiok:
- [x] Start a submission and for a relationship-controlled field (e.g. author), create multiple plain-text values 
- [x] Start a lookup from one of the plain-text value
- [x] Select a couple new relationships in the lookup modal (at least two)
- [x] Close the lookup
- [x] Verify the plain-text value you started the lookup from is not present anymore
- [x] Verify the first selected relationship is placed where the removed plain-text value was
- [x] Verify all other selected relationships are placed at the end of the list

I interpreted the steps as the following:
 I tested on the metadata tab using the author element. I entered plain text author names, clicked on the lookup, and replaced with a value in the list.

jrcwiok (approved):
- [x] Start a submission and for a relationship-controlled field (e.g. author), create multiple plain-text values 
- [x] Start a lookup from one of the plain-text value
- [x] Select a couple new relationships in the lookup modal (at least two)
- [x] Close the lookup
- [x] Verify the plain-text value you started the lookup from is not present anymore
- [x] Verify the first selected relationship is placed where the removed plain-text value was
- [x] Verify all other selected relationships are placed at the end of the list

I interpreted the steps as the following:
 I tested on the metadata tab using the author element. I entered plain text author names, clicked on the lookup, and replaced with a value in the list.

Atmire-Kristof:
Some clarification to my latest changes:

There was a bug with selecting multiple relationships in one action (e.g. selecting the entire page of results at once, using the dropdown). The modal would emit `selectObject` more than once per selected object because it was being called from two places:
- `selectPage` was calling `this.selectObject.emit` manually for the selected page
- Because `selectPage` also calls `this.selectableListService.select`, `ds-search` would internally fire another `selectObject` event for the same objects

This caused the multiple select to behave in very strange ways, where only one of the selected objects would actually be selected, the others were ignored.

The easiest fix here is to remove the manual emission that `selectPage` does and just let the results of `this.selectableListService.select` do their thing.

tdonohue (changes requested):
@Atmire-Kristof : I'm testing this today and the behavior is still a bit buggy to me.  That said, the basic behavior you describe works.  If I select *multiple* relationships, it works fine. But, if I select just *one* relationship, it doesn't seem to work.

For instance, this is what I tried which **doesn't work**:
1. Create a new Publication
2. Add two plain text authors to that publication
3. Next to one author, click the search button. Select *one* relationship to add instead
4. Go back to the submission form. The plain text author still exists
5. Click the search button again. The relationship appears selected still.  Close the popup
6. Now there are two copies of the same plain text author back on the submission form.

However, this *does work*:
1. Create a new Publication
2. Add two plain text authors to that publication
3. Next to one author, click the search button. Select *two* relationships.
4. Now, back on the form, the plain text author is replaced with the relationship and the other relationship is listed at the end (as a third author).

So, this PR appears to be a _partial fix_.  It works if you always select multiple relationships. But, if you try to do a 1-to-1 replacement, then it doesn't work. So, it's not a complete fix for #4843

Atmire-Kristof:
@tdonohue Thank you for the review,

meanwhile, we had also discovered similar issues and it turns out there was an issue with how I tried to remove the old plain-text metadata from the submission item. I implemented it as a PATCH on the ITEM, not on the submission object (WorkspaceItem or WorkflowItem).

I have now reworked it to perform a PATCH on the submission object instead (patch on the section metadata). To achieve this, I modified the submission-object service(s) to allow patches and extended the replace-relationship-action and effects to pass the section ID.
I've also made tracking the place of the to-be-replaced plain-text metadata value more reliable by passing the array index along, since there were instances where the metadata value had not recieved a proper place yet when the lookup happens.

Also updated this PR with the latest main.

tdonohue (changes requested):
@Atmire-Kristof : I tried testing this today again, and it's still not working for me when I tried to replace a single author.

Now, if I try to replace a plain string author...initially nothing happens. But if I open up the relationship lookup window again, the author is duplicated.

Here's a video of the behavior I'm seeing:

https://github.com/user-attachments/assets/528e46c9-8a7c-4308-adc3-bcf6a73f6035

Atmire-Kristof:
@tdonohue , thank you for the very clear example on how to reproduce the issue. I was able to replicate it on a local setup and found a fix for the issue.

The issue being that there's a check on the end of the update-and-refresh-form cycle to ensure we're only refreshing when a change actually occurred. This check only triggers when the changed value's "value" has been changed. This is not the case for us, as the "value" (in your case, "Donohue, Tim") doesn't change, but only the authority key does.

My fix is to change this check to also accept metadata where any of the other properties were changed, such as language, authority, etc.

tdonohue:
@Atmire-Kristof : Just a quick note. It looks like this PR has linting issues now.

PR #4814, opened 2025-11-14
[DSpace-CRIS] Creation of a custom URL for Items (Frontend)

FrancescoMolinaro:
## References

Related to https://github.com/DSpace/RestContract/pull/318
Related to https://github.com/DSpace/DSpace/pull/11549

## Description

This PR migrates the custom URL feature from DSpace-CRIS to core DSpace, allowing to define a custom url of item in submission. 

## Instructions for Reviewers

The custom URL feature allows items to have human-friendly URLs instead of UUID-based paths.
To test it a rest config for the newly introduced section must be in place.

ref: https://github.com/DSpace/DSpace/pull/11549#issue-3622340959

Once the section is configured the user can provide a value for the custom url in submission.
After the item has been deposited it will be possible to access it via the custom url, e.g. /entities/person/customUrlValue

List of changes in this PR:

Adapted item resolver.
New submission section for custom url.

Additional change:

Fixed sync-i18n script (broken after latest updates on main)
Executed sync on labels.

## Checklist
_This checklist provides a reminder of what we are going to look for when reviewing your PR. You do not need to complete this checklist prior creating your PR (draft PRs are always welcome).
However, reviewers may request that you complete any actions in this list if you have not done so. If you are unsure about an item in the checklist, don't hesitate to ask. We're here to help!_

- [ ] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [ ] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [ ] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [ ] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [ ] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [ ] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [ ] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC9x/Accessibility)** if it makes changes to the user interface.
- [ ] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
[...]

PR #4956, opened 2025-12-19
[DSpace-CRIS] Porting of authority framework functionalities (Frontend)

### LINKED ISSUE #11748: [DSpace-CRIS] Data model preparation work (Authority Framework enhancements, etc)
> [!NOTE]
>  This ticket represents a **work package** to be completed as part of the [merger of DSpace-CRIS into DSpace](https://dspace.org/announcement-dspace-dspace-cris-merger-approved-by-dspace-governance/).  A work package is a series of related changes which must be created and tested together.

# Description

This ticket describes the **first** work package of the DSpace-CRIS merger into DSpace.  It represents a set of changes and features that prepare the system for applying data model changes to DSpace from DSpace-CRIS.  Nonetheless, this is a standalone work package and all changes are self contained and testable.

This "Data model preparation" work package includes the following (as described in the [DSpace-CRIS and DSpace Differences](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences)):
* Enabling entities by default by using a generic "item" entity (`#1` in "Architectural Differences")
* Improvements to the authority framework (more flexibility by entity type and collection)  (related to `#3` in "Architectural Differences")
* Ability to use the authority in the UI to navigate to the linked entity
* Improvement to the submission UI
* Two DSpace-CRIS features
    * "Control when linked Items/Entities are created" (`#4` in "Feature Differnces")
    * "Multilingual support for content" (`#6` in "Feature Differences")

This work package may be completed in one or more PRs.

FrancescoMolinaro:
## References
* Fixes DSpace/DSpace#11748
* Requires backend PR: https://github.com/DSpace/DSpace/pull/11718
* Relates to: https://github.com/DSpace/dspace-angular/pull/4814

## Description

Initial port of the DSpace-CRIS authority framework into core DSpace, enabling real-time entity lookup, confidence-based linking, and automatic graph maintenance through server-side consumers. 

On the UI this PR introduces a new way to link authorithy controlled metadata to the related item, generating a link in search results or item page, this connection is handled in submission. (for configuration check the rest PR)

image

Through a rest configuration, is possible also to activate an "auto-fill" functionality, that will populate the submission fields based on the selected value (value must be authority controlled).

Exaple of config in cris-authority-metadatagenerator.xml:

    <bean class="org.dspace.content.authority.ItemSimpleAuthorityMetadataGenerator">
        <property name="authorityName" value="AuthorAuthority"/>
        <property name="keyId" value="person_identifier_orcid"/>
        <property name="useForDisplay" value="true"/>
        <property name="useAsData" value="true"/>
        <property name="schema" value="person"/>
        <property name="element" value="identifier"/>
        <property name="qualifier" value="orcid"/>
        <property name="singleResultOnAggregate" value="true"/>
    </bean>

[auto-fill.webm](https://github.com/user-attachments/assets/7a8b3ded-c57a-46b2-8baa-01fe3b5631cd)

The information of the related item are previewed via a new pop-up component enabled on hover, visible on both search results and item page:

image

image

In addition the PR introduce also multilanguage support for authority controlled vocabularies, to achieve this we need to have a vocabulary version for each language enabled by the property: webui.supported.locales

For instance if we have 

webui.supported.locales = en, de

we can create a translated vocabulary (srsc_de in this example) and then configure the authority as follows (authority.cfg):

vocabulary.plugin.srsc.hierarchy.store = true
vocabulary.plugin.srsc.hierarchy.suggest = true
vocabulary.plugin.srsc.delimiter = "::"
vocabulary.plugin.authority.store = true
authority.controlled.dc.subject = true

This will enable to display the translated value in the UI, if the language matches the translated one:

EN:

image

DE:

image

This last feature will be consolidated later in the merger with the cust layout functionality from CRIS (at the moment this is not presente in the exact same way on CRIS as there we rely on the layout feature that is not yet in scope).
The implementation will remain anyway as it will be useful also as fallback for not configured layouts.

## Instructions for Reviewers

Once the item relation is enstablished (submitted) the metadata will be linking to the related item, with a preview of metadata visible on link hover.

The new configuration for the pop-hover preview is available in default-app.config.ts, under the name:

**metadataLinkViewPopoverData**

It provides the opportunity to customize the preview metadata based on entity type.

In submission is possible also to configure specific icons for each source of authority, see config:

**sourceIcons** (under authority config)

In general for the newly added configuration you can refer to the config.example.yml file.

Once submitted, the authorithy controlled value should then result in a clickable link the search results (if the value is displayed) and in a clickable link in the item page (see images above).

Is possible to configure also auto-fill and generated metadata based on the authority (see descriptio or rest PR).

For the multilanguage support on vocabularies a translated vocabulary and submission form are needed, in addition to the properties based on the vocabulary name and metadata (see description).

##  List of changes in this PR:

Enhanced search results and item page with linkable related items (authority controlled)
Added new popover component for metadata preview on hover.
Implemented custom pipe for icon/info handling in onebox component.
Improved editing of authority via admin view, improved handling of search filters (filter by authorithy not value).
Implemented auto-fill functionality in submission.
Added support for multilanguage visualization of controlled vocabulary values in item page.

## Checklist


config/config.example.yml
```diff
@@ -629,3 +635,98 @@ geospatialMapViewer:
...
+
+# The maximum number of metadata values to process for each metadata key
+# when following authority metadata values.
+followAuthorityMetadataValuesLimit: 5;
+
+# When the search results are retrieved, for each item type the metadata with a valid authority value are inspected.
+# Referenced items will be fetched with a find all by id strategy to avoid individual rest requests to efficiently display the search results.
+followAuthorityMetadata:
+  - type: Publication
+    metadata: dc.contributor.author
```
artlowel: These are duplicated above

src/app/core/shared/item.model.ts
```diff
@@ -99,6 +99,13 @@ export class Item extends DSpaceObject implements ChildHALResource, HandleObject
   @autoserializeAs(Boolean, 'withdrawn')
   isWithdrawn: boolean;
 
+  /**
+   * A boolean representing if this Item is currently withdrawn or not
+   */
```
artlowel, line 104: This comment is for something else

config/config.example.yml
```diff
@@ -235,6 +235,12 @@ submission:
           style: text-muted
           icon: fa-circle-xmark
 
+    # Icons to be displayed next to an authority controlled value, to give indication of the source.
+    sourceIcons:
+      # Example of configuration for authority logo based on sources.
+      # The condigured icon will be displayed next to the authority value in submission and on item page or search results.
```
artlowel: Typo: "condigured"

src/app/core/data/external-source-data.service.ts
```diff
@@ -111,4 +122,18 @@ export class ExternalSourceDataService extends IdentifiableDataService<ExternalS
...
+   * @param entryId           The id of the entry to retrieve
+   */
+  getExternalSourceEntryById(externalSourceId: string, entryId: string): Observable<RemoteData<ExternalSourceEntry>> {
+    const href$ = this.getEntryIDHref(externalSourceId, entryId).pipe(
+      isNotEmptyOperator(),
+      distinctUntilChanged(),
+    );
+
+    return this.findByHref(href$) as any;
+  }
```
artlowel: This method doesn't seem to be used anywhere

src/app/core/data/item-data.service.ts
```diff
@@ -425,6 +430,57 @@ export abstract class BaseItemDataService extends IdentifiableDataService<Item>
...
+  public findById(id: string, useCachedVersionIfAvailable = true, reRequestOnStale = true, ...linksToFollow: FollowLinkConfig<Item>[]): Observable<RemoteData<Item>> {
+
+    if (uuidValidate(id)) {
+      const href$ = this.getIDHrefObs(encodeURIComponent(id), ...linksToFollow);
+      return this.findByHref(href$, useCachedVersionIfAvailable, reRequestOnStale, ...linksToFollow);
+    } else {
+      return this.findByCustomUrl(id, useCachedVersionIfAvailable, reRequestOnStale, linksToFollow);
+    }
+  }
+
```
artlowel, line 523: I think this check should be moved to the resolver. findById does one, well defined thing in all dataservices. The resolver should parse the route and determine whether to call findById or findByCustom URL, or even cleaner: move custom URLs to a different subroute, so you can have a separate resolver as well.

src/app/core/shared/metadata.models.ts
```diff
@@ -56,6 +58,24 @@ export class MetadataValue implements MetadataValueInterface {
...
+   * If this is a virtual Metadatum, it returns everything in the authority key after 'virtual::'.
+   * Returns undefined otherwise.
+   */
+  get virtualValue(): string {
+    if (this.isVirtual) {
+      return this.authority.substring(this.authority.indexOf(VIRTUAL_METADATA_PREFIX) + VIRTUAL_METADATA_PREFIX.length);
+    } else {
+      return undefined;
+    }
+  }
```
artlowel: These methods were moved out of the models, to a service in https://github.com/DSpace/dspace-angular/pull/2875. In general we want to avoid having methods in models, as they cause issues with caching serialization etc

src/app/shared/form/builder/ds-dynamic-form-ui/models/onebox/dynamic-onebox.component.ts
```diff
@@ -169,17 +182,15 @@ export class DsDynamicOneboxComponent extends DsDynamicVocabularyComponent imple
...
+    this.initVocabulary();
     this.isHierarchicalVocabulary$ = this.vocabulary$.pipe(
-      map((result: Vocabulary) => result.hierarchical),
+      filter((vocabulary: Vocabulary) => isNotEmpty(vocabulary)),
+      map((vocabulary: Vocabulary) => vocabulary.hierarchical),
+      tap((isHierarchical: boolean) => {
+        if (this.model.value) {
+          this.setCurrentValue(this.model.value, isHierarchical);
+        }
+      }),
```
artlowel: It's not a good idea to do "work" in a `tap()`. Because that code will only be executed if something is subscribed to the pipe. It's better to have a dedicated subscription to do the value updates, that way you can be sure it happens every time the source observable updates.

src/app/shared/metadata-link-view/metadata-link-view-orcid/metadata-link-view-orcid.component.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  orcidUrl$: Observable<string>;
+
+  constructor(protected configurationService: ConfigurationDataService) {}
+
+  ngOnInit(): void {
+    this.orcidUrl$ = this.configurationService
+      .findByPropertyName('orcid.domain-url')
+      .pipe(
+        getFirstSucceededRemoteDataPayload(),
```
artlowel: I suspect that, if `orcid.domain-url` isn't set on the server, there will never be a succeeded remote data, so this will never emit. Better deal with the failed case as well

artlowel (changes requested):
Thanks @FrancescoMolinaro 

I haven't tested it yet, just did a code review for now. Most of it looks good, but I have a few inline comments

FrancescoMolinaro:
Hi @artlowel ,

thank you very much for the review and for the feedback. I have addressed all the points raised, with the exception of the custom URL aspect.

The custom URL functionality originates from https://github.com/DSpace/dspace-angular/pull/4814. Since this pull request builds upon that work, I believe it would be more appropriate to address any related issue directly within that original PR, if needed.

During the implementation, we chose to place the ID format control within the `findById` method because route resolvers and guards primarily rely on this method to retrieve item information from UUID/ID parameters in the URL.

Following the referenced PR, when the relevant metadata is present, the custom URL replaces the standard UUID/ID parameter in the page URL and effectively becomes a unique identifier for the item.

For this reason, we considered it preferable to keep the validation logic inside the `findById` method, as this approach minimizes the impact across the application and avoids the need to modify multiple code paths where the UUID/ID is extracted from the URL.

I would really appreaciate your thoughts on this reasoning and of course if it is not enough to justify the choice made I will make sure to adapt the code accordingly in the referenced pull request.

config/config.example.yml
artlowel: The followAuthorityMetadata section is still duplicated

src/app/core/data/external-source-data.service.ts
```diff
@@ -66,6 +66,17 @@ export class ExternalSourceDataService extends IdentifiableDataService<ExternalS
...
+  /**
+   * Get the endpoint for an external source's entryValues by given entry id
+   * @param externalSourceId  The id of the external source to fetch entry for
+   * @param entryId           The id of the external source entry to retrieve
+   */
+  getEntryIDHref(externalSourceId: string, entryId: string): Observable<string> {
+    return this.getBrowseEndpoint().pipe(
+      map((href) => href + '/' + externalSourceId + '/entryValues/' + entryId),
+    );
+  }
```
artlowel: Now this is unused. I expect it was used by the getExternalSourceEntryById you removed earlier.

artlowel (changes requested):
Hi @FrancescoMolinaro

Apologies for the delayed response, I’ve been out sick.

It looks like #4814 has already been merged in the meantime.

>Following the referenced PR, when the relevant metadata is present, the custom URL replaces the standard UUID/ID parameter in the page URL and effectively becomes a unique identifier for the item.
>
>For this reason, we considered it preferable to keep the validation logic inside the `findById` method, as this approach minimizes the impact across the application and avoids the need to modify multiple code paths where the UUID/ID is extracted from the URL.

The reason for my original comment is that findById is used consistently throughout the application to retrieve resources by their UUID.

Handles are also unique identifiers for a resource, as are self links, and these are handled separately through dedicated methods and resolvers. While I understand that placing this logic inside findById may reduce the number of changes required, it introduces inconsistency across dataservices and increases the risk of edge cases.

In my view, this logic should be moved out of findById, either as part of this PR or in a separate, focused PR before the release of DSpace 10.

MarieVerdonck:
**Simple item page issues**

- If multiple authors linked this way after each other, they show up next to each other on item page instead of underneath each other
  - Both if no linked item could be created (for example if there is no collection with Person entity) => The authors on simple item page, which still have `will be generated::ORCID::{orcid}` authority but confidence -1
  - And if link does happen 
  
  
Screenshot 2026-02-20 at 12 39 47
Screenshot 2026-02-20 at 12 39 23
Screenshot 2026-02-20 at 12 38 51
Screenshot 2026-02-20 at 12 38 35

- Clicking on linked author => person page has search with `search.filters.applied.f.isAuthorOfPublication: {person-uuid}`
  - i18n message not resolved
  - This shows no results? Either indexing or filter name issue? => Expect at least Publication you just came from to show up, and all other publications this Person is linked to as author
  
Screenshot 2026-02-20 at 12 57 46

MarieVerdonck:
**Submission UI improvements**

- Once selected from dropdown in submission there is no visual indicator to distinguish plain text authors in the list from authority controlled authors added this way. So it's not clear to user whether the "authority" has been successfully select/applied, so it would be an improvement if an alternate way to render mdv in list with authority was applied.
Screenshot 2026-02-20 at 12 58 49

FrancescoMolinaro:
Hi @MarieVerdonck, thanks for the feedback, much appreciated.
Regarding the submission improvment to detect if the authority has been applied, I wanted to say that it will  be ported as part of https://github.com/DSpace/dspace-angular/pull/5097 , the UI can be configured to display some chips as follows:

image


Here we will have icons with colors matching the confidence of the authority.

For the label in the search component in item page I believe is unrelated to this PR as the issue is present also on the demo instance, e.g.
https://demo.dspace.org/entities/person/bcf6b178-dcfb-4fbe-b21b-6e0a936b531c .
The missing search results I believe are a rest issue, not sure what in specific, but if I use the demo or the sandbox for the rest connection I can properly see results in the search.


Instead for the positioning of the authors in the item page I have pushed a fix to have the list stacked vertically as it was before.

Additional note:

In the regular form the authority feedback is given by the circle icon in the input, that will have a different color based on the confidence:

image


It would be really appreciated if you could check this again, many thanks.

tdonohue (changes requested):
@FrancescoMolinaro : Thanks for this PR! I gave this some testing today (alongside the backend PR).  Mostly, it's working.  However, there are some display issues in the User interface that I'm seeing:

1. When submitting a Publication, if I search up an author's name, I can choose someone from *either* local database or ORCID sandbox.  However, the icons **always show a database icon**. In the screenshot below, these are all authors that only exist in the ORCID sandbox but there is no ORCID icon (like in your screenshot in the description of this PR).

Image

2. The display of authors in search results is now messed up.  Authors will **not** be displayed unless you click the "Show More" button.  But, after clicking "Show More" you see the author names vertically like this screenshot:

Image

3. If I create a Publication with three authors and then visit the publication page, the author hover-overs work well.  However, it's very difficult to click the "More info" link in the hover-over for the second or third author. As soon as you move the mouse up, the popup for the author above appears.  You have to carefully move the mouse in a diagonal to get to the link. This is a smaller thing, but it seems like a usability issue.  

Image

4. Finally, if you succeed in clicking "More Info" in the hover-over (see the third item just above), then that hover-over will "flash" briefly in the upper right corner of the page before loading the Person page for this Author.  This is also minor, but I'm consistently seeing a "flash" anytime I click this "More infor" link.

I'm still testing & haven't done a code review here yet.  But, that's what I've found so far.

config/config.example.yml
```diff
@@ -238,6 +238,12 @@ submission:
           style: text-muted
           icon: fa-circle-xmark
 
+    # Icons to be displayed next to an authority controlled value, to give indication of the source.
+    sourceIcons:
```
tdonohue: Small thing. I think this is intended to be embedded under `authority:` above, but it's currently not.  It needs to be indented more

config/config.example.yml
```diff
@@ -641,3 +647,86 @@ geospatialMapViewer:
 accessibility:
   # The duration in days after which the accessibility settings cookie expires
   cookieExpirationDuration: 7
+
+# Configuration for custom layout
+layout:
```
tdonohue, line 676: I'm not understanding what a custom "layout" is.  Will this make more sense in future PRs?  Is it specific to authority (as that seems to be the only thing under it right now)?

We should add more details to the comments here to describe how/where this section is used.

config/config.example.yml
```diff
@@ -641,3 +647,86 @@ geospatialMapViewer:
...
+    - entityType: PROJECT
+      entityStyle:
+        default:
+          icon: fas fa-project-diagram
+          style: text-success
+
+# When the search results are retrieved, for each item type the metadata with a valid authority value are inspected.
+# Referenced items will be fetched with a find all by id strategy to avoid individual rest requests
+# to efficiently display the search results.
+followAuthorityMetadata:
```
tdonohue: I'm not a fan of having three new top-level configs that all seem related:
* `followAuthorityMetadata`
* `followAuthorityMaxItemLimit`
* `followAuthorityMetadatValuesLimit`

Would it be possible to consider grouping these together?  Maybe even putting these **under** the `searchResults` section below like this..
```
searchResults:
    followAuthority:
        metadata: [ ]
        maxItemLimit: 100
        metadataValuesLimit: 5
```

MarieVerdonck:
**Regular author lookup issue**

When reviewing this PR, I’m encountering a console error during author lookup (magnifying glass icon next to author lookup through which previously existing local or import-from-orcid entity Person items were linked) in the submission flow: `NG0309: Directive NgbNavLinkBase matches multiple times on the same element`. As a result, nothing is displayed in the lookup popup. From initial investigation, this appears to be related to an ng-bootstrap dependency/version issue. However, upgrading ng-bootstrap to resolve this leads to different build errors, so the situation is not straightforward. 

Screenshot 2026-02-25 at 18 01 43
Screenshot 2026-02-25 at 18 01 54

Console message:

```
ERROR Error: NG0309: Directive NgbNavLinkBase matches multiple times on the same element. Directives can only match an element once.
    Angular 5
    DsDynamicLookupRelationModalComponent_Conditional_7_Template dynamic-lookup-relation-modal.component.html:13
    Angular 4
    DsDynamicLookupRelationModalComponent_Template dynamic-lookup-relation-modal.component.html:11
    Angular 21
    RxJS 6
    Angular 10
root_effect_scheduler.mjs:3637:23
```

Note I’m not *certain* this issue is introduced by this PR, but the problem does not occur on https://demo.dspace.org/. When building the current `main` branch locally, the sidebar does not render, and when trying to create a Publication item via `/mydspace` + button, the inputs in submission (Subscribe/Describe) do not render so can't check if issue already occurs there.

config/config.example.yml
```diff
@@ -641,3 +647,86 @@ geospatialMapViewer:
...
+    - dc.date.issued
+    - dc.type
+  # Metadata fields to be displayed in the search results for the author section
+  authorMetadata:
+    - dc.contributor.author
+    - dc.creator
+    - dc.contributor.*
+
+# Configuration of metadata to be displayed in the item metadata link view popover
+metadataLinkViewPopoverData:
```
tdonohue: Since this appears to be related to the Item page (primarily), could we group this under `item`?  For example:

```
item:
   metadataLinkViewPopoverData:
```

config/config.example.yml
```diff
@@ -641,3 +647,86 @@ geospatialMapViewer:
...
+      metadataList:
+        - person.affiliation.name
+        - person.email
+      # Title metadata (popover header)
+      titleMetadataList:
+          - person.givenName
+          - person.familyName
+# Configuration for identifier subtypes, based on metadata like dc.identifier.ror where ror is the subtype.
+# This is used to map the layout of the identifier in the popover and the icon displayed next to the metadata value.
+identifierSubtypes:
```
tdonohue: Since this is related to the popover, is this supposed to be nested under `metadataLinkViewProperites`? E.g.
```
item:
  metadataLinkViewProperites:
    identifierSubtypes:
```

In general, we've been trying to group related configurations together in the UI configuration as much as possible. So, all my suggestions above are trying to create better "groupings" of related settings

src/app/core/shared/item.model.ts
```diff
@@ -99,6 +99,13 @@ export class Item extends DSpaceObject implements ChildHALResource, HandleObject
   @autoserializeAs(Boolean, 'withdrawn')
   isWithdrawn: boolean;
 
+  /**
+   * A string representing the entity type of this Item
+   */
+  @autoserializeAs(String, 'entityType')
+    entityType: string;
```
tdonohue: Tiny thing. This line appears to be indented, but probably shouldn't be.

src/app/core/submission/vocabularies/models/vocabulary-entry-detail.model.ts
```diff
@@ -37,7 +43,21 @@ export class VocabularyEntryDetail extends VocabularyEntry {
...
+   */
+  @link(VOCABULARY_ENTRY_DETAIL)
+    parent?: Observable<RemoteData<VocabularyEntryDetail>>;
+
+  /**
+   * The submitter for this SubmissionObject
+   * Will be undefined unless the submitter {@link HALLink} has been resolved.
+   */
+  @link(VOCABULARY_ENTRY_DETAIL, true)
+    children?: Observable<RemoteData<PaginatedList<VocabularyEntryDetail>>>;
```
tdonohue: These two new properties (here and a few lines above) also appear indented & probably shouldn't be.

src/app/core/submission/vocabularies/models/vocabulary-entry.model.ts
```diff
@@ -44,6 +45,12 @@ export class VocabularyEntry extends ListableObject {
   @autoserialize
   otherInformation: OtherInformation;
 
+  /**
+   * A value representing security level value of the metadata
+   */
+  @autoserialize
+    securityLevel: number;
```
tdonohue: Again, property is wrongly indented

src/app/dso-shared/dso-edit-metadata/dso-edit-metadata-value/dso-edit-metadata-value.component.html
```diff
@@ -48,6 +48,15 @@
       <div class="dont-break-out preserve-line-breaks">{{ mdValue.newValue.language }}</div>
     }
   </div>
+  <div class="ds-flex-cell ds-authority-cell" role="cell">
+    @if (!mdValue.editing) {
+      <div class="dont-break-out preserve-line-breaks">{{ mdValue.newValue.authority }}</div>
+    }
+    @if(mdValue.editing) {
+      <textarea class="form-control" rows="5" [(ngModel)]="mdValue.newValue.authority"
```
tdonohue, line 56: This `<textarea>` should have a label or an `aria-label` for accessibility purposes.

src/app/shared/entity-icon/entity-icon.directive.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+   * Attach icon to HTML element
+   *
+   * @param crisStyle
+   * @private
+   */
+  private addIcon(entityStyle: AuthorityRefEntityStyleConfig): void {
+    const iconElement = `<i class="${entityStyle.icon} ${entityStyle.style}"></i>`;
```
tdonohue: We may want to add an `aria-hidden="true"` attribute to this `<i>` tag, similar to other icons in the UI.  We tend to hide these from accessibility tools unless they require "alt" text or similar.

src/app/shared/form/builder/ds-dynamic-form-ui/models/onebox/dynamic-onebox.component.html
```diff
@@ -7,60 +7,83 @@
...
+        @if (!searching && !loadingInitialValue) {
+          <i
+            dsAuthorityConfidenceState
+            class="far fa-circle fa-2x fa-fw my-auto p-0"
+            aria-hidden="true"
+            [authorityValue]="currentValue"
+            (whenClickOnConfidenceNotAccepted)="whenClickOnConfidenceNotAccepted($event)"></i>
+        }
+        @if(otherInfoValues.length > 0 && (model.id === otherInfoKey || otherInfoKey === alternativeNamesKey || 'data-' + model.id === otherInfoKey)) {
+          <i class="fa-solid fa-angle-down fa-fw fa-2x fa-fw position-absolute mt-1 p-0 additional-items-icon"
```
tdonohue: I think this `<i>` tag is also missing an `aria-hidden="true"`, unless I'm mistaken.

src/app/shared/form/builder/ds-dynamic-form-ui/models/onebox/dynamic-onebox.component.ts
```diff
@@ -323,19 +351,156 @@ export class DsDynamicOneboxComponent extends DsDynamicVocabularyComponent imple
...
+    }
+
+    return itemValue.substring(0, itemValue.lastIndexOf('::'));
+  }
+
+  toggleOtherInfoSelection() {
+    this.additionalInfoSelectIsOpen = !this.additionalInfoSelectIsOpen;
+  }
+
+  selectAlternativeInfo(info: string) {
```
tdonohue, line 407: Can we add TypeDocs to this new method?

src/app/shared/form/builder/ds-dynamic-form-ui/models/onebox/dynamic-onebox.component.ts
```diff
@@ -323,19 +351,156 @@ export class DsDynamicOneboxComponent extends DsDynamicVocabularyComponent imple
...
+      item: this.currentValue,
+    } as any;
+
+    this.onSelectItem(event);
+    this.searching = false;
+    this.toggleOtherInfoSelection();
+  }
+
+
+  setMultipleValuesForOtherInfo(keys: string[], item: any) {
```
tdonohue, line 450: Same here, it'd be good to add TypeDocs to describe the method.

src/app/shared/form/builder/ds-dynamic-form-ui/models/dynamic-vocabulary.component.ts
```diff
@@ -147,4 +220,92 @@ export abstract class DsDynamicVocabularyComponent extends DynamicFormControlCom
...
+    }
+  }
+
+  protected createChangeEventOnUpdate(models: DynamicFormControlModel[]) {
+    if (models.length > 0) {
+      this.onCustomEvent({ updatedModels: models }, 'authorityEnrichment');
+    }
+  }
+
+  getOtherInformationValue(value: string, key: string): FormFieldMetadataValueObject[] {
```
tdonohue, line 267: Same here, please add TypeDocs to describe this method.

src/app/shared/metadata-link-view/metadata-link-view-avatar-popover/metadata-link-view-avatar-popover.component.ts
```diff
@@ -0,0 +1,46 @@
...
+  styleUrls: ['./metadata-link-view-avatar-popover.component.scss'],
+  imports: [
+    AsyncPipe,
+    NgClass,
+    SafeUrlPipe,
+    ThemedLoadingComponent,
+    TranslateModule,
+  ],
+})
+export class MetadataLinkViewAvatarPopoverComponent extends ThumbnailComponent implements OnInit {
```
tdonohue, line 39: Please add a comment at the top of this component to describe how/where it is used.

src/app/shared/metadata-link-view/metadata-link-view-orcid/metadata-link-view-orcid.component.html
```diff
@@ -0,0 +1,23 @@
...
+  } @else {
+    {{ metadataValue }}
+  }
+
+  @if (hasBadge) {
+    <img
+      placement="top"
+      ngbTooltip="{{ 'orcid.badge.tooltip' | translate }}"
+      class="orcid-icon"
+      alt="orcid-logo"
```
tdonohue: Should this `alt` text be translatable?  It also should probably be "ORCID logo", as it should be English text.  (UPDATE: similar to comments below, if this is just an icon, we could replace this with `aria-hidden="true"`)

src/app/shared/metadata-link-view/metadata-link-view-orcid/metadata-link-view-orcid.component.html
```diff
@@ -0,0 +1,23 @@
...
+      {{ metadataValue }}
+    </a>
+  } @else {
+    {{ metadataValue }}
+  }
+
+  @if (hasBadge) {
+    <img
+      placement="top"
+      ngbTooltip="{{ 'orcid.badge.tooltip' | translate }}"
```
tdonohue, line 17: I don't believe this `orcid.badge.tooltip` key exists in the `en.json5` yet.

src/app/shared/metadata-link-view/metadata-link-view-orcid/metadata-link-view-orcid.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  selector: 'ds-metadata-link-view-orcid',
+  templateUrl: './metadata-link-view-orcid.component.html',
+  styleUrls: ['./metadata-link-view-orcid.component.scss'],
+  imports: [
+    AsyncPipe,
+    NgbTooltipModule,
+    TranslateModule,
+  ],
+})
+export class MetadataLinkViewOrcidComponent implements OnInit {
```
tdonohue, line 38: Please add a comment at the top of this component to describe where/how it is used.

src/app/shared/metadata-link-view/metadata-link-view-popover/metadata-link-view-popover.component.html
```diff
@@ -0,0 +1,77 @@
...
+                  <a [href]="getSourceSubTypeIdentifier(metadata)?.link + '/' + item.firstMetadataValue(metadata)" target="_blank" >
+                    {{ item.firstMetadataValue(metadata) }}
+                  </a>
+                }
+              </ng-container>
+              <img
+                placement="top"
+                [ngbTooltip]="getSourceSubTypeIdentifier(metadata)?.link"
+                class="source-icon"
+                alt="source-logo"
```
tdonohue: Same here, this `alt` text should likely be translatable and changed to "Source logo" (or "Logo of source").  Alternatively, if this is meant to just be an icon, we could remove the `alt` text and add `aria-hidden="true"` to hide it from accessibility tools

src/app/shared/metadata-link-view/metadata-link-view-popover/metadata-link-view-popover.component.ts
```diff
@@ -0,0 +1,142 @@
...
+    MetadataLinkViewAvatarPopoverComponent,
+    MetadataLinkViewOrcidComponent,
+    NgbTooltipModule,
+    NgOptimizedImage,
+    RouterLink,
+    TranslateModule,
+    VarDirective,
+  ],
+})
+export class MetadataLinkViewPopoverComponent implements OnInit {
```
tdonohue, line 44: Again, add a comment at the top of the component to describe where/how it is used.

src/app/shared/metadata-link-view/metadata-link-view.component.html
```diff
@@ -0,0 +1,51 @@
...
+        {{metadataView.value}}
+      </span>
+    </a>
+  </span>
+  @if (metadataView.orcidAuthenticated) {
+    <img
+      placement="top"
+      ngbTooltip="{{ metadataView.orcidAuthenticated }}"
+      class="orcid-icon"
+      alt="orcid-logo"
```
tdonohue: Same as above. This should either be translatable, or, if it's an icon, we could remove it and add `aria-hidden="true"`.  It looks like it might be an icon which is not necessary for assistive technologies to "read"

src/app/shared/metadata-link-view/metadata-link-view.component.ts
```diff
@@ -0,0 +1,182 @@
...
+    MetadataLinkViewPopoverComponent,
+    NgbPopoverModule,
+    NgbTooltipModule,
+    NgTemplateOutlet,
+    RouterLink,
+    StickyPopoverDirective,
+    VarDirective,
+  ],
+})
+export class MetadataLinkViewComponent implements OnInit {
```
tdonohue, line 77: Again, please describe this component with a comment.

src/app/shared/metadata-link-view/metadata-view.model.ts
```diff
@@ -0,0 +1,6 @@
+export interface MetadataView {
```
tdonohue, line 6: Can we describe this interface with a comment.  Describe where/how it is used.

src/app/shared/object-list/my-dspace-result-list-element/item-list-preview/item-list-preview.component.ts
```diff
@@ -90,6 +95,7 @@ export class ItemListPreviewComponent implements OnInit {
   ngOnInit(): void {
     this.showThumbnails = this.appConfig.browseBy.showThumbnails;
     this.dsoTitle = this.dsoNameService.getHitHighlights(this.object, this.item, true);
+    this.authorMetadataList = this.item.allMetadata(['dc.contributor.author', 'dc.creator', 'dc.contributor.*'], undefined, true);
```
tdonohue: Rather than hardcoding this list, should it use the new `searchResults > authorMetadata` configuration you've added in this PR?  Either way is fine, but just wanted to point out this is the same list as that `authorMetadata` config.

src/app/shared/image.utils.ts
```diff
@@ -0,0 +1,34 @@
+import {
+  Observable,
+  of,
+} from 'rxjs';
+import { map } from 'rxjs/operators';
+
+export const getDefaultImageUrlByEntityType = (entityType: string): Observable<string> => {
```
tdonohue, line 20: Please add TypeDocs to this exported utility method.

src/app/submission/edit/submission-edit.component.ts
```diff
@@ -147,7 +147,7 @@ export class SubmissionEditComponent implements OnDestroy, OnInit {
   ngOnInit() {
 
     this.collectionModifiable = this.route.snapshot.data?.collectionModifiable ?? null;
-
+    // TODO: add all languages projection
```
tdonohue: Is this still a necessary TODO?

src/config/additional-metadata.config.ts
```diff
@@ -0,0 +1,23 @@
...
+  | 'link'
+  | 'link.email'
+  | 'identifier'
+  | 'valuepair'
+  | 'date'
+  | 'authors'
+  | 'currentRole'
+  | 'lastRole';
+
+export interface AdditionalMetadataConfig extends Config {
```
tdonohue: Could we add a TypeDocs description of this interface & where/how it is used?

src/config/default-app-config.ts
```diff
@@ -252,6 +261,32 @@ export class DefaultAppConfig implements AppConfig {
           },
 
         ],
+        sourceIcons: [
+          {
+            source: 'orcid',
+            path: 'assets/images/orcid.logo.icon.svg',
```
tdonohue, line 262: NOTE: This ORCID icon is currently missing from this PR.

src/config/default-app-config.ts
```diff
@@ -646,4 +681,97 @@ export class DefaultAppConfig implements AppConfig {
   accessibility: AccessibilitySettingsConfig = {
     cookieExpirationDuration: 7,
   };
+
+  layout: LayoutConfig = {
```
tdonohue, line 707: Similar to all the other settings in this `default-app-config.ts`, each new top-level configuration should have a comment which describes it.  You've added a number of new top-level configurations to this file and very few have any description.  Please describe them all in comments.

src/config/layout-config.interfaces.ts
```diff
@@ -0,0 +1,19 @@
+import { Config } from './config.interface';
+
+export interface AuthorityRefEntityStyleConfig extends Config {
```
tdonohue, line 11: The new interfaces in this file should all have a comment to describe them

src/config/metadata-link-view-popoverdata-config.interface.ts
```diff
@@ -0,0 +1,27 @@
+export interface MetadataLinkViewPopoverDataConfig {
```
tdonohue, line 12: Same here, please add a comment to describe each interface in this file

tdonohue (changes requested):
@FrancescoMolinaro : Thanks again for this PR!  I've finished a code review and have some minor comments inline below.  Most of the comments are requests for TypeDocs or small accessibility improvements.  

However, I've also asked for some possible restructuring of the configuration (to group together related configs).  While I'd rather do this restructuring now, we could potentially delay it if necessary.  I just wanted to stress that we are trying to maintain some "order" / "hierarchy" in our config.yml, and this PR doesn't really obey those current best practices.

See inline comments below.  Thanks again!

config/config.example.yml
FrancescoMolinaro: Hi @tdonohue , thanks for the feedback, you are absolutely right it does not make much sense right now and as you said it will make more sense in next PRs as it is the front configuration of what is called CRIS layout, the customizable layout that will use an excel file.
The config relates to the different metadata and renderings (in future CRIS layout) used in item page, not only for the authority.

I will improve the description to make it more clear.

FrancescoMolinaro:
Hi @tdonohue,  many thanks for the review, much appreciated.
I have reorganized the configuration properties as you suggested and I realized that I have ported a configuration that was not in scope for this PR (AdditionalMetadataConfig) and that some of the added ones where not yet fully used, therefore I have removed what wasn't necessary and integrated what was missing.

Regarding the missing ORCID icon in submission, I think a config on the rest side was missing under authority.cfg:

cris.ItemAuthority.AuthorAuthority.source = orcid

Now it should be present also on the rest PR.
For the popover instead I made some improvments regarding the placement and the closing delay so that should be easier for the user to navigate them.

I have also fixed the search results layout so that the authors are visible as before and correctly placed:

image


In addition I have improved the closing method so to clear the HTML reference on navigation, so that the flash on the page corner won't happen.

FrancescoMolinaro:
Hi @MarieVerdonck , thanks for the feedback, I just checked the issue and is indeed caused by the update we did for ng-bootstrap: https://github.com/DSpace/dspace-angular/pull/4898

I have fixed it in this PR but I have also open a separated issue that I will address asap, as this issue is unrelated from this PR: https://github.com/DSpace/dspace-angular/issues/5169

tdonohue (changes requested):
@FrancescoMolinaro :  I gave this another test today (with the backend PR). The prior bugs that I found in [my first test from last week](https://github.com/DSpace/dspace-angular/pull/4956#pullrequestreview-3850219767) have _all been resolved!_  Thanks!

However, the code review feedback that I also gave last week (in [a separate review](https://github.com/DSpace/dspace-angular/pull/4956#pullrequestreview-3850551442)) is mostly **unresolved**. For example, I'm not seeing any of the accessibility improvements I suggested. (It looks like only around the first 5 requests were fixed, while the other 20+ have mostly not been solved yet.)

In my testing today, most everything is looking must better.  However, I've also found a new bug that is appearing in SSR logs.  Here's what I'm seeing:
1. Start the frontend in production mode (npm run build:prod && npm run serve:ssr)
2. Watch the SSR logs where you ran "npm run serve:ssr")
3. Visit the homepage, and you'll see an immediate error in the SSR logs which looks like this:
```
ERROR ReferenceError: document is not defined
    at new StickyPopoverDirective2 (C:\dspace-angular\dist\server\main.js:1:1673632)
    at NodeInjectorFactory.ɵfac [as factory] (C:\dspace-angular\dist\server\main.js:1:1674926)
    at getNodeInjectable (C:\dspace-angular\dist\server\main.js:1:3915934)
    at instantiateAllDirectives (C:\dspace-angular\dist\server\main.js:1:3979403)
    at createDirectivesInstances (C:\dspace-angular\dist\server\main.js:1:3979712)
    at Object.ɵɵelementStart (C:\dspace-angular\dist\server\main.js:1:4131435)
    at MetadataLinkViewComponent_ng_template_2_Template (C:\dspace-angular\dist\server\main.js:1:1676800)
    at executeTemplate (C:\dspace-angular\dist\server\main.js:1:3978418)
    at renderView (C:\dspace-angular\dist\server\main.js:1:3985607)
    at createAndRenderEmbeddedLView (C:\dspace-angular\dist\server\main.js:1:3986989)
```

This SSR error appears to be referencing the `MetadataLinkViewComponent` which is newly added in this PR.

If you can find time to address my other prior feedback, I'd love to give this an updated review.  As I'm sure you're aware, this is blocking several other merger PRs.  I'd really like to see this PR (and the backend) merged soon to unblock everything else.

src/app/submission/edit/submission-edit.component.ts
FrancescoMolinaro: Hi @tdonohue , thanks for the catch, I have added the projection in one of the PR based on this one: https://github.com/DSpace/dspace-angular/pull/5145.

I will remove the TODO as not necessary.

FrancescoMolinaro:
Hi @tdonohue , thanks for the feedback and sorry if missed some of the comments, now I should have addressed all the remaining open points and the new issue with the SSR error.

The issue was related to the window object being accessed in SSR inside StickyPopoverDirective, therefore I have adapted the logic so that it won't execute any rendering logic if not in the browser.

I believe this should be ready for another review, thanks in advance  for the help.

tdonohue (approved):
👍 Thank you @FrancescoMolinaro !  I retested & re-reviewed this today, and all the prior bugs that I've reported are fixed.  I can also verify my feedback has all been addressed in this PR.  (That said, there's still [feedback on the backend PR](https://github.com/DSpace/DSpace/pull/11718#pullrequestreview-3899954988) that needs addressing before I'm +1 on that as well)

This looks good to me now!  I'd encourage other testers/reviewers (like @MarieVerdonck , @artlowel and @nwoodward ) to provide updated feedback soon.  I'd really like to see this PR (and the backend) move forward as quickly as plausible in order to unblock all the other DSpace-CRIS merger PRs that depend on them.

src/app/submission/sections/form/section-form.component.ts
```diff
@@ -338,6 +340,8 @@ export class SubmissionSectionFormComponent extends SectionModelComponent {
       );
       const sectionMetadata = this.sectionService.computeSectionConfiguredMetadata(this.formConfig);
       this.sectionService.updateSectionData(this.submissionId, this.sectionData.id, sectionData, errorsToShow, serverValidationErrors, sectionMetadata);
+      // Add created model to formBulderService
+      this.formBuilderService.addFormModel(this.formId, this.formModel);
```
artlowel, line 344: It looks like FormModels are added using `this.formId` as the identifier…

src/app/submission/sections/form/section-form.component.ts
```diff
@@ -548,4 +597,11 @@ export class SubmissionSectionFormComponent extends SectionModelComponent {
       this.previousValue,
       null);
   }
+
+  ngOnDestroy(): void {
+    super.ngOnDestroy();
+    // Remove this model from formBulderService
+    this.formBuilderService.removeFormModel(this.sectionData.id);
```
artlowel: … and removed using `this.sectionData.id` as the identifier, which could be a memory leak.

I thought those IDs might be identical, but I tried logging them to the console, and I get something like this:
```
addFormModel 3_traditionalpagetwo
addFormModel 2_traditionalpageone
removeFormModel traditionalpageone
removeFormModel traditionalpagetwo
```

You'll probably want to use `this.formId` for the removal as well

artlowel (changes requested):
Thanks @FrancescoMolinaro!

I tested this again today and it looks good. I did find another potential issue in the code

FrancescoMolinaro:
Hi @artlowel , thanks for the feedback and the catch, you are absolutely right.
I have changed the removal identifier so that the form will be correctly removed from the state, avoiding possible memory leaks.

artlowel (approved):
Thanks @FrancescoMolinaro!

I'll give this a +1 now so it can move forward, assuming you'll still address [the findByID issue](https://github.com/DSpace/dspace-angular/pull/4956#pullrequestreview-3789909260) in a separate PR

tdonohue:
Merging as this is at +2.  Thanks again @FrancescoMolinaro !

@artlowel : Could you create a bug ticket for [that `findByID` issue](https://github.com/DSpace/dspace-angular/pull/4956#pullrequestreview-3945510470)?  That way we can ensure it is solved in the future.  This is the same approach we've taken with outstanding feedback on the backend (e.g. I created a blocker ticket for a backend issue that I feel must be fixed prior to 10.0, see https://github.com/DSpace/DSpace/issues/12119).  You can decide if you feel this is also a "blocker" for 10.0 or if this just needs to be cleaned up at some point (perhaps in 10.1 or similar)

tdonohue:
@FrancescoMolinaro : I forgot to mention, I'm adding the needs documentation label to the PR because we need to add documentation related this this PR (and it's backend) to our DSpace 10 documentation: https://wiki.lyrasis.org/display/DSDOC10x/

config/config.example.yml
```diff
@@ -393,6 +398,30 @@ item:
     pageSize: 5
     # Show the bitstream access status label on the item page
     showAccessStatuses: false
+  # Configuration of metadata to be displayed in the item metadata link view popover
+  metadataLinkViewPopoverData:
+    # Metdadata list to be displayed for entities without a specific configuration
+    fallbackMetdataList:
+      - dc.description.abstract
+      - dc.description.note
```
MMilosz, line 406: Where does `dc.description.note` come from? It's referenced only here but I don't see it defined in any `*-types.xml` file in the backend. Can it be safely deleted?

FrancescoMolinaro: Hi @MMilosz , you are right is a metadata that was present in CRIS and should have been removed, I have updated the example config as part of this PR https://github.com/DSpace/dspace-angular/pull/5145 that is in review, as this one is already merged.

tdonohue:
@FrancescoMolinaro : I see that the documentation for this feature has been started at https://wiki.lyrasis.org/display/DSDOC10x/Authority+framework+functionalities   However, I wanted to point out that those docs currently include **several broken images**.  If the broken images can be cleaned up, then we can remove the `needs documentation` label on this PR.

PR #4961, opened 2025-12-28
Fix for issue #4960 PM2 cluster mode ECONNREFUSED errors

### LINKED ISSUE #4960: PM2 fails in cluster mode for localhost because it defaults to IPv6 over IPv4
## Describe the bug
If you follow the DSpace installation instructions, that includes a PM2 snippet to run PM2 in cluster mode, it doesn't work.

## To Reproduce
1. Install DSpace 10 bleeding edge front & back end (including Angular 20)
2. Run angular with pm2 start ... with pm2 in cluster mode

## Expected behavior
It should work

## Observed behaviour
The pm2 logs have various flavours of no _links section found errors

## Cause

In a localhost setup, where tomcat is only listening on ipv4, Angular can't connect because it's trying over IPv6

## Related work
This appears since Angular 17: https://github.com/nodejs/node/issues/40537

bram-atmire:
## References
  * Fixes #4960

  ## Description
  Node.js 17+ changed DNS resolution to prefer IPv6 over IPv4. When running in PM2 cluster mode, this causes "ECONNREFUSED ::1:8080" errors if the backend only listens on IPv4 (127.0.0.1).

  This fix sets dns.setDefaultResultOrder('ipv4first') at the very start of the SSR bootstrap to ensure localhost resolves to IPv4.

  See: https://github.com/nodejs/node/issues/40537

  ## Instructions for Reviewers

  List of changes in this PR:
  * Added `dns.setDefaultResultOrder('ipv4first')` at the top of `src/main.server.ts` before any other imports
  * This ensures Node.js resolves "localhost" to IPv4 (127.0.0.1) instead of IPv6 (::1)

  **How to test:**
  1. Build the production app: `npm run build:prod`
  2. Start with PM2 in cluster mode: `pm2 start dspace-ui.json` (ensure `exec_mode: "cluster"` in config)
  3. Access http://localhost:4000/ - should return HTTP 200
  4. Without this fix, you would see "ECONNREFUSED ::1:8080" errors in the PM2 logs and HTTP 500 responses

  **Note:** This issue only occurs when:
  - Running Node.js 17+ (which prefers IPv6 for DNS resolution)
  - Using PM2 cluster mode (fork mode is unaffected)
  - The backend (Tomcat) only listens on IPv4

  ## Checklist

  - [ ] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC9x/Accessibility)** if it makes changes to the user interface. *(N/A - no UI changes)*
  - [ ] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations. *(N/A - no user-facing text)*
  - [ ] If my PR includes new libraries/dependencies (in `package.json`), I've made sure their licenses align with the [DSpace BSD License](https://github.com/DSpace/DSpace/blob/main/LICENSE) based on the [Licensing of Contributions](https://wiki.lyrasis.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-LicensingofContributions) documentation. *(N/A - uses built-in Node.js `dns` module)*

### DIFF
--- a/src/main.server.ts
+++ b/src/main.server.ts
@@ -1,9 +1,3 @@
-/**
- * Fix for Node.js 17+ where DNS resolution prefers IPv6 over IPv4.
- * This causes "ECONNREFUSED ::1:8080" errors in PM2 cluster mode when
- * the backend only listens on IPv4.
- * See: https://github.com/nodejs/node/issues/40537
- */
 import 'core-js/es/reflect';
 import 'zone.js';
 import 'reflect-metadata';
@@ -23,7 +17,10 @@ import {
 import { AppComponent } from './app/app.component';
 import { serverAppConfig } from './modules/app/server-app.config';
 
-// Apply DNS resolution order fix for Node.js 17+
+// Apply DNS resolution order fix for Node.js 17+ by preferring IPv4 over IPv6.
+// This fixes "ECONNREFUSED ::1:8080" errors in PM2 cluster mode when
+// the backend only listens on IPv4
+// See https://github.com/DSpace/dspace-angular/issues/4960
 setDefaultResultOrder('ipv4first');
 
 const bootstrap = (context: BootstrapContext) => bootstrapApplication(AppComponent, serverAppConfig, context);


jesielviana (commented):
Good fix!
Would it be better to make the DNS resolution order configurable through config.yml (e.g. something like `defaultDnsResultOrder: ipv4first`)?
That way, we could easily change the preference between IPv4 and IPv6 without needing to update the code.

bram-atmire:
@jesielviana Thanks for the suggestion! I'm curious: are there scenarios where you'd want to configure this differently? I'm trying to think through when ipv6first (or other values) would be beneficial.

  My (limited!!!) understanding is that most backend deployments use dual-stack (both IPv4 and IPv6), so ipv4first should work regardless of whether the backend supports IPv6. Even if the backend has IPv6 support, trying IPv4 first just means it falls back to IPv6 if IPv4 fails.

  The issue we're fixing is specifically that Node.js 17+ changed to prefer IPv6, which causes connection failures in PM2 cluster mode when the Node process tries to connect via ::1 instead of 127.0.0.1. This happens even when the backend supports both protocols.

  That said, I might be missing something! If there's a deployment scenario where configurable DNS resolution order would be helpful, I'd be happy to add that. Do you have a specific use case in mind?

jesielviana:
@bram-atmire  I don’t have much experience in this area either, but I was thinking that in a dual-stack setup, some users might want to prioritize IPv6 over IPv4, depending on their infrastructure or network preferences.

That said, I agree that ipv4first is a sensible default since it should work reliably in most environments.

src/main.server.ts
```diff
@@ -1,3 +1,9 @@
+/**
+ * Fix for Node.js 17+ where DNS resolution prefers IPv6 over IPv4.
+ * This causes "ECONNREFUSED ::1:8080" errors in PM2 cluster mode when
+ * the backend only listens on IPv4.
+ * See: https://github.com/nodejs/node/issues/40537
```
tdonohue: @bram-atmire : Can we remove this comment from the header of the file, as this comment doesn't apply to the entire file.  Instead, you can move some of these details down next to the new `setDefaultResultOlder()` call.. maybe say something like this:
```
// Apply DNS resolution order fix for Node.js 17+ by preferring IPv4 over IPv6.
// This fixes "ECONNREFUSED ::1:8080" errors in PM2 cluster mode the backend only listens on IPv4
// See https://github.com/DSpace/dspace-angular/issues/4960
setDefaultResultOrder('ipv4first');
```

bram-atmire: @tdonohue done!

PR #4999, opened 2026-01-14
fix: use built-in Angular email validator for eperson form

### LINKED ISSUE #4338: Not able to create an account with an email containing uppercase letters from eperson form
## Describe the bug
When I type in an email address containing uppercase letters on the EPerson input form (/access-control/epeople), I receive the info "This E-mail is not a valid email". The "Save" button is greyed out, I am not able to save the form.

## To Reproduce
1. Open the EPerson form under /access-control/epeople
2. Type in a valid email-address containing uppercase letters

## Expected behavior
When I use an email address containing uppercase letters on the EPerson input form, I am able to create the EPerson.

## Related work
- The [regex pattern](https://github.com/DSpace/dspace-angular/blob/96d2b660683d8df114e13237632bc51cece68a75/src/app/access-control/epeople-registry/eperson-form/eperson-form.component.ts#L304) behind this form does not allow uppercase letters in an email address, although uppercase letters would be no problem for email communication.
- It is likely a solution to use the regex pattern from the [register email form](https://github.com/DSpace/dspace-angular/blob/96d2b660683d8df114e13237632bc51cece68a75/src/app/register-email-form/register-email-form.component.ts#L141).
- There are [numerous](https://www.regular-expressions.info/email.html) [discussions](https://en.wikibooks.org/wiki/Coding_Cookbook/Validate_Email_Address) about email address validation and email regexes on the internet. I feel like we should not make the email input super restrictive in DSpace, since we anyway send emails to the users making sure the email address is valid through that route.

pzlakowski:
## References

Fixes [Not able to create an account with an email containing uppercase letters from eperson form #4338](https://github.com/DSpace/dspace-angular/issues/4338)

## Description

I have replaced regex pattern to use built-in one from Angular when we create e-person via form. 

I think we should refrain from using our own regex when Angular provides one similar. Unless there is a reason which I do not know about yet. 

e.g. issue shows to use pattern from `src/app/register-email-form/register-email-form.component.ts`

```regex
^[a-zA-Z0-9.!#$%&\'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$
```

where Angular tries to adhere to WHATWG spec too https://github.com/angular/angular/blob/4755bbd94964f27921ec203fc1a9233a01c84792/packages/forms/src/validators.ts#L151

Angular adds 3 different constraints compared to WHATWG:

* Disallow `local-part` to begin or end with a period (`.`).
* Disallow `local-part` length to exceed 64 characters.
* Disallow total address length to exceed 254 characters.

## Instructions for Reviewers

List of changes in this PR:
* I have replaced pattern to use built-in email validator provided by Angular
* I have updated test to include uppercased case

Steps test this PR:
1. Visit http://localhost:4000/access-control/epeople/create
2. Fill a form with uppercased e-mail e.g. `TEST@example.com`
3. Create account and should success

## Checklist

### DIFF
--- a/src/app/access-control/epeople-registry/eperson-form/eperson-form.component.spec.ts
+++ b/src/app/access-control/epeople-registry/eperson-form/eperson-form.component.spec.ts
@@ -328,15 +328,28 @@ describe('EPersonFormComponent', () => {
       });
     });
 
+    describe('with uppercased email specified', () => {
+      beforeEach(() => {
+        component.formGroup.controls.firstName.setValue('test');
+        component.formGroup.controls.lastName.setValue('test');
+        component.formGroup.controls.email.setValue('TEST@test.com');
+        fixture.detectChanges();
+      });
+
+      it('passes validation check', () => {
+        expect(component.formGroup.controls.email.valid).toBeTrue();
+        expect(component.formGroup.controls.email.errors).toBeNull();
+      });
+    });
 
     describe('after inserting email wrong should show pattern validation error', () => {
       beforeEach(() => {
-        component.formGroup.controls.email.setValue('test@test');
+        component.formGroup.controls.email.setValue('test');
         fixture.detectChanges();
       });
       it('email should not be valid because the email pattern', () => {
         expect(component.formGroup.controls.email.valid).toBeFalse();
-        expect(component.formGroup.controls.email.errors.pattern).toBeTruthy();
+        expect(component.formGroup.controls.email.errors.email).toBeTruthy();
       });
     });
 
--- a/src/app/access-control/epeople-registry/eperson-form/eperson-form.component.ts
+++ b/src/app/access-control/epeople-registry/eperson-form/eperson-form.component.ts
@@ -309,12 +309,12 @@ export class EPersonFormComponent implements OnInit, OnDestroy {
       name: 'email',
       validators: {
         required: null,
-        pattern: '^[a-z0-9._%+-]+@[a-z0-9.-]+\\.[a-z]{2,4}$',
+        email: null,
       },
       required: true,
       errorMessages: {
         emailTaken: 'error.validation.emailTaken',
-        pattern: 'error.validation.NotValidEmail',
+        email: 'error.validation.NotValidEmail',
       },
       hint: this.translateService.instant(`${this.messagePrefix}.emailHint`),
     });


alanorth:
Thanks @pzlakowski. I agree that we should not be maintaining our own validation logic for something common like emails when Angular provides one that is likely better tested.

Having said that, I tested this on DSpace 8 and noticed that the email `TEST@example` is valid with this PR. I'm surprised that Angular's validator would consider that valid. Is the behavior different in DSpace 9/main?

pzlakowski:
@alanorth this is desired behaviour https://angular.dev/api/forms/Validators#email given WHATWG spec. Angular recommends to use custom regex pattern if business needs are different. There should be no difference.

I think if we wanted to support to require top-level domain we could use regex like this:

```regex
^(?=.{1,254}$)(?=.{1,64}@)[-!#$%&'*+/0-9=?A-Z^_`a-z{|}~]+(\.[-!#$%&'*+/0-9=?A-Z^_`a-z{|}~]+)*@[A-Za-z0-9]([A-Za-z0-9-]{0,61}[A-Za-z0-9])?(\.[A-Za-z0-9]([A-Za-z0-9-]{0,61}[A-Za-z0-9])?)+$
```

This is derived from Angular regex, simply swapped from `*` to `+` at the end to enforce it.

alanorth:
Thanks @pzlakowski. I don't understand the upstream use case of allowing `test@example`, but perhaps I don't need to! If we can use the Angular email validator to reduce custom code we need to maintain, and in that process gain support for addresses like `TEST@example.com` then I am in support of this. Also, as this only affects the "create eperson" functionality that is only accessible by admins I think the negative impact is low. 

+1 by me by review and testing.

saschaszott:
Which email address validation is performed by the backend? Both validation rules (of frontend and backend) should be identical. Do we need to adapt the backend validation rule?

tdonohue (approved):
👍 Thanks @pzlakowski !  This looks reasonable to me, and I agree with @alanorth that'd I'd rather we adhere to rules established elsewhere (than to create our own custom regex).  I gave this some light testing and it appears to work.

tdonohue:
@pzlakowski : Just a note that this was only able to be auto-backported to 9.x and 8.x.  However, we only guarantee backports to 9.x anyways, so that's good enough.  That said, if you or anyone else wants to see this also fixed in 7.x then someone will need to create a manual backport PR against the `dspace-7_x` branch.

PR #5016, opened 2026-01-21
[Docker] Fix bug in `docker-deploy` test by always pulling backend image from docker.io

tdonohue:
## Description
Currently, our `docker-deploy` step is *failing* (on all active branches) whenever it runs *post-merger* of a PR.

Examples:
* https://github.com/DSpace/dspace-angular/actions/runs/21182082205/job/60933641993
* https://github.com/DSpace/dspace-angular/actions/runs/21184451358/job/60936020654

The reason for this failure is that we're attempting to pull the *backend* image from ghcr.io (in order to start a backend to test against).  But, we run into a permissions error every time as our ghcr.io images are private (and only used for build verifications).

This small PR just updates our GitHub actions `docker.yml` to ensure the backend image is always pulled from `docker.io` (our public images).   This fixes a minor flaw in the new `docker-deploy` tests added in #4987

(NOTE: The frontend images continue to be taken from `ghcr.io` in order to test the latest built frontend image.  This remains unchanged because the `dspace-angular` repo can easily pull frontend images from `ghcr.io`.  It just doesn't have permissions to pull the **backend images**)

## Instructions for Reviewers
* Unfortunately, this cannot be manually tested as it is only updating our GitHub actions `docker.yml` script.   Therefore, assuming our automated tests all pass, I'll merge this immediately and port to all active branches (as this same flaw appears on all branches)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -74,10 +74,6 @@ jobs:
       DSPACE_REST_HOST: 127.0.0.1
       # Override default dspace.ui.url to also use 127.0.0.1.
       dspace__P__ui__P__url: http://127.0.0.1:4000
-      # Docker Registry to use for Docker compose scripts below.
-      # If this is a PR, then we need to use docker.io (as the registry must be public),
-      # Otherwise we default to ghcr.io to avoid aggressive rate limits at DockerHub.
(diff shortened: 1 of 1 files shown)


tdonohue:
👍 Merging immediately as all automated tests pass & this fixes a build bug on all maintenance branches.

tdonohue:
Verified this fixed the bug in `docker-deploy` post-merger of this PR.  The tests are succeeding again.

PR #5045, opened 2026-01-27
Limit configuration available publicly in config.json

### LINKED ISSUE #5030: Limit configuration available publicly in `config.json`
## Describe the bug

As [reported on dspace-tech](https://groups.google.com/g/dspace-tech/c/uuaEONDXytM), the `config.json` public availability could be flagged as an "information disclosure" vulnerability by security scanners.  This was also discussed in [today's Developers Meeting](https://wiki.lyrasis.org/display/DSPACE/2026-01-22+DSpace+Developers+Meeting).

This `config.json` is used by the DSpace UI for runtime configuration of the user interface.  The purpose of this configuration is to tell the DSpace UI which features are enabled on specific pages. The `config.json` should not include secure or private information.

The `config.json` is initially loaded via SSR (Server Side Rendering) which does not require it to be publicly available.  However, the `config.json` **must be publicly available** in CSR (Client Side Rendering) mode, unless the application state has been transferred from SSR mode (see [code in `main.browser.ts`](https://github.com/DSpace/dspace-angular/blob/main/src/main.browser.ts#L31-L44)). 

This means that, if you block access to the `config.json`, it's possible that CSR mode may not function properly in all situations.  This is especially true if your initial access to the site is a page that doesn't undergo SSR (e.g. search pages or similar).  (_NOTE: These statements require further testing._)

Overall, this means that we should **limit the configurations available in the `config.json`** to only those configurations which are required by CSR or pose no "information disclosure" threat.

Possible configurations to move out of the `config.json` include:
* Any configuration that is specific to SSR (e.g. `rest.ssrBaseUrl` and the `serverSide` cache settings)\
* Potentially configurations that are better off moved to the backend.

Settings that may need to remain in the `config.json` include:
* `ui` settings - These tell Node.js where to start the UI.  However, they should always reference blocked/private ports in production scenarios because the UI is run behind a proxy in production mode.
* Other `rest` settings - These tell the UI where the public REST API is. So they always point to a public, findable server.
* Other settings used by the UI to change behavior (includes things like configuring results per page, languages, etc.). All of these are discoverable from the UI anyways.

## To Reproduce
1. See that the `config.json` is publicly available on the demo site: https://demo.dspace.org/assets/config.json

## Expected behavior
The `config.json` likely needs to remain public.  But, we can do a better job of limiting the configs available in that file to only those configs *required by CSR mode*.   Any SSR specific configs should be moved to a private configuration file which is only available on the server itself.

Aryakoste:
## References

* Fixes #5030 

## Description
Successfully split the configuration. config.server.ts now filters sensitive keys (rest.ssrBaseUrl, cache.serverSide, ui.rateLimiter, ui.useProxies) before writing to assets/config.json. The server process retains the full configuration via the return value of buildAppConfig

## Checklist

### DIFF
--- a/src/config/config.server.ts
+++ b/src/config/config.server.ts
@@ -169,6 +169,13 @@ const buildBaseUrl = (config: ServerConfig): void => {
   ].join('');
 };
 
+/**
+ * Removes all server-side specific settings from the application configuration.
+ * This method is used to ensure the "assets/config.json" that provides runtime
+ * configuration to CSR (client side rendering) excludes these server-side keys.
+ *
+ * @param config  the application configuration
+ */
 const removeServerSideConfig = (config: AppConfig): any => {
   const clientConfig = JSON.parse(JSON.stringify(config));
   delete clientConfig.rest.ssrBaseUrl;


src/config/config.server.ts
```diff
@@ -169,6 +169,16 @@ const buildBaseUrl = (config: ServerConfig): void => {
   ].join('');
 };
 
+const removeServerSideConfig = (config: AppConfig): any => {
```
tdonohue, line 179: @Aryakoste : Please add an inline comment above this to describe why we're doing this.  It could be something like this:
```
/**
 * Removes all server-side specific settings from the application configuration.
 * This method is used to ensure the "assets/config.json" that provides runtime
 * configuration to CSR (client side rendering) excludes these server-side keys.
 *
 * @param config  the application configuration
 */
```

src/config/config.server.ts
```diff
@@ -169,6 +169,16 @@ const buildBaseUrl = (config: ServerConfig): void => {
...
+  const clientConfig = JSON.parse(JSON.stringify(config));
+  delete clientConfig.rest.ssrBaseUrl;
+  delete clientConfig.rest.hasSsrBaseUrl;
+  delete clientConfig.cache.serverSide;
+  delete clientConfig.ui.rateLimiter;
+  delete clientConfig.ui.useProxies;
+  return clientConfig;
+};
```
strawburster, line 187: Could this be inverted to only return client config properties (instead of deleting server properties)? In the case that a config property is added in the future without someone knowing about this vulnerability, it should default to not being in the client config.

tdonohue:
Closing, as this was replaced by #5107.  Apologies we were not able to accept this PR.  However, thank you @Aryakoste for your contribution in helping us to discuss the best direction to fix this bug.

PR #5099, opened 2026-02-10
Fix end-user agreement checkbox reset and navigation hang

JorgeEscire:
### **References**
Fixes #5085

### **Description**
This PR fixes two critical bugs in the end-user agreement flow:

Part 1 - Checkbox reset preventing form submission: When new users accessed the end-user agreement page, checking the accept checkbox would sometimes get reset automatically, making the Save button appear non-functional until page reload.

Part 2 - Navigation hanging after successful submission: After successfully accepting the agreement (PATCH request succeeded), the page would stall indefinitely without redirecting to the intended destination.

**Root Causes**
Both issues were caused by improper handling of ngrx store observables that never complete:

1. Checkbox reset: The initAccepted() subscription to hasCurrentUserOrCookieAcceptedAgreement() lacked take(1), allowing store re-emissions to override user interactions by resetting this.accepted to false.

2. Navigation hang: After the PATCH operation, the endUserAgreementCurrentUserGuard would check user acceptance status using store selectors that could return stale data or never complete, causing the Angular router to hang. 

**Solution**
Applied established DSpace patterns for handling store-backed observables:

Added take(1) to all store selector subscriptions to ensure they complete after first emission
Implemented cookie fallback after successful PATCH operations for synchronous state verification
Updated the guard to use hasCurrentUserOrCookieAcceptedAgreement() which checks the cookie first before querying potentially stale store data
Instructions for Reviewers
This PR modifies the end-user agreement component, service, and guard to properly handle RxJS observables from the ngrx store.

**List of changes in this PR:**

1. end-user-agreement.component.ts: Added take(1) to initAccepted() subscription to prevent store re-emissions from overriding user checkbox interactions
2. end-user-agreement.service.ts:
    - Added take(1) to hasCurrentUserAcceptedAgreement() when calling getAuthenticatedUserFromStore()
    - Modified setUserAcceptedAgreement() to set cookie immediately after successful PATCH as a synchronous fallback

3. end-user-agreement-current-user.guard.ts: Updated to use hasCurrentUserOrCookieAcceptedAgreement() with take(1), leveraging cookie-based checks to prevent guard hangs

**How to test / review:**
Setup:

1. Ensure info.enableEndUserAgreement is set to true in your config
2. Clear browser cookies (especially hasAgreedEndUser)
3. Create a new test user account OR remove the dspace.agreements.end-user metadata from an existing user

**Test Part 1 - Checkbox functionality:**

1. Log in as a user without end-user agreement metadata
[...]

PR #5106, opened 2026-02-12
[DSpace-CRIS] Port CMS edit and User Agreement edit from DSpace CRIS

### LINKED ISSUE #11751: [DSpace-CRIS] Edit Content Management System (CMS) features
> [!NOTE]
>  This ticket represents a **feature** to be ported from DSpace-CRIS as part of the [merger of DSpace-CRIS into DSpace](https://dspace.org/announcement-dspace-dspace-cris-merger-approved-by-dspace-governance/).

# Description

The DSpace-CRIS features "Edit Homepage News, Header and Footer (Edit CMS Metadata)" and "Edit User Agreement" should be ported to DSpace.

Below is their description from the analysis of [DSpace-CRIS and DSpace Differences](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences):

> 18. Edit Homepage News, Header and Footer (“[Edit CMS Metadata](https://wiki.lyrasis.org/x/1IG3Fg#HowisitdifferentthanDSpace-EdituseragreementviaUI)”): From the admin sidebar menu, administrators can choose to edit the Homepage header, footer or news section, and translate it into other major languages.
>     1. Stores metadata on the "Site" object in DSpace.  This metadata is used to store the Header, Footer, and Homepage News (and the User Agreement below).  Multilingual support in the metadata allows you to save the News in different languages.
>          1. Content can include fragments of HTML or Markdown or similar.  Not every HTML tag is accepted – HTML is sanitized.
>     1. Discussed during DSpace 7 development, this topic came up as this feature used to exist in the DSpace 6 JSPUI.  But, at that time, there was disagreement around where to store this CMS-like data.  The DSpace-CRIS approach is to store it as metadata on the Site object.  During DSpace 7 early development, an alternative approach was suggested to store it in one or more Bitstreams (see https://github.com/DSpace/RestContract/pull/45).  Agreement was never reached, and the feature fell off the priority list.
>     1. This concept continues to be of interest to DSpace users, see https://github.com/DSpace/dspace-angular/issues/3200
>         1. Several related Pull Requests also use a similar idea of storing layout / CMS-like content in metadata fields, e.g. https://github.com/DSpace/dspace-angular/pull/3250, https://github.com/DSpace/dspace-angular/pull/3256
>
> 19. [Edit User Agreement](https://wiki.lyrasis.org/x/1IG3Fg#HowisitdifferentthanDSpace-EdituseragreementviaUI): From the admin sidebar menu, administrators can edit the “User Agreement” text (using Markdown), and translate it into other major languages. 
>     1. This uses the same infrastructure as the prior feature (Edit Homepage News, Header and Footer). It's managed by using metadata on the Site object.
>     1. If user agreement is modified after agreed on, is there are record which agreement was signed or agreed to? Similar to regular DSpace, with difference to be able to edit agreement from UI; this refers to the first login agreement, not the deposit license
>         1. This behavior is enforced at API level

This port task may be claimed and completed by anyone, but must involve porting the existing DSpace-CRIS feature code into DSpace. This ported code may be enhanced or updated as necessary. Volunteers must provide an estimate for when they will likely complete this port.

Dawnkai:
## References
* Resolves DSpace/DSpace#11751

## Description
Port CMS metadata editing and user agreement editing as part of CRIS merger.

## Instructions for Reviewers
This PR ports two functionalities: CMS metadata editing, and User Agreement editing from DSpace CRIS. I prioritized moving code from CRIS to DSpace as is, without drastic changes in logic and functionality. However, I can see a few places for improvement.

List of changes in this PR:
* A new component called `AdminEditCmsMetadata` has been added. This view is responsible for editing metadata on the Site object, which will then be displayed in selected parts of DSpace.
* A new component called `AdminEditUserAgreement` has been added. This view is responsible for editing the user agreement metadata, which is displayed instead of the usual User Agreement text. 
* Four components have been edited to support new metadata and will display it if it has been set through CMS edit or User Agreement edit: `FooterComponent`, `HomeNewsComponent`, `HomePageComponent` and `EndUserAgreementContentComponent`.
* New translation keys have been added to translation files along with the default English translation.
* New entries have been added in the Admin Panel sidebar - one for CMS metadata edit and the other for User Agreement content edit.
* When User Agreement is edited, administrators will have an option to reset the user agreement status using the `metadata-deletion` process on the backend. This means that all users will be forced to accept UA again.

In order to test this PR, you have to first import the metadata registry entries from DSpace/DSpace#11896.

### CMS metadata edit

Then, when you log in as an admin, you should see two new buttons on the admin panel:

image

Then, you should be able to select three metadata values: `cris.cms.home-header`, `cris.cms.home-news` and `cris.cms.footer`. Select the value you wish to edit. I recommend selecting `cris.cms.home-header`, as that one will be visible right away. Then, type something in the textarea for the langauge of your choice:

image

And then click `Save`. After returning to the home page, you should see that text instead of the regular DSpace banner:
image

In order to test the other metadata, you will have to not only repeat the steps for those metadata, but also make some changes in the code:
- `cris.cms.footer` requires you to edit `showTopFooter` in `FooterComponent` to `true`.
- `cris.cms.home-news` requires you to remove the `cris.cms.home-header` metadata and then set `cris.cms.home-news` metadata. Or you can simply temporarily pull the `ds-home-news` component out of the `@if` statement in the `home-page.component.html` file.

### User Agreement edit

In order to test User Agreement editing, you have to select the `Edit User Agreement` option from the admin panel. You should see a page that will display all available languages and textareas for them. Find the language you have currently selected. For example, I set it for English:

image

Then, you want to scroll down and press `Save`. You will have two options available:
image

There are two options here:
- `No, update only`, which will save the metadata and do nothing else. Then you will need to create a new account and check if the user license content that is displayed to new users has changed.
- `Yes, update and force`, which will save the metadata **and** force all users to accept it again. You can use this and then you have to refresh the page to see the effect.

In the end, you should see this:
image

## Checklist

- [X] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [X] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [X] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [X] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC9x/Accessibility)** if it makes changes to the user interface.
- [X] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
- [X] My PR **includes details on how to test it**. I've provided clear instructions to reviewers on how to successfully test this fix or feature.
- [X] If my PR includes new libraries/dependencies (in `package.json`), I've made sure their licenses align with the [DSpace BSD License](https://github.com/DSpace/DSpace/blob/main/LICENSE) based on the [Licensing of Contributions](https://wiki.lyrasis.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-LicensingofContributions) documentation.
- [X] If my PR fixes an issue ticket, I've [linked them together](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).


Dawnkai:
When creating this PR, I decided to limit the amount of edits that I made to the code in CRIS and tried to port it as close to 1:1 as possible. This was caused by the time restriction (and busy work week). I have a few potential ideas on how to improve this solution in mind, and I will add them in the code in the form of comments for further potential discussion.

tdonohue:
@Dawnkai : Thanks for creating this PR! Your [approach sounds correct](https://github.com/DSpace/dspace-angular/pull/5106#issuecomment-3891477545). I'd recommend keeping this PR as close to 1:1 port of code as possible.  

That said, you are welcome to add suggestions for changes here for discussion. Or, you also could create a new ticket to describe the improvements (linking them back to this PR).  Assuming we all can agree on the improvements, those could always be made in a future, follow-up PR.

(As a sidenote, it looks like some automated tests are failing in this PR. However, the error looks to me like it could be a random failure, so I've triggered a re-run of the tests)

tdonohue:
@Dawnkai : After rerunning the tests, they are now returning a different error.  It looks like this new error is saying that a few unit tests / specs are broken. Could you look into that and get them fixed?

Dawnkai:
@tdonohue I fixed it. The problem was that in the old version in the tests for the footer component an empty store was mocked, which was a problem - because in CRIS code the component is making a request (and thus fetching non-existing request cache from it). I mocked the SiteService (which is where the request was made) instead. Now it works.

Now all the tests pass.
When it comes to the improvements, since you want this to be merged asap, I will make a separate Issue once this is merged as a field for discussion with my recommendations.

Considering you want 1:1 port (which is of course not fully possible), here is a list of changes I made compared to a "pure" CRIS port:
- I removed all the `*ng` commands from templates (`*ngIf`, `*ngFor`, etc) and replaced them with newer angular format (`@if`, `@for`, etc)
- Changed import locations in tests, because some files are in different place in clean DSpace compared to CRIS
- Changed names of the CMS edit and User Agreement edit Components to have `Admin` prefix (so `EditCmsMetadataComponent` turns into `AdminEditCmsMetedataComponent`) in order to match the standard of other components on admin panel
- Added typedocs and docstrings in functions that didn't have them in CRIS (there were a few of them)
- _Slightly_ refactored footer component tests to be more readable and to fix the error I mentioned at the beginning of this comment
- Because `getCurrentLanguageCode` is an Observable in DSpace and in CRIS it is not, I had to use `forkJoin` or `combineLatest` to pull metadata from Site object and current language for displaying CMS metadata and User Agreement
- CRIS is using special layout objects to manage components display on the website (to integrate with CRIS layout tool). These layout object elements are not defined in DSpace and I don't think it's worth moving the whole thing just to make a port, so I simply took what was used (in this case, just the metadata key) from those objects and used it directly in code.

So in terms of functionality, no changes were introduced.

tdonohue:
Thanks @Dawnkai !  Just to clarify, I realize a 1:1 port will never be fully possible when merging these DSpace-CRIS features back to DSpace.  

The Merger Management Team has just made a decision that none of these ports can change the feature's behavior in any significant way.  But, we do realize that some code may require updates to be compatible with DSpace... and those small changes for compatibility purposes are welcome.

Thank you for describing the changes you had to make.  At a glance, those all look like changes for "compatibility purposes" and would be allowed.

src/config/default-app-config.ts
```diff
@@ -646,4 +647,12 @@ export class DefaultAppConfig implements AppConfig {
   accessibility: AccessibilitySettingsConfig = {
     cookieExpirationDuration: 7,
   };
+
+  cms: CmsMetadata = {
+    metadataList: [
+      'cris.cms.home-header',
+      'cris.cms.home-news',
+      'cris.cms.footer',
```
vins01-4science: Change these metadata accordingly ( using the `dspace` schema )

src/environments/environment.test.ts
```diff
@@ -477,4 +477,12 @@ export const environment: BuildConfig = {
   accessibility: {
     cookieExpirationDuration: 7,
   },
+
+  cms: {
+    metadataList: [
+      'cris.cms.home-header',
+      'cris.cms.home-news',
+      'cris.cms.footer',
```
vins01-4science: Change these metadata accordingly ( using the `dspace` schema )

src/app/footer/footer.component.ts
```diff
@@ -62,6 +73,16 @@ export class FooterComponent implements OnInit {
...
     this.showSendFeedback$ = this.authorizationService.isAuthorized(FeatureID.CanSendFeedback);
+
+    this.footerMetadataValue$ = combineLatest({
+      site$: this.siteService.find().pipe(
+        take(1),
+      ),
+      language$: this.locale.getCurrentLanguageCode(),
+    }).pipe(
+      take(1),
+      map(({ site$, language$ }) => site$?.firstMetadataValue('cris.cms.footer', { language: language$ })),
```
vins01-4science: Change these metadata accordingly ( using the `dspace` schema )

src/app/home-page/home-news/home-news.component.ts
```diff
@@ -1,13 +1,51 @@
...
+
+  ngOnInit() {
+    this.homeNewsMetadataValue$ = combineLatest({
+      site$: this.route.data.pipe(
+        map((data) => data.site as Site),
+      ),
+      language$: this.locale.getCurrentLanguageCode(),
+    }).pipe(
+      take(1),
+      map(({ site$, language$ }) => site$?.firstMetadataValue('cris.cms.home-news', { language: language$ })),
```
vins01-4science: Change these metadata accordingly ( using the `dspace` schema )

src/app/home-page/home-page.component.ts
```diff
@@ -56,6 +71,14 @@ export class HomePageComponent implements OnInit {
     this.site$ = this.route.data.pipe(
       map((data) => data.site as Site),
     );
+
+    this.homeHeaderMetadataValue$ = combineLatest({
+      site: this.site$,
+      language: this.locale.getCurrentLanguageCode(),
+    }).pipe(
+      take(1),
+      map(({ site, language }) => site?.firstMetadataValue('cris.cms.home-header', { language })),
```
vins01-4science: Change these metadata accordingly ( using the `dspace` schema )

EikLoe:
@Dawnkai Thanks for the PR. I tested it locally and it worked as inspected. The only issue I had was the display of the cris.cms.home-header: 

Screenshot_Example_Home-news


As someone who don't really know any DSpace-CRIS-System, I wondered, why the position of the header is outside of our normal _container_-borders, but as this is the described behavior, I'll agree withe PR in general 👍🏾

vins01-4science (changes requested):
Hello @Dawnkai, thank you for supporting us with this PR.

The development works as expected, I've only found only few problems:

1. The user agreement says that can be fullfilled in MD style, but in reality it can be fullfilled only with HTML syntax: 
Image
I think it's okay, but we may need to explicitly say that.

2. The Save button on end-user-agreement page ( after the login ) is not working anymore. If I click it won't load anything at all, it just shows a loading spinner endlessly ( the patch is sent to the BE, so if I manually change the route it loads the page I requested correctly ). [redirect-url](http://localhost:4000/info/end-user-agreement?redirect=%252Fhome)

Image

3. I've noticed that also the CMS metadata rendering are not able to display the MD, that it's okay, but I think it would be better to specify that as of now only HTML would be rendered correctly.
4. I think we need to change metadata definition, at least for these ones that are new metadata and easily replaceable by using the `dspace` schema.

---

cc: @tdonohue @atarix83

Dawnkai:
Update from me about the progress and response to the comments so far:

@EikLoe Yes, it also irks me, but just as instructed - I tried to limit the altering of CRIS code to the absolute minimum. CRIS has a different format of displaying that section of the page because 4Science is using dynamic components to specify what is to be displayed and in regular DSpace it's hard coded into the HTML. I can apply the container styles to that section of the home page, but only if I am allowed to make such change by @tdonohue.

 @vins01-4science Thank you for your feedback. To address your points:

> The user agreement says that can be fullfilled in MD style, but in reality it can be fullfilled only with HTML syntax

Are you sure? Remember that you need to set the environment variable called `markdown.enabled` to `true` in your `config.yml`. I set it to true and markdown works on my end:

image

> The Save button on end-user-agreement page ( after the login ) is not working anymore. If I click it won't load anything at all, it just shows a loading spinner endlessly

Yes, that is true. I am trying to fix it now. It seems to be related to caching and subscriptions in the `endUserAgreementGuard`. It has a function called `returnEndUserAgreementUrlTreeOnFalse` which detects if a user has accepted the end user agreement. For some reason the request that is being sent in that function to the store has state set to `Pending` and it never finishes. Initially I thought it was because the request was cached, so I disabled caching, but that broke even more things, so I have to dig deeper.

> I think we need to change metadata definition, at least for these ones that are new metadata and easily replaceable by using the dspace schema.

I edited the metadata used for CMS on both frontend and backend PR to use `dspace` schema instead of `cris` schema.

tdonohue:
@Dawnkai :  I've not had time to look at this PR myself, so I'm unclear on the exact details of the display issues that you and @EikLoe are referencing (and the screenshot @EikLoe [previously shared](https://github.com/DSpace/dspace-angular/pull/5106#issuecomment-3971013255) isn't loading for me).

However, to clarify, you are welcome to make any changes that help this feature to **work better within DSpace**.  So, if style changes are necessary, please make them.  The only changes that the Merger Management Team will *not* accept are changing the **feature behavior**... the feature's behavior must remain "as-is". In other words, we want to avoid redesign of the feature itself during migration from DSpace-CRIS to DSpace.

Hopefully that clarifies things. But, if I'm misunderstanding the question, then please let me know.

EikLoe:
> (and the screenshot @EikLoe https://github.com/DSpace/dspace-angular/pull/5106#issuecomment-3971013255 isn't loading for me).

@tdonohue Oh, sorry. I updated my comment. It should be displayed correctly now.

tdonohue:
@Dawnkai:  Based on @EikLoe 's [updated screenshot](https://github.com/DSpace/dspace-angular/pull/5106#issuecomment-3971013255), I think that display is also incorrect.  Obviously, the header shouldn't appear in a sidebar, so that's something that should be fixed within this PR.  So, please do update this PR to get the header to display in the proper location for DSpace.

tdonohue:
@Dawnkai : Could we get this PR and the backend PR updated / cleaned up (both currently have merge conflicts)?  Our 10.0 feature deadline is Friday, and I'd like to see if we can move this forward.  Thanks!

Dawnkai:
@tdonohue I am updating both frontend and backend now.
I managed to fix the styling issue mentioned by EikLoe. Now the margins are applied properly:
image

As for the issue mentioned by Vincenzo, I don't have good news.
It seems that there is some issue with requests deep within DSpace (without CRIS). When you force users to re-accept the end user agreement and they do it while their session is active, the entire User object in the store becomes invalidated - then frontend triggers the `this.epersonService.findById` function many times, which results in the request to the backend being marked as `Stale` endlessly, and then because of:
```typescript
const response$: Observable<RemoteData<T>> = this.rdbService.buildSingle<T>(requestHref$, ...linksToFollow).pipe(
      // This skip ensures that if a stale object is present in the cache when you do a
      // call it isn't immediately returned, but we wait until the remote data for the new request
      // is created. If useCachedVersionIfAvailable is false it also ensures you don't get a
      // cached completed object
      skipWhile((rd: RemoteData<T>) => rd.isStale || (!useCachedVersionIfAvailable && rd.lastUpdated < startTime)),
      this.reRequestStaleRemoteData(reRequestOnStale, () =>
        this.findByHref(href$, useCachedVersionIfAvailable, reRequestOnStale, ...linksToFollow)),
    );
```

The request for new user object never finishes. Fixing this without affecting other components is problematic. I asked another coworker of mine who is much better at angular than me for help, but I am not sure if we will be able to get it fixed by Friday.

The only "fix" that works is replacing this fragment in the `submit` function called when you accept the End User Agreement:
```typescript
if (isNotEmpty(redirectUrl)) {
     this.router.navigateByUrl(decodeURIComponent(redirectUrl));
}
````

which is just redirecting the user to the proper page after they accepted the agreement, into a **hard redirect**, which will still do the redirect, but also refresh the page:
```typescript
if (isNotEmpty(redirectUrl)) {
    const fullRedirectUrl = `${this.hardRedirectService.getCurrentOrigin()}${decodeURIComponent(redirectUrl)}`;
    this.hardRedirectService.redirect(fullRedirectUrl);
}
```

This will force the frontend to reload the user object and everything works well from that point onwards. However, since that is a change (albeit small) to the DSpace functionality, I need confirmation if that is okay. If not, then we will just keep trying to find the source of the issue with my coworker and try to implement a proper fix.

Dawnkai:
Okay, so:
As agreed on Developer's Meeting (26th of March 2026) with @tdonohue I had a look on a PR that attempts to fix the same issue I reported:
https://github.com/DSpace/dspace-angular/pull/5099
Unfortunately, while the fix in that PR works for infinite loading, it breaks other functionalities as Marcin Miłosz has reported in that PR's comments. My coworker also tested it and confirmed that it sadly does not fully fix the problem.

He also found the culprit and has come up with a fix. The issue is caused by recursive calls to /eperson endpoint that are cancelled all at once in RequestService but still inserted into the cache, which results in that permanent ResponsePending status. He also found out that this isn't the first time this issue happens, as he found other attempts at fixing this in git history.

**In any case, we decided to go forward with the fix I proposed**: using hard redirect instead of regular `nagivate` function when user accepts the license. This effectively refreshes the store and stops the recursive chain from happening. I also used `URLCombiner` (proposed by Art I believe).

My coworker has a better fix for this problem, however it can potentially contaminate other components, so he will simply make a new bug Issue where he will describe the problem in more detail and provide a fix. Then hopefully we can get it tested thoroughly since we won't be pressured by the DSpace 10 Release Deadline.

I tried moving around the site after accepting the user agreement with the hard redirect and found no issues, but if someone wants to give it another test that would be nice. If not then that is all we can do for now.

Thanks to everyone involved for the feedback and helpful tips again.

tdonohue:
Thanks for the update @Dawnkai !  That all sounds good to me. I'm also glad to hear your colleague has found a possible fix & look forward to seeing that bug fix PR once it's ready.  

I'll try to test this PR later today myself, so that we can try to get this feature into 10.0.  Even if it isn't possible to merge today, the Merger Management Team already approved an extension for merger-related PRs until next Wednesday.  So, if you can help to continue to address any feedback until then, I'll also do my best to ensure this keeps moving forward so that we can get it into 10.0. 

Thanks again.

tdonohue:
@EikLoe or @vins01-4science : If one of you could find time to help me with final testing (early next week), it'd be appreciated.  I really want to see this get into 10.0, but we'd ideally need to then get it merged on Monday or Tuesday (as this PR has an extension). 

I'll also get this this, but all PRs require two +1 votes. So, I'd really need one of you (or both) to help move this along.  Thanks!

tdonohue (changes requested):
@Dawnkai : I had a chance to finally test this today.

Overall, the Homepage *header* is working well, but I'm finding everything else to be a bit buggy.  Here's the issues that I've run into:

1. First, on the "Edit CMS Metadata" page, if I select any of the options, I'm presented just with text areas for each language.  I was expecting the see the same note that you see under "Edit User Agreement" which says that it supports Markdown.  (Otherwise, users may not understand how to use this page as there's no clue regarding what sort of text these boxes support.)
2. If I go to "Edit CMS Metadata" and select "dspace.cms.home-header" to add a Header, then I see it works properly.  But, if I then go in and edit "dspace.cms.home-news" or "dspace.cms.footer", then I *cannot* get those to work.  I'm using the default "dspace" theme (the one that's turned on by default in DSpace) though, so maybe that theme doesn't support those fields?
    * Even if I just put basic text in the `home-news` and `footer`, I cannot seem to get it to display at all.
4. After changing the "dspace.cms.home-header" to something with Markdown (e.g. a list of items, like I've shown below).  If I visit the homepage and reload the page, then I first see the unformatted text briefly, then the page reloads and I see the nicely formated Markdown text.
    * I'm running in production mode (`npm run build:prod && npm run serve:ssr`)
    * This behavior seems to occur when the page "flash" occurs.  It sometimes causes things to jump around on the page if the formatted text suddenly takes up more space.  For example, try to add a list like below. You'll see that list will first appear all as one line of text, then the page flashes and it reloads and you'll see a formatted list.
    ```
    _Three things:_
    * One
    * Two
    * Three
    ```

5. If I visit "Edit User Agreement", I notice that the languages appear in a different order.  Or maybe it's just that English isn't appearing first on this page.  But, I find it confusing that the languages are listed differently than on "Edit CMS Metadata".
6. The "Edit User Agreement" is still a bit buggy for me. 
    * I edited the user agreement 
    * Then I clicked on the "End User Agreement" link in the default footer
    * Then I clicked "Save".
    * After this, if I try to do anything else (like go to the Homepage) the site just hangs. I cannot do anything again until I reload the page.
    * (It's possible this issue isn't caused by this PR though. I have to dig around a bit more, but this entire User Agreement process is still not working properly)

Overall, I really appreciate your effort to port this work to DSpace.  However, I'm feeling like we need to find a way to get these bugs fixed before we can add it into a DSpace release.  If you feel these are quick to fix, we could still try for 10.0.  But, if you feel any of these are more complex, then we may need to aim this at 11.0 instead.

Dawnkai:
Thanks for the feedback, @tdonohue.

So, let me address your concerns one by one:
> First, on the "Edit CMS Metadata" page, if I select any of the options, I'm presented just with text areas for each language. [...]  I was expecting the see the same note that you see under "Edit User Agreement" which says that it supports Markdown. [...]

Sure, that is an easy change. If I remember correctly it doesn't work like that in CRIS, which is why it's not there, but it's just a simple `ds-alert` with a new translation key, so it should be a few minutes tops.

> If I go to "Edit CMS Metadata" and select "dspace.cms.home-header" to add a Header, then I see it works properly. But, if I then go in and edit "dspace.cms.home-news" or "dspace.cms.footer", then I cannot get those to work.

Be sure to remember these notes from the PR testing instructions:

- `dspace.cms.footer` requires you to edit showTopFooter in FooterComponent to true.
- `dspace.cms.home-news` requires you to remove the `dspace.cms.home-header` metadata and then set `dspace.cms.home-news` metadata. Or you can simply temporarily pull the `ds-home-news` component out of the @if statement in the `home-page.component.html` file.

Also, I have found an issue on top of it all - the footer does update, but only after page refresh. It makes sense considering that it only loads the footer CMS metadata once. This should be done using some sort of push-pull mechanism, like a service (if we want it to update without a refresh, though that will only work for the person who edited the footer, so I am not sure if that is sufficient).

Frankly, it might be worth to make some small documentation for this, because I also find this all to be pretty confusing.

> After changing the "dspace.cms.home-header" to something with Markdown (e.g. a list of items, like I've shown below). If I visit the homepage and reload the page, then I first see the unformatted text briefly, then the page reloads and I see the nicely formated Markdown text.

Yes, it's probably the SSR. I asked my colleague who is helping me out to take a look.

> If I visit "Edit User Agreement", I notice that the languages appear in a different order.

Good catch. The reason is because the User Agreement editing page is using the `keyvalue` pipe on the languages object, which usually sorts the object keys. Since CMS and User Agreement editing pages have been probably developed by 4Science in different timeframes, someone probably didn't notice. The Edit CMS page is using a regular for loop, so it is not sorting the languages before displaying them. That is also an easy few minutes fix.

> The "Edit User Agreement" is still a bit buggy for me [...]

I tested this and I think I can replicate it, but not every time. It's a pretty elusive bug. I think it's caused by the fact that the `site` object does not always load properly. If it doesn't then the editing page is sending a `null` object to the patch service, which doesn't have `uuid`, which not only dumps an error in the console, but also prevents the frontend from working correctly. Probably needs some sort of investigation, because it does not seem to be related to the Edit User Agreement page specifically.

>  If you feel these are quick to fix, we could still try for 10.0. But, if you feel any of these are more complex, then we may need to aim this at 11.0 instead.

The points: 1, 2 and 4 can be fixed pretty quickly, but 3 and 5 might be problematic. We will try to get it done asap, but I am not able to give a specific date when things should be fixed, sorry.

Maybe it would be a good idea to create bug tickets for point 3 and 5 and solve them after the merge in order to not delay it any further.

vins01-4science:
I've re-reviewed this PR and it's okay for me. Just found the same issues that @tdonohue listed above.
I am currently +1 on this, but we need to address that issues with a follow-up.

tdonohue:
@Dawnkai : Thanks for your responses.  Apologies I missed some of the testing instructions.  I'll try to retest things later today to verify the expected behavior of `dspace.cms.footer` and `dspace.cms.home-news`.  I had previously misunderstood how those were to be used (and I also agree that we'll need clear documentation on these settings).

As a sidenote, I also plan to test #5323 alongside this PR to see if that improves the behavior of the "Edit User Agreement".

src/app/home-page/home-page.component.html
```diff
@@ -1,5 +1,12 @@
-<ds-home-coar></ds-home-coar>
-<ds-home-news></ds-home-news>
+@if (homeHeaderMetadataValue$ | async; as homeHeaderMetadataValue) {
+  <div class="container">
+    <ds-markdown-viewer [value]="homeHeaderMetadataValue"></ds-markdown-viewer>
+  </div>
+}
+@else {
+  <ds-home-coar></ds-home-coar>
```
tdonohue: Why is the COAR metadata only added to the homepage if you do NOT customize the header?  This seems odd because it assumes the COAR embedded links are not important if you have a header.

I think `<ds-home-coar>` needs to always be displayed regardless of whether there's a header or not.

tdonohue (changes requested):
@Dawnkai : I've retested this today with your latest changes & alongside #5323

From [my prior feedback](https://github.com/DSpace/dspace-angular/pull/5106#pullrequestreview-4023790692), issues 1, 2, 4, and 5 are all resolved.  Issue 5 was fixed by #5323 

That leaves my prior issue 3 being the remaining problem from my initial review.  It also occurs on `dspace.cms.home-news`:
> After changing the "dspace.cms.home-header" to something with Markdown (e.g. a list of items, like I've shown below). If I visit the homepage and reload the page, then I first see the unformatted text briefly, then the page reloads and I see the nicely formated Markdown text.
>
>  * I'm running in production mode (`npm run build:prod` && `npm run serve:ssr`)
>  * This behavior seems to occur when the page "flash" occurs. It sometimes causes things to jump around on the page if the formatted text suddenly takes up more space. For example, try to add a list like below. You'll see that list will first appear all as one line of text, then the page flashes and it reloads and you'll see a formatted list.
>  ```
> _Three things:_
>  * One
> * Two
> * Three
> ```

A few additional things that I feel we may want to consider:
* I'd argue that `dspace.cms.footer` is not very useful as-is because  by default `showTopFooter=false` in `footer.component.ts`.  While I know this would be modifying the behavior, I feel we should make `showTopFooter` configurable by adding a new `homePage > showTopFooter` configuration like this:
    ```
    homePage:
        # When false, the `dspace.cms.footer` option is DISABLED (i.e. it shouldn't appear as an option to edit because it does nothing)
        # When true, the `dspace.cms.footer` option is ENABLED
        showTopFooter: true/false
    ```
* I'd also argue that it doesn't make sense for `dspace.cms.header` and `dspace.cms.homepage-news` to be mutually exclusive.  It should be possible to configure both at the same time, as it doesn't make sense that only one can be used.

Changing the behavior as I describe would also solve the documentation problem. Currently, the editing interface is very confusing because it's not clear that `dspace.cms.footer` won't work by default and `dspace.cms.header` and `dspace.cms.homepage-news` are mutually exclusive.

Finally, I found small bug where the COAR links are NOT included in the header if you have custom header.  I've noted it inline below.

src/app/home-page/home-page.component.html
Dawnkai: No idea, Tim. This is also how it was done in CRIS:
https://github.com/4Science/dspace-angular/blob/main-cris/src/app/home-page/home-page.component.html
image


I will take it out of that if, as you requested.

Dawnkai:
@tdonohue : I have added an option to change the `showTopFooter` using the appConfig instead of a hard-coded value like you asked. I have also moved the coar component outside of the `@if` in the home page component. If the `showTopFooter` is not enabled, the cms metadata for the footer will not appear in the CMS metadata select like you asked as well. I have also added a little bit of the vertical margin to that home header so that the margins are even:
image

As for the flashing problem:
I have investigated the Markdown flashing issue during SSR with my colleague, and we've identified the root cause.
The behavior occurs during a brief window between the moment the browser receives the HTML from the backend and the loading of the Markdown and Mathjax libraries. When the libraries are not loaded yet, the browser displays the text in unformatted form. Since this is a byproduct of browser rendering cycles rather than a bug within the CMS PR itself, a permanent "fix" isn't possible, but we can mitigate the visual effect.
We've seen in other places in the code that a temporary overlay is used to mask this transition. We can also use the same approach here. However, due to the merge deadline for the current PR, I suggest we merge the current PR and open a separate issue specifically for this loading problem.
Then we can address the problem in a follow-up PR without delaying the integration of CMS functionality.
Is that okay with you?

Also, the issue with the flashing is not exclusive to the home news component. If you try to add markdown to the simple page component (for example: to the item abstract), the same flashing problem appears:
image
You can see the unformatted list at the end of the abstract in the screenshot above. Then, after some time, it renders properly, as shown below:
image

tdonohue:
Thanks for your updates & hard work on this PR, @Dawnkai !  I'll retest this today.  I'm OK with moving the flashing problem to a separate bug ticket as you suggested, so that we can merge this PR into 10.0 without that fix.  I can create that new ticket once I finish my test/review of this PR.

tdonohue (approved):
👍 Thanks @Dawnkai !  This all looks good enough to me now.  I like the improved behavior of adding the `showTopFooter` config and displaying both the `home-news` and `home-header` when they are both configured.

That said, in my testing today I found a few minor bugs.  I'm going to move them to bug tickets though for follow-up work.

In total, there will be three bug tickets that I'm going to log for follow-up work:
1. The flashing issue previously mentioned above when using Markdown. Moved to #5338
2. There's a small difference in behavior between the End User Agreement and the Header/News/Footer with regards to how they "fallback" if not all language fields have been filled out.  Moved to #5339
     * For example, if you only add an English "End User Agreement", then if you switch languages to Deutsch (German), the fallback is to display the English "End User Agreement" that you configured. (I think this behavior makes sense)
     * However, the bug is that if you only add an English Header/News/Footer, and then switch languages to Deutsch (German), the fallback is to **not display any custom Header/News/Footer**.  This seems like a bug to me as the default language is English.
3. The forms used to edit all these fields currently have a number of accessibility issues. Moved to #5340

I'll create the above three tickets later today.  But, as this PR is at +2 (from me and @vins01-4science ), I'll merge it immediately.  Thanks again @Dawnkai for all your hard work in getting this ready to go.

Finally, as a sidenote, I'm flagging this as "needs documentation" as we need to add appropriate docs for this feature to https://wiki.lyrasis.org/display/DSDOC10x

Dawnkai:
Hi @tdonohue,
I added the documentation for both functionalities to the DSpace 10 Wiki:
https://wiki.lyrasis.org/display/DSDOC10x/Content+Management+System

tdonohue:
Thanks for the docs @Dawnkai !  They look good to me.  Removing the "needs documentation" label from this ticket

PR #5116, opened 2026-02-15
fix: align action buttons in empty bundles on Bitstreams tab

### LINKED ISSUE #5115: Visual misalignment of action buttons in empty DSpace bundles (Bitstreams tab in Edit Item view)
## Describe the bug
Observed in Chrome v145.

**For empty bundles**, the 'Actions' column button group (with the upload and settings buttons) is not vertically aligned. Observed in sandbox.dspace.org. Screenshot from there:

Image

Everything looks okay if a bundle has **at least one bitstream**:

Image

## To Reproduce
1. Go to Item -> select Edit this item -> select Bitstreams tab
2. Press Upload, insert name for a bundle & press Create bundle
3. Go back to the Bitstreams tab. This way, create at least two bundles.
4. Observe the Actions column for button misalignment.

## Expected behavior
Action buttons should be vertically aligned no matter how many bitstreams are in a bundle

## Related work
Not found

LakshmiSravyaVedantham:
## Summary
- Add missing `columnSizes.columns[3].buildClasses()` to the bundle-row action `<td>` in `item-edit-bitstream-bundle.component.html`
- The header `<th>` and bitstream-row `<td>` already use this class to size the Actions column, but the bundle-row `<td>` was missing it
- Without this class, when a bundle is empty (no bitstreams), the action buttons don't align vertically because the column width isn't constrained

Closes #5115

## Test plan
- Create 2+ empty bundles in Edit Item → Bitstreams tab
- Verify the upload and settings buttons are vertically aligned across all empty bundles
- Verify bundles with bitstreams still display correctly

### DIFF
--- a/src/app/item-page/edit-item-page/item-bitstreams/item-edit-bitstream-bundle/item-edit-bitstream-bundle.component.html
+++ b/src/app/item-page/edit-item-page/item-bitstreams/item-edit-bitstream-bundle/item-edit-bitstream-bundle.component.html
@@ -34,7 +34,7 @@
               <th id="{{ bundleName }}" class="row-element" colspan="3" scope="colgroup">
                 {{'item.edit.bitstreams.bundle.name' | translate:{ name: bundleName } }}
               </th>
-              <td class="text-center row-element">
+              <td class="text-center row-element {{ columnSizes.columns[3].buildClasses() }}">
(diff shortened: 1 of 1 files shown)


MMilosz:
Tested locally. LGTM!

# Before
image

# After
image
image

Issue #3384, opened 2024-10-04
Accessibility or discoverability of bitstream download links


alanorth:
## Is your feature request related to a problem? Please describe.
We have reports from users that it is not very obvious how to download bitstreams from the item view page. The links under the "Files" section of the sidebar are apparently not enough. I'm not sure if this is an accessibility or discoverability issue with regards to UI/UX.

We always encourage our users to use Handles instead of direct bitstream links when they share, but they complain that the item page is not very user friendly.

## Describe the solution you'd like
In earlier DSpace versions a user could click the thumbnail to download the associated bitstream. I can't recall exactly if this was a customization in our XMLUI theme or not. It would be nice in cases where an item has at least one bitstream with a thumbnail, that a user could click the thumbnail and download the associated bitstream.

For sure when an item has a thumbnail generated by DSpace it is such a tempting target to click—yet it does nothing!

![Screenshot 2024-10-04 at 13-00-22 The Impact of Dietary Fiber on Gut Microbiota in Host Health and Disease](https://github.com/user-attachments/assets/e154ac06-bacd-42df-8665-1b1638f73c77)

See: https://demo.dspace.org/handle/123456789/1142

## Additional information

- May need consideration regarding "primary" bitstreams
- May need consideration regarding items with multiple bitstreams

oscar-escire:
Hi @tdonohue I've been working on a PR with a proposal for this Issue. Can you Assign it to me? I'll share the PR over next days

abollini:
Just found this issue, please note that it is addressed in DSpace-CRIS, so it will come in DSpace as part of the merger effort via this PR https://github.com/DSpace/dspace-angular/pull/5132

PR #12, opened 2016-12-05
Added production startup script: this connects to #11

wwelling:
Added production startup script.

Please test pull requests with ```npm run start:prod```

This will build and run in production using AoT compilation.

PR #5112, opened 2026-02-13
[DSpace-CRIS] Share links via social networks

pzlakowski:
## References

Fixes DSpace/DSpace#11781

## Description

This PR ports social networks feature from DSpace-CRIS

We might need to adjust the [wiki page too](https://wiki.lyrasis.org/display/DSPACECRIS/Share+Content). 

## Instructions for Reviewers

List of changes in this PR:
* Ports share social links from CRIS
* Fixes the feature due to breaking changes in AddToAny library

Instructions to test this PR:

1. Enable social links since they are disabled by default

`config/config.development.yml`

```yaml
# ...

addToAnyPlugin:
  scriptUrl: "https://static.addtoany.com/menu/page.js"
  socialNetworksEnabled: true
  buttons:
    - facebook
    - twitter
    - linkedin
    - email
    - copy_link
  showPlusButton: true
  showCounters: true
  title: DSpace demo
```

2. Visit home page, search page and item page
    * these are the only pages which use this feature in DSpace-Cris, the metadata `showSocialButtons` boolean needs to be added in routes for parent definition

## Checklist
_This checklist provides a reminder of what we are going to look for when reviewing your PR. You do not need to complete this checklist prior creating your PR (draft PRs are always welcome).
However, reviewers may request that you complete any actions in this list if you have not done so. If you are unsure about an item in the checklist, don't hesitate to ask. We're here to help!_

- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [ ] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [ ] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [ ] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
- [ ] My PR **aligns with [Accessibility guidelines](https://wiki.lyrasis.org/display/DSDOC9x/Accessibility)** if it makes changes to the user interface.
- [ ] My PR **uses i18n (internationalization) keys** instead of hardcoded English text, to allow for translations.
[...]

PR #5132, opened 2026-02-20
[DSpace-CRIS] Download button & Allow authors to download restricted files

### LINKED ISSUE #12169: [DSpace-CRIS] Download button & Allow authors to download restricted files
# References
* This is a sub-issue of #11749 

# Description

The DSpace-CRIS feature "“Download” button instead of file link" should be ported to DSpace.

Below is the description from the analysis of [DSpace-CRIS and DSpace Differences](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences):

> 12. [Item page] “Download” button instead of file link. ([see page 126 of the pdf technical documentation](https://github.com/4Science/DSpace/releases/download/dspace-cris-2024.02.00/dspace-cris-2024.02.00.pdf)).  This is just an alternative option available. Can choose whether to show file links (like DSpace) or a download button.
>     1. Should be mostly independent of other features

FrancescoMolinaro:
## References
* Fixes https://github.com/DSpace/DSpace/issues/12169
* Relates to: [Feature differences #12](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences#:~:text=%5BItem%20page%5D%20%E2%80%9CDownload%E2%80%9D%20button%20instead%20of%20file%20link.%20(see%20page%20126%20of%20the%20pdf%20technical%20documentation).%C2%A0%20This%20is%20just%20an%20alternative%20option%20available.%20Can%20choose%20whether%20to%20show%20file%20links%20(like%20DSpace)%20or%20a%20download%20button.) and [Feature differences #13](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences#:~:text=Authors%20can%20download%20restricted%20bitstreams%3A%C2%A0Regardless%20of%20permissions%20set%20on%20Bitstream%2C%20the%20author%20can%20always%20download%20it.)
## Description

Add new way of rendering download link in item page, the new UI component can be activated via the config:

showDownloadLinkAsAttachment 

inside the layout config in default.app.config.ts.

This PR adds a route guard that allows to run the required authorization checks also on direct download links, supporting in this way the [Feature differences #13](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences#:~:text=Authors%20can%20download%20restricted%20bitstreams%3A%C2%A0Regardless%20of%20permissions%20set%20on%20Bitstream%2C%20the%20author%20can%20always%20download%20it.), which has been implemented on https://github.com/DSpace/DSpace/pull/11964 , even though is not strictly depending on that rest PR.

image

The metadata to be visualized in the attachment box are handled with a separated config param:

advancedAttachmentRendering

Which allows granural control on which information to show in the template.

In addition a new route guard has been added for the bitstream download path, so that we can check the authorization before loading the page and handle in advance eventual authorizations needed.

On the rest side the canDownload authorization feature will be enhanced to allow authors the download of restricted items.

## Instructions for Reviewers

The new config has been set as true for the time being so that every link in item page will be displayed with the new component.
The positioning was also adapted from the left side of the page to the right one, so to use the wider space for the additional informations. 

List of changes in this PR:

Add new UI component and related configs
Add new guard for bitsream download path

## Checklist


tdonohue:
@FrancescoMolinaro : Same here, please specify in the PR description which [DSpace-CRIS differences](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences) this PR implements/relates to.

tdonohue:
@FrancescoMolinaro : Just to verify, does this PR require any backend PR?  Or is it standalone?  It's unclear to me whether this requires specific backend changes to work.

FrancescoMolinaro:
Hi @tdonohue , sorry for missing to specify it, this PR is mainly a UI variant of the normal link and it is indeed standalone.

lgeggleston:
@aprilherron General reminder for all new feature PRs currently under review: there is still a possibility of getting this into 10.0, in case you have the capacity to complete the review at the moment.   Feature merge deadline will be Friday 3/27. Thank you!

src/app/item-page/simple/item-types/untyped-item/untyped-item.component.html
```diff
@@ -30,7 +30,9 @@
         <ds-media-viewer [item]="object"></ds-media-viewer>
       </div>
     }
-    <ds-item-page-file-section [item]="object"></ds-item-page-file-section>
+    @if (showDownloadLinkAsAttachment) {
```
aprilherron: This should be `showDownloadLinkAsAttachment !== true` so that the default file links show if the new Download section is disabled

aprilherron:
Review notes:

#### Config syntax

Feature is enabled by default.
To turn it off, use config:

```
layout:
  showDownloadLinkAsAttachment: false
```


#### Tests

1. If it is enabled the new bitstream view shows on the item-page.
2. If it is disabled the new bitstream view doesn't show on the item-page
    => But the normal download links **DO NOT** appear (marked spot in code for fix in code-review)


3. Make sure every field in the bitstream-view is correct:

    - [x] Name = dc.title
    - [x] Size = the same in edit > bitstream
    - [x] Format = the same in edit > bitstream
    - [x] Checksum = the same as in the DB


4. Multiple bitstream:

    - [x] Both bitsteams still have the correct field values
    - [x] The each download button downloads the correct bitstream


5. Unexpected values:

    - [x] Creating a bitstream with a huge filename doesn't overflow the bitstream-view


6. Authorization:

    - [x] Anon users can see and use the bitstream-view -> if the bitstream is openAccess
    - [x] Normal users can see and use the bitstream-view -> if the bitstream is openAccess
    - [x] Admin users can see and user the bitstream-view -> if the bitstream is openAccess
    - [x] Anon users can see and use the bitstream-view, but are unable to download the bitstream. They can request a copy instead -> if the bitstream is **not** openAccess
    - [x] Normal users can see and use the bitstream-view, but are unable to download the bitstream. They can request a copy instead -> if the bitstream is **not** openAccess
    - [x] Admin users can see and use the bitstream-view, and are able to download the bitstream -> if the bitstream is **not** openAccess

7. Security:

    - [x] Valid html in the filename `<b>test.pdf</b>` does not render as HTML

FrancescoMolinaro:
Hi @aprilherron , thanks for the feedback, I have set the property as false by default so that the new layout will be optional. In addition I have also fixed the issue with the missing default download link and implemented some test to enforce the behavior, now it should work as expected, thanks again for the review.

tdonohue:
@aprilherron : If you can find time to rereview this (based on the updates), I'd appreciate it.  Please also remember to explicitly give the PR a "+1" if you feel it's working well and ready to be merged.

@FrancescoMolinaro : I've yet to review/test this, but we may want to consider enabling this by default if it accurately fixes #3384 (as that's a long standing accessibility/usability issue).  That said, I'll try to give this a test tomorrow to see if this should be enabled by default.

aprilherron (approved):
Bug in untyped-item view has been addressed. This PR looks good to be merged 👍

tinsch:
@FrancescoMolinaro I ran some response time performance tests on your branch, because I recently worked on a similar area of the code (https://github.com/DSpace/dspace-angular/pull/5028). I found out that your new code runs much faster on larger page sizes with `showDownloadLinkAsAttachment` enabled than the 'old' code for the bitstream download links. Congrats 👏🎉

Could you maybe explain a bit what you are doing with this bitstream route guard, and if we could also use it for the 'normal' bitstream download links?

Here is a diagram from my performance tests, so that you can see how much better the new feature performs:
branch-review

tdonohue:
@FrancescoMolinaro : Could you quickly rebase this PR?  It appears to have gained merge conflicts after the merger of #5112

FrancescoMolinaro:
Hi @tinsch , thanks for the feedback much appreciated, the bitstream route guard is currently already active for all the downloads as it runs on the download page indipendently from the property showDownloadLinkAsAttachment.
It's purpose is to check if the user is authorized to access the download path, the main point is to check authorizations on direct link, e.g. accessing a link like https://demo.dspace.org/bitstreams/55113c70-2f96-4f66-acbc-692987fcedcc/download from a shared link.

I believe that what makes the file-download-link.component.html slow to render on pages, is the template of the component itself, which creates many unnecessary subscriptions, creating the risk of memory leaks.

I have made a commit to change this as I think is a good improvment in any case, you can check the last commit of this PR for confirmation on your tests, hopefully performaces will be better also with the property set to false.

tinsch:
Thanks a lot! I will look into it and retest it 👍

config/config.example.yml
```diff
@@ -706,6 +706,18 @@ layout:
         default:
           icon: fas fa-project-diagram
           style: text-success
+  # If true the download link in item page will be rendered as an advanced attachment, the view can be then configured with the layout.advancedAttachmentRendering config
+  showDownloadLinkAsAttachment: true
+  # Configuration for advanced attachment rendering in item pages. This controls how files are displayed when showDownloadLinkAsAttachment is enabled.
+  # Each configuration maps a bundle name to specific metadata fields that should be displayed alongside the file.
+  advancedAttachmentRendering:
```
tdonohue, line 713: The `pagination` setting appears to be missing from the examples here.

Also, I'd recommend the configuration in `config.example.yml` match the default configuration that you've set in `default-app-config.ts`.  That makes it much easier for users to understand what the default configs are.

config/config.example.yml
```diff
@@ -706,6 +706,18 @@ layout:
...
+    - bundle: ORIGINAL
+      metadata:
+        - dc.description
+        - dc.title
+    - bundle: LICENSE
+      metadata:
+        - dc.rights
```
tdonohue: Why is the structure of these `bundle` and `metadata` configs so much different here then in `default-app-config.ts`?  Are these settings correct?

src/config/default-app-config.ts
```diff
@@ -748,6 +749,42 @@ export class DefaultAppConfig implements AppConfig {
         },
       },
     ],
+    showDownloadLinkAsAttachment: false,
+    advancedAttachmentRendering: {
+      pagination: {
+        enabled: true,
+        elementsPerPage: 2,
```
tdonohue: This default setting of 2 download sections per page doesn't appear to be working for me.  If I visit an Item that has 5+ Bitstreams, then I see a list of 5 download sections before pagination (which corresponds to the `item > bitstream > pageSize` configuration default value).  

For example, this screenshot shows the last three bistreams (out of 5 total) and then the "Show more" button appears:

Image

We should either fix this configuration, or (alternatively) we could **remove these configurations**.  Pagination should always be enabled, and the elements per page is already configurable via `item > bitstream > pageSize`.  

So, honestly, I think we could remove these settings.

src/app/bitstream-page/bitstream-download-redirect.guard.ts
```diff
@@ -0,0 +1,102 @@
...
+} from '../core/shared/bitstream.model';
+import { FileService } from '../core/shared/file.service';
+import { getFirstCompletedRemoteData } from '../core/shared/operators';
+import {
+  hasValue,
+  isNotEmpty,
+} from '../utils/empty.util';
+
+
+export const bitstreamDownloadRedirectGuard: CanActivateFn = (
```
tdonohue, line 64: Please add code comments to describe the behavior of this new guard.

src/app/core/services/server-hard-redirect.service.ts
```diff
@@ -97,4 +104,12 @@ export class ServerHardRedirectService extends HardRedirectService {
   getBaseUrl(): string {
     return environment.ui.baseUrl;
   }
+
+  /**
+   * Set CORS header to allow embedding of redirected content.
+   * The actual security header will be set by the rest
+   */
+  setCorsHeader() {
+    this.responseService.setHeader('Access-Control-Allow-Origin', '*');
```
tdonohue: I want to understand why this is being set to a wildcard/asterisk.  This can be a potential security flaw as it allows resources to be accessed by any domain (even untrusted ones).

Why do we need this to be set to an asterisk?  Is there a way to instead set it to the UI's URL?  As of #5276 it should now be possible to access the URL of the User Interface via the `environment.ui.baseUrl` setting.  So, that might be a better value here, assuming it works as well.

src/assets/i18n/en.json5
```diff
@@ -1747,6 +1747,32 @@
 
   "cookies.consent.purpose.sharing": "Sharing",
 
+  "cris-layout.advanced-attachment.dc.title": "Name",
```
tdonohue: Could we consider changing all these new keys to start with `layout` instead of `cris-layout`?  That way they'd match with the new `layout` configuration section where they are used.

tdonohue (changes requested):
@FrancescoMolinaro : I gave this a review and test today.  Overall, it works, but I've found some issues in the code which I think need cleanup/correction (see inline below).

I also have a few basic questions here:
1.  I don't understand how to configure `advancedAttachmentRendering > metadata` fields, as the examples provided in `config.example.yml` seem to be invalid.
2. I also don't see how this PR is providing authors with the ability to download restricted files.  Is that provided via a backend PR?  If so, which backend PR does this feature require?

I have been able to verify that the basics of rendering bitstreams in a list with a download button works properly.

config/config.example.yml
```diff
@@ -706,6 +706,18 @@ layout:
         default:
           icon: fas fa-project-diagram
           style: text-success
+  # If true the download link in item page will be rendered as an advanced attachment, the view can be then configured with the layout.advancedAttachmentRendering config
+  showDownloadLinkAsAttachment: true
```
tdonohue: @FrancescoMolinaro : One last note, this setting should match the default value in `default-app-config.ts`.

For now, I'd say maybe we should keep this disabled by default.  However, if we receive positive feedback on the display, we can modify it at a later time.

Personally, I like the new look overall. But I do find displaying the same thumbnail twice on the Item page seems a bit odd. For example, with an Item with one bitstream, you see its thumbnail twice near the top of the page:

Image

tinsch:
> Thanks a lot! I will look into it and retest it 👍

I ran the performance tests again, and the results are nearly the same now with and without the new `showDownloadLinkAsAttachment` feature enabled. Both better than on current main. Thanks a lot! 👏

src/app/core/services/server-hard-redirect.service.ts
FrancescoMolinaro: Hi @tdonohue , this bit of code was implemented to allow the embedding of the mirador viewer in external pages, so under url different from the UI one.
Is nevertheless connected to a logic that is not yet in place so I am going to remove it.

config/config.example.yml
FrancescoMolinaro: Hi @tdonohue , I agree that is a bit odd to see the double thumbnail, this new layout is indeed used in CRIS together with the layout configuration through the excel sheet, in that scenario we can configure for which bitstream we display the attachment component and which image to use for the thumbnail of the item page, so that we don't have the duplicated thumbnail.
Considering this I agree that this should be kept as disabled by default, when the CRIS custom layout mechanism will be ported we will be able to configure it granularly on each entity type.

src/app/core/services/server-hard-redirect.service.ts
tdonohue: Thanks for letting me know @FrancescoMolinaro !  I'd definitely recommend removing it until it is necessary.  That said, I will have the same concerns at the point where this code is necessary.  I'm worried about the security of this change, especially since I don't quite understand the purpose.  But, we can deal with that discussion in a future PR.

FrancescoMolinaro:
Hi @tdonohue , regarding the functionality that makes possible for authors to download restricted files the logic can be found on https://github.com/DSpace/DSpace/pull/11964, it has been included in the same PR of the edit in submission style.
The frontend PR is not necessarily depending on that one as the code doesn't add any extra check, but adds a route guard that allows to run the required checks also on direct download links.

I will update the description adding this information.

src/config/default-app-config.ts
tdonohue: @FrancescoMolinaro : I see you fixed all the other issues that I mentioned except for this one.  Any thoughts on whether we should remove the new `advancedAttachmentRendering > pagination > elementsPerPage` setting in favor the existing `item > bitstream > pageSize` setting.  Based on my testing, the existing `pageSize` setting is the one being used, and the new `elementsPerPage` seems to do nothing.

FrancescoMolinaro: Hi @tdonohue , you are absolutely right, I ported also the pagination config that will be necessary only at a later point in the merger activity when the configurable layout will be ported as well.
For the time being I removed the pagination configuration entirely as not necessary, thanks for the catch.

config/config.example.yml
```diff
@@ -706,6 +706,32 @@ layout:
         default:
           icon: fas fa-project-diagram
           style: text-success
+  # If true the download link in item page will be rendered as an advanced attachment, the view can be then configured with the layout.advancedAttachmentRendering config
+  showDownloadLinkAsAttachment: false
+  # Configuration for advanced attachment rendering in item pages. This controls how files are displayed when showDownloadLinkAsAttachment is enabled.
+  # Defines pagination settings and which metadata/attributes to display for bitstream attachments.
+  advancedAttachmentRendering:
+    # Pagination configuration for attachment lists
+    pagination:
```
tdonohue: @FrancescoMolinaro : Tiny thing, but you forgot to remove the `pagination` examples from the `config.example.yml`.  You should also update the comment above so it no longer says "pagination".. so it should say:
"Defines which metadata/attributes to display for bitstream attachments."

FrancescoMolinaro: Hi @tdonohue, sorry for the oversight, it should be ok now.

config/config.example.yml
```diff
@@ -706,6 +706,28 @@ layout:
...
+  showDownloadLinkAsAttachment: false
+  # Configuration for advanced attachment rendering in item pages. This controls how files are displayed when showDownloadLinkAsAttachment is enabled.
+  # Defines which metadata/attributes to display for bitstream attachments.
+  advancedAttachmentRendering:
+    # Metadata and attributes to display for each attachment
+    metadata:
+      - name: dc.title
+        type: metadata
+        truncatable: false
+      - name: dc.type
```
tdonohue, line 719: These metadata settings don't appear to work for me?  For example, if I remove some of the fields like this:
```
layout:
  showDownloadLinkAsAttachment: true
  metadata:
    - name: dc.title
      type: metadata
      truncatable: false
    - name: format
      type: attribute
```

I'd expect to then **no longer see the checksum and size** in the display.  However, I still see the same display as before.

Image

Am I misunderstanding how these settings should work?

tdonohue (commented):
@FrancescoMolinaro : Thanks for all the updates!  I'm nearly a +1 to this PR.  Almost everything works well. However, I'm not able to get the `metadata` settings to work properly.  See inline comment below.

config/config.example.yml
FrancescoMolinaro: Hi @tdonohue , I think you are missing a level of the configuration, the metadata config are inside the advancedAttachmentRendering property, I have tested with the following config and seems working as expected:

```
layout:
  showDownloadLinkAsAttachment: true
  advancedAttachmentRendering:
    metadata:
      - name: dc.title
        type: metadata
        truncatable: false
      - name: format
        type: attribute
```

tdonohue: Ah, my mistake!  I'll retest today then and get this PR merged

tdonohue: Thanks @FrancescoMolinaro !  I retested today and verified it works well (now that I'm using the correct configuration)!  Thanks again for correcting my mistake :)

tdonohue (approved):
👍 Thanks @FrancescoMolinaro !  This all looks great to me now.  Retested today and it's all working well.  

As with all other DSpace-CRIS merger PRs, I'm flagging this with "needs documentation" until documentation can be added to https://wiki.lyrasis.org/display/DSDOC10x

tdonohue:
@FrancescoMolinaro : This PR also appears to still be missing documentation for the new settings.  They should minimally be added to https://wiki.lyrasis.org/display/DSDOC10x/User+Interface+Configuration

FrancescoMolinaro:
Hi @tdonohue , thanks for the note and the patience, the new config should now be described under: https://wiki.lyrasis.org/spaces/DSDOC10x/pages/408944956/User+Interface+Configuration#UserInterfaceConfiguration-Layoutsettings

tdonohue:
Docs found at:
* https://wiki.lyrasis.org/spaces/DSDOC10x/pages/408944956/User+Interface+Configuration#UserInterfaceConfiguration-Layoutsettings
* https://wiki.lyrasis.org/spaces/DSDOC10x/pages/438698097/Allow+Authors+to+download+restricted+bitstreams

Issue #14, opened 2016-12-07
Add an npm script to install global dependencies


artlowel:
Some npm scripts won't work unless certain dependencies are installed globally. `webpack-bundle-size-analyzer` is one such example. We need to identify the others, and add an npm script to install them.

wwelling:
Here is what I would suggest:
```"global": "npm install -g angular-cli nodemon npm-check-updates rimraf ts-node typedoc typescript webpack webpack-bundle-size-analyzer marked node-gyp"```

PR #5134, opened 2026-02-20
[DSpace-CRIS] Inverse Relations via Authority Framework (Frontend)

FrancescoMolinaro:
## References

* Fixes https://github.com/DSpace/DSpace/issues/12164
* Requires backend PR: https://github.com/DSpace/DSpace/pull/11963
* Relates to: [Architectural difference #3](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences#:~:text=Relationships%20(between%20Entities)%20are%20changed%3A%20While%20DSpace%2DCRIS%20uses%20the%20same%20Entities%20concept%20as%20DSpace%2C%20it%20changes%20the%20%E2%80%9CRelationships%E2%80%9D%20between%20these%20Entities.)

## Description

Add new component for tabbed relations in item page.
The new relations are based on a scope parameter which represent the authority of the item in question.
Visually the component still uses the standard search page and is identical:

image

The new relations can be activated via a flag param in the default.app.config.ts, inside the item config:

showAuthorityRelations

if true the ui components will use the new authority based relations.

## Instructions for Reviewers

List of changes in this PR:

Add new UI component for new scope/authority based relations

**Include guidance for how to test or review your PR.** This may include: steps to reproduce a bug, screenshots or description of a new feature, or reasons behind specific changes. 

## Checklist
_This checklist provides a reminder of what we are going to look for when reviewing your PR. You do not need to complete this checklist prior creating your PR (draft PRs are always welcome).
However, reviewers may request that you complete any actions in this list if you have not done so. If you are unsure about an item in the checklist, don't hesitate to ask. We're here to help!_

- [ ] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [ ] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [ ] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [ ] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [ ] My PR **includes [TypeDoc](https://typedoc.org/) comments** for _all new (or modified) public methods and classes_. It also includes TypeDoc for large or complex private methods.
- [ ] My PR **passes all specs/tests and includes new/updated specs or tests** based on the [Code Testing Guide](https://wiki.lyrasis.org/display/DSPACE/Code+Testing+Guide).
[...]

PR #5145, opened 2026-02-20
[DSpace-CRIS] Administrative Edit of archived items via submission form and security configuration for metadata visibility (Frontend)

### LINKED ISSUE #12166: [DSpace-CRIS] Administrative Edit of archived items via submission form and security configuration for metadata visibility
# References
* This is a sub-issue of #11749 

# Description

The five DSpace-CRIS features "Edit Item page provides a submission-like display" (`#9`), "Granualr Edit/Security Privileges" (`#3`) and "Shared Workspaces" (`#14`), "Authors can download restricted bitstreams" (`#13`) and "Submission forms and workflow linked to collection via UI" (`#29`) should be ported to DSpace.

Below is the descriptions from the analysis of [DSpace-CRIS and DSpace Differences](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences):

> 9. [Item page] ["Edit Item" page provides a submission-like display for easier editing](https://wiki.lyrasis.org/display/DSPACECRIS/How+is+it+different+than+DSpace#HowisitdifferentthanDSpace-Friendly&safeeditingofexistingrecord%E2%80%8B): ([documented at pages 31-135 of the pdf](https://github.com/4Science/DSpace/releases/download/dspace-cris-2024.02.00/dspace-cris-2024.02.00.pdf)) When editing an Item, you see the same forms used in the Submission process.  This allows for better validation of fields, etc.  You still have the option to "Administer" which brings you to the DSpace "Edit Item" display.
>     1. Can configure additional edit modes / options, which link to different submission form configurations.  You can then define who has permissions to each edit mode, to give different users the ability to edit different parts of the Item/Entity
>     1. Depends on Granular Security Privileges feature?

> 3. [Granular Edit/Security Privileges](https://wiki.lyrasis.org/x/1IG3Fg#HowisitdifferentthanDSpace-GranularSecurityMetadata): ([documentation 114, 207 - 211 of the pdf](https://github.com/4Science/DSpace/releases/download/dspace-cris-2024.02.00/dspace-cris-2024.02.00.pdf)) DSpace-CRIS allows sites to configure more granular edit privileges or security on a metadata field by metadata field basis. Each entity type can have a custom “edit configuration” (defining who can edit individual fields, including non-admins).  Individual metadata field values may also be marked “restricted” (which hides those values from public view) in an additional metadata table column.
>     1. These tools also allow Researchers more control over their [Researcher Profile](https://wiki.lyrasis.org/display/DSDOC8x/Researcher+Profiles) (e.g. what fields are displayed publicly vs private, etc)
>     1. Managed via a Spring configuration file.

> 14. [Submission] [Shared workspaces](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences#):  Can be enabled at the Collection level.  It allows you to share a WorkspaceItem with all other Collection Submitters to work on it collaboratively.  Or you can define a custom rule to share the WorkspaceItem just with other authors of the publication.
>     1. Based on the same functionality as Supervisor Orders in DSpace, but it's not the same as the Supervisor Order.  WorkspaceItem is shared between all the users (via Resource Policies), but is also indexed into Solr.
>     1. Does not manage concurrent editing of an Item.  But, this problem also exists in DSpace with Supervisor Orders, etc.
>     1. Should be mostly independent of other features

> 13. Authors can download restricted bitstreams: Regardless of permissions set on Bitstream, the author can always download it.
>     1. If your Researcher Page is linked as a "dc.contributor.author" on the publication, then you can download any restricted files.  This is a similar concept to the ORCID feature already in DSpace.  
>     1. This is a general DSpace-CRIS feature that allows you to define permissions based on a "business rule".  These "business rules" are based on the metadata on the object.  This allows you to say, if you are an author on this object, you have extra permissions on that object.  These business rules are configurable, and this "Author can download restricted bitstreams" rule is a default configuration provided in DSpace-CRIS
>     1. Depends on Entities being enabled. Other dependencies need to be determined.

> 29. submission forms and workflow linked to collection via UI: DSpace-CRIS has additional fields in the edit-collection screen that allow to select which submission-form and workflow apply to the collection overriding what is defined in the corresponding xml files

FrancescoMolinaro:
## References

* Fixes DSpace/DSpace#12166
* Requires backend PR https://github.com/DSpace/DSpace/pull/11964

Relates to:
* [Feature differences #9](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences#:~:text=%5BItem%20page%5D,Edit%20Item%22%20display.)
* [Feature differences #3](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences#:~:text=DSpace%2DCRIS%20allows%20sites,additional%20metadata%20table%20column.)
* [Feature differences #14](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences#:~:text=%5BSubmission%5D%C2%A0Shared%20workspaces%3A%C2%A0%20Can%20be%20enabled%20at%20the%20Collection%20level.%C2%A0%20It%20allows%20you%20to%20share%20a%20WorkspaceItem%20with%20all%20other%20Collection%20Submitters%20to%20work%20on%20it%20collaboratively.%C2%A0%20Or%20you%20can%20define%20a%20custom%20rule%20to%20share%20the%20WorkspaceItem%20just%20with%20other%20authors%20of%20the%20publication.)

## Description

With this PR we introduce a new menu voice in the item page menu to access the editing in submission form, even if archived.

image

the old menu voice for editing has been renamed to "Administer" to differentiate the different set of functionalities behind the menu.

The new menu will allow to open a submission form populated with the existing metadata for further enrichment of the item.

Along this feature, the possibility to pick wich form to use  for each collection submission has been introduced:

image

This PR brings also a new configuration for the MyDspace page, called otherworkspace , which allow to have a shared workspace accessible to administrators of collection where they will be placed, to other Eperson according to a configurable set of metadata, other than to their submitter. 

image

Along the editing changes, this PR brings also the possibility to define security levels for the metadata, this require a rest config as follows, inside metadata-security.cfg:

metadatavalue.visibility.settings = [0 1 2]
metadatavalue.visibility.Person.settings = [0 1]
metadatavalue.visibility.Person.dc.date.available.settings = [0 1]

as you can see is possible to have three different level values and is possible to define them globally, based on entity type or based on the couple entity type/metadata.
Configuration lookup follows a fallback logic, if for a given metadata, no value is defined, default metadata security settings for the entity (metadatavalue.visibility.<EntityType>.settings) applies. If neither at entity level metadata security is defined, default metadatavalue.visibility.settings field value is used. 
A null value means that for a metadata, or its fallback, no security rules must be proposed to the submitter or to the editor.

Out of the box,  CRIS has 3 different security levels: 

level 0 (Public): metadata value is available to all users, even Anonymous in case entity is available to them
level 1 (Trusted): metadata value is available only to logged in users members of a defined group, named “Trusted”, as a prerequisite, this group must be available in DSpace-CRIS 7 installation
level 2 (Admin and Owner): metadata value is available only to users belonging to the “Administrators” group or to the owner of the DSpace-CRIS 7 entity. 

On the UI a toggle representing each value will be diplayed in submission or in administrative view:

image

image

The effectivness of the security level is although not yet visible in this PR, as it will require the CRIS custom layout feature that will be made available with https://github.com/DSpace/DSpace/issues/11779.

Once the layout feature is ported the metadata will be visible based on the security level and the user permission.

List of changes in this PR:

Adapted collection form for submission definition and entity type selection.
Added new way to edit archived item in submission style.
Implemented possibility to define security levels on metadata.
Added possibility to define a shared workspace.

**There are currently some failure on the e2e tests, due to missing rest changes on the test instance; the tests expected to fail are my-dspace.cy.ts and submission.cy.ts.**

**To review the e2e test locally the linked rest PR is needed.**

## Checklist


tdonohue:
@FrancescoMolinaro : Same with this PR, please add in the description which [DSpace-CRIS differences](https://wiki.lyrasis.org/display/DSPACE/DSpace-CRIS+and+DSpace+Differences) this PR implements/includes.

pnbecker:
Is there also a backend PR for the security configuration for metadata visibility?

FrancescoMolinaro:
Hi @pnbecker yes there is a Rest PR but is not ready yet, I will link it as soon as possible.

tdonohue:
@pnbecker and @FrancescoMolinaro : It looks like the backend PR for this PR is https://github.com/DSpace/DSpace/pull/11964.  I'll update the description of this PR to make that clear

KevinVdV:
**This is by no means a full review, just issues/things we noticed when we were reviewing the REST PR (https://github.com/DSpace/DSpace/pull/11964) **

# Issue 1

In the admin edit on the metadata page the `Security level` column is always displayed even if no metadata fields have security levels configured. As security levels are by default only enabled for person entities this will confuse users.

# Issue 2
Related to issue 1, there is an empty column at the end of the admin edit for a metadata page

# Issue 3
The `person.birthDate.settings` metadata field is configured to use the security feature, but the submission form UI doesn't show the configuration (prob because it's a date element, it worked for a onebox field.

# Issue 4

The dropdown menu on the item page has been extended with an option to edit the item in the submission forms. This is useful but the naming is kind of confusing, cause it's named "Edit all the details", maybe a better naming would be "Edit (submission form)" ?

FrancescoMolinaro:
Hi @KevinVdV , many thanks for the feedback, much appreciated.

I have fixed the issue 2 and 3, now it should be possible to set the security level also on date fields and in the admin page there won't be an empty column.

Regarding the issue 1 we are following the same logic of other columns, like the language, which is always shown even if empty so I am not sure any action is necessary, I see although the difference from the other columns as this one is configurable field by field.
For the time being I will keep the column always visible as simplifies a lot the layout, but if the documentation is not enough to explain the new feature and the column is confusing we could think to refactor the table into a more dynamic layout.

For the issue 4 I agree the namig might not be as intuitive, we can decide upon a better one considering although that we have multiple edit modes based on the user role and the labels in question are the following:

  "menu.section.FULL": "Edit all the details",

  "menu.section.OWNER": "Edit",

  "menu.section.DIRECTOR": "Edit",

  "menu.section.INVESTIGATOR": "Edit",

tdonohue:
@FrancescoMolinaro : I've noticed this PR has resulted in several e2e test failures.  Are you working on cleaning those up?  (It appears they've existed for some time)

FrancescoMolinaro:
Hi @tdonohue , the e2e are failing because some of the new endpoints are missing on the test instance, in specific we miss the endpoint for security settings and the new projections added for the workspace  item.
I have fixed some of the tests by catching the error caused by this discrepancy, as those tests were not really related to the changes from this PR.

For the remaining failing tests, my-dspace.cy.ts  and submission.cy.ts , I think would be better to wait for the rest PR to be merged, as the submission is impacted by the changes in this PR and would be best to test the whole flow correctly.

In alternative I could adapt the loading mechanism of the submission to not necessarely expect security settings and to use a fallback call in case the workspace item fails to load with the newly added embed parameters and projections.

I would really appreciate to hear your thoughts on this matter and thanks for the feedback.

tdonohue:
@FrancescoMolinaro : Sorry, I wasn't aware the tests were failing simply because they depend on the REST API.  Since that's the case, I recommend noting those expected test failures in the description of this PR.  

The best solution would then be for a reviewer/tester to run the  tests *manually* with the REST API updates installed to verify they all pass locally.  Then, once both PRs are approved, we can just ensure the REST API PR is merged first, so that we can trigger a re-run of the tests here to doublecheck that they also pass in our CI process via GitHub.

FrancescoMolinaro:
Hi @tdonohue no problem at all, I forgot to highlight this problem in the description, I have now noted the failures so that is clear why there are failures, many thanks for the feedback and the suggestion.

tdonohue:
@FrancescoMolinaro : Could you quickly rebase this PR?  It has merge conflicts. Thanks!

FrancescoMolinaro:
Hi @tdonohue , conflicts have been resolved, this should be now aligned with main.

src/assets/i18n/en.json5
```diff
@@ -3743,6 +3773,14 @@
 
   "menu.section.workflow": "Administer Workflow",
 
+  "menu.section.FULL": "Edit all the details",
```
tdonohue: I think we should rename this to simply be "Edit" or "Edit Item".

I'd also recommend this be displayed *first* in the menu, directly above "Administer", as I think this should be the recommended way to edit item metadata/files.

tdonohue:
@FrancescoMolinaro : Because #5134 was just merged, this PR will need to be rebased on `main` again to resolve merge conflicts.  (In the meantime though, I can still test this an give it feedback today.  So, if it cannot be rebased immediately, no worries.)

FrancescoMolinaro:
Hi @tdonohue , this PR should be again aligned.

src/app/core/config/submission-definitions-config-data.service.ts
```diff
@@ -0,0 +1,36 @@
...
+import { FindListOptions } from '../data/find-list-options.model';
+import { PaginatedList } from '../data/paginated-list.model';
+import { RemoteData } from '../data/remote-data';
+import { RequestService } from '../data/request.service';
+import { HALEndpointService } from '../shared/hal-endpoint.service';
+import { ConfigDataService } from './config-data.service';
+import { ConfigObject } from './models/config.model';
+
+@Injectable({ providedIn: 'root' })
+export class SubmissionDefinitionsConfigDataService extends ConfigDataService {
```
tdonohue, line 35: Please add TypeDocs or code comments to describe this new data service, including the `findAll()` method below

src/app/core/submission/models/edititem-mode.model.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import { typedObject } from '../../cache/builders/build-decorators';
+import { CacheableObject } from '../../cache/cacheable-object.model';
+import { IDToUUIDSerializer } from '../../cache/id-to-uuid-serializer';
+import { HALLink } from '../../shared/hal-link.model';
+import { ResourceType } from '../../shared/resource-type';
+import { excludeFromEquals } from '../../utilities/equals.decorators';
+
+/**
+ * Describes a EditItem mode
```
tdonohue: Could we provide a bit more details in these TypeDocs to describe where/how EditItem mode is used?

src/app/core/submission/models/edititem.model.ts
```diff
@@ -0,0 +1,55 @@
...
+import { PaginatedList } from '../../data/paginated-list.model';
+import { RemoteData } from '../../data/remote-data';
+import { ResourceType } from '../../shared/resource-type';
+import { EditItemMode } from './edititem-mode.model';
+import { SubmissionObject } from './submission-object.model';
+
+/**
+ * A model class for a EditItem.
+ */
+@typedObject
```
tdonohue, line 32: Same here. Could we provide a bit more details in these TypeDocs to describe where/how EditItem is used?

src/app/core/submission/models/metadata-security-configuration.ts
```diff
@@ -0,0 +1,54 @@
...
+import { excludeFromEquals } from '../../utilities/equals.decorators';
+import { METADATA_SECURITY_TYPE } from './metadata-security-config.resource-type';
+
+interface MetadataCustomSecurityEntries {
+  [metadata: string]: number[];
+}
+/**
+ * A model class for a security configuration of metadata.
+ */
+@typedObject
```
tdonohue, line 37: Same here. Could we provide a bit more details in these TypeDocs to describe where/how these securitysettings are used?

src/app/item-page/edit-item-page/edit-metadata-security/edit-metadata-security.component.html
```diff
@@ -0,0 +1,21 @@
+@if (securityConfigLevel?.length > 1) {
+  <div class="switch-container">
+    @for (secLevel of securityLevelsMap; track secLevel.value) {
+      <div>
+        <div class="switch-opt"
+             [ngStyle]="{'background-color': securityLevel === $any(secLevel).value ? 'white' : 'rgb(204,204,204)' }">
+          <button class="btn btn-link text-primary" data-test="switch-btn"
+                  title="Metadata security level {{ $any(secLevel).value }}"
```
tdonohue: Should this `title` be translatable?

src/app/item-page/edit-item-page/edit-metadata-security/edit-metadata-security.component.html
```diff
@@ -0,0 +1,21 @@
...
+                  [dsBtnDisabled]="readOnly"
+                  (click)="changeSelectedSecurityLevel($any(secLevel).value)">
+            <i [class]="$any(secLevel).icon"
```
tdonohue, line 11: We may want to add `aria-hidden="true"` to this icon so that it's ignored from accessibility tools.

src/app/item-page/edit-item-page/edit-metadata-security/edit-metadata-security.component.ts
```diff
@@ -0,0 +1,106 @@
...
+@Component({
+  selector: 'ds-edit-metadata-security',
+  templateUrl: './edit-metadata-security.component.html',
+  styleUrls: ['./edit-metadata-security.component.scss'],
+  imports: [
+    BtnDisabledDirective,
+    NgStyle,
+  ],
+})
+export class EditMetadataSecurityComponent implements OnInit {
```
tdonohue, line 54: Please add a Typedocs description for this component.

src/app/shared/form/builder/ds-dynamic-form-ui/models/relation-group/modal/dynamic-relation-group-modal.components.ts
```diff
@@ -439,6 +467,33 @@ export class DsDynamicRelationGroupModalComponent extends DynamicFormControlComp
     return item;
   }
 
+  private initSecurityLevelConfig(chipModel: DynamicInputModel, modelGroup: DynamicFormGroupModel) {
```
tdonohue, line 493: Could we add a basic TypeDoc comment to this larger method to describe how it's initializing securityLevel

src/app/shared/form/builder/ds-dynamic-form-ui/models/relation-inline-group/dynamic-relation-inline-group.components.ts
```diff
@@ -166,8 +184,60 @@ export class DsDynamicRelationInlineGroupComponent extends DynamicFormControlCom
...
+          const modelRow = groupRow as DynamicFormGroupModel;
+          modelRow.group.forEach((model: DynamicInputModel) => {
+            this.initSecurityLevelConfig(model, modelRow);
+          });
+        });
+      });
+    });
+  }
+
+  private initSecurityLevelConfig(model: any, modelGroup: DynamicFormGroupModel) {
```
tdonohue, line 221: Same here. This is a larger method that likely should have TypeDocs just to describe how it's initializing the settings

src/app/shared/menu/providers/edit-item-details.menu.ts
```diff
@@ -0,0 +1,61 @@
...
+import { getEditItemPageRoute } from '../../../app-routing-paths';
+import { MenuItemType } from '../menu-item-type.model';
+import { PartialMenuSection } from '../menu-provider.model';
+import { DSpaceObjectPageMenuProvider } from './helper-providers/dso.menu';
+
+/**
+ * Menu provider to create the "Edit details" option in the DSO audit menu
+ */
+@Injectable()
+export class EditItemDetailsMenuProvider extends DSpaceObjectPageMenuProvider {
```
tdonohue: If we change the name of the menu option (like I suggested), we could rename this to be `EditItemMenuProvider` (or similar).  

(This change isn't required, but I just noticed that this provider is named the same as the new menu option)

src/app/submission/edit/pending-changes/pending-changes.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+  Observable,
+  of,
+} from 'rxjs';
+import { switchMap } from 'rxjs/operators';
+
+import { ConfirmationModalComponent } from '../../../shared/confirmation-modal/confirmation-modal.component';
+import { SubmissionEditCanDeactivateService } from '../submission-edit-can-deactivate.service';
+import { ThemedSubmissionEditComponent } from '../themed-submission-edit.component';
+
+export const pendingChangesGuard: CanDeactivateFn<ThemedSubmissionEditComponent> = (
```
tdonohue, line 42: Please add a TypeDocs style description to this guard.

src/app/submission/edit/submission-edit-can-deactivate.service.ts
```diff
@@ -0,0 +1,32 @@
...
+import { SubmissionService } from '../submission.service';
+
+interface PendingChangesGuardComponentInterface {
+  canDeactivate: (id: string) => boolean | Observable<boolean>;
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SubmissionEditCanDeactivateService implements PendingChangesGuardComponentInterface {
```
tdonohue, line 42: Please describe this service with a TypeDocs comment

src/app/submission/objects/submission-objects.actions.ts
```diff
@@ -212,6 +219,46 @@ export class DisableSectionAction implements Action {
   }
 }
 
+export class DisableSectionSuccessAction implements Action {
```
tdonohue, line 222: Can we describe this new action with a TypeDocs style comment?

src/app/submission/objects/submission-objects.actions.ts
```diff
@@ -212,6 +219,46 @@ export class DisableSectionAction implements Action {
...
+   *    the submission's ID to remove
+   * @param sectionId
+   *    the section's ID to remove
+   */
+  constructor(submissionId: string, sectionId: string) {
+    this.payload = { submissionId, sectionId };
+  }
+}
+
+export class DisableSectionErrorAction implements Action {
```
tdonohue, line 242: Same here. Please describe this new action with a comment

src/app/submission/objects/submission-objects.actions.ts
```diff
@@ -268,6 +315,35 @@ export class CleanDuplicateDetectionAction implements Action {
   }
 }
 
+export class UpdateSectionErrorsAction implements Action {
```
tdonohue, line 318: Same here. Please describe this action & where it's used

src/app/submission/objects/submission-objects.effects.ts
```diff
@@ -262,9 +294,28 @@ export class SubmissionObjectEffects {
...
+        catchError((rd: unknown) => {
+          if (rd instanceof RemoteData) {
+            return observableFrom(
+              this.parseErrorResponse(false, rd.errors, action.payload.submissionId, rd.statusCode, rd.errorMessage),
+            );
+          }
+        }));
     })));
 
+  removeSection$ = createEffect(() => this.actions$.pipe(
```
tdonohue, line 313: Please add a comment to describe this new effect, similar to all other effects in this file

src/app/submission/sections/form/section-form-operations.service.ts
```diff
@@ -543,6 +567,36 @@ export class SectionFormOperationsService {
     } else if (previousValue.isPathEqual(this.formBuilder.getPath(event.model))) {
       this.operationsBuilder.remove(pathCombiner.getPath(segmentedPath));
     }
+  }
 
+  protected changeSecurityLevel(pathCombiner: JsonPatchOperationPathCombiner,
```
tdonohue, line 601: As this is a larger method, please add TypeDocs to describe its behavior

src/assets/i18n/en.json5
```diff
@@ -3832,6 +3866,8 @@
 
   "mydspace.show.workspace": "Your submissions",
 
+  "mydspace.show.otherworkspace": "Other Workspace Submissions",
```
tdonohue: Should we consider changing this text to be "Shared Submissions"?  It seems like that might be a name that will make more sense to users on the MyDSpace page.

src/config/metadata-security-config.ts
```diff
@@ -0,0 +1,11 @@
+import { Config } from './config.interface';
+
+export interface MetadataSecurityConfig extends Config {
```
tdonohue, line 12: Please add a basic description to both of these interfaces.  It just helps make it clear *which* configs they reference

src/config/default-app-config.ts
```diff
@@ -823,6 +824,27 @@ export class DefaultAppConfig implements AppConfig {
...
   };
 
+
+  security: MetadataSecurityConfig = {
+    levels: [
+      {
+        value: 0,
+        icon: 'fa fa-globe',
+        color: 'green',
+      },
```
tdonohue: These configurations and their default values should be added to `config.example.yml`.  We also should provide a description there for how the configuration is used and what valid values exist.

Could we also consider moving this configuration under the existing `item` configurations?  Currently, these `security` settings seem item specific.  So, for example, they might appear like this in `config.example.yml`:
```
item:
  security:
    levels:
```
Or even under `item > edit`:
```
item:
  edit:
    security:
      levels:
```

Either of those approaches would be fine with me. I just feel this shouldn't be a top-level setting because these `security` settings are not applicable on all types of objects.

tdonohue (changes requested):
@FrancescoMolinaro : Thanks for this PR!  I gave it a code review today.  Mostly the code looks great, but I've found some minor things that I'd recommend addressing (mostly additional code comments / TypeDocs). 

If any of these are too complex to address now, we can always create a follow-up ticket to address them later. However, I'm hopeful this will not be too complex to cleanup (even though I did add 21 inline comments below).

I have **not** yet tested this feature. But, I plan to still do so today.  I'll submit any testing feedback as a separate comment/review.

src/assets/i18n/en.json5
tdonohue: I noticed just now that @KevinVdV had a similar comment about this menu option.  I feel "Edit all the details" is misleading here because the "Administer" menu offers **more details** (e.g. permissions, etc) which you can edit.  So, I'd rather this menu option just say "Edit" or I'm OK with something like "Edit via form" or "Edit metadata and files"

tdonohue (changes requested):
@FrancescoMolinaro and @AdamF42 : I've finished my initial testing of this feature.  I'm just submitting my results in this frontend review even though I know some of the fixes will require backend changes.  Overall, I've found this works well!  

But, I have found some bugs / possible issues.  If any of these are too complex to fix now, then we could create follow-up tickets for the effort:

1. First, a minor thing. The "Edit all the details" menu option should have an icon next to it.  As you can see in the screenshot in the description of this PR, it's missing an icon.  I'd recommend using the *pencil* for this Edit menu, and perhaps change the "Administer" menu option to use either a [wrench](https://fontawesome.com/icons/wrench?f=classic&s=solid) or [gear](https://fontawesome.com/icons/gear?f=classic&s=solid) icon (I think I have a slight preference for the wrench)
2. The "Edit all the details" menu does *not* appear for traditional Items (i.e. Items which are *not* Entities).  Is there a way we could support traditional items as well?  Or is this only possible to do for Entities?
3. When using the "Edit all the details" menu to edit an existing Item, if I upload a new file and click "Save and return", then I return to the Item page **but the new file does not appear on that page**. If I reload the page, then the new file appears.  It appears there's a caching issue.
4. If I click on "Edit all the details" menu for a Project or a Person, then the page **hangs indefinitely**.  Behind the scenes, I see there's an error in the backend log which says the submission-process named `project-edit` or `person-edit` doesn't exist.  Could we add all the `-edit` forms (on the backend) for all default entities to avoid this issue?
    * Additionally, if I create an `person-edit` form, I still get an error. But it says I'm now missing an `admin-person-edit` form (maybe because I'm logged in as an Admin?)
    * Alternatively, we could see if there's a way to **avoid displaying** the "Edit all the details" menu if this submission-process configuration is missing.  Or maybe show the menu option but disable it (to make it clear it's not configured).
5. When creating/editing a Collection, I verified the "Submission definition" option works well. However, I wonder if we should find a way to **hide** all the `*-edit` forms because these forms are only used for **editing objects**. 
    * For example, I selected the `publication-edit` definition for a collection and I noticed that the License section is then missing from the submission form for new items.  So, it might be best to see if we can avoid people selecting these `*-edit` forms as they are not meant for new submissions.
6. Finally, I tested the "Shared Workspace" options at the Collection level.  They work well overall.  However, I'm confused by the "Other Workspace Submissions" view on my MyDSpace.  I only have a **single** Collection which has "Shared Workspace" enabled, but I see Items under "Other Workspace Submissions" that belong to **other collections**.  Is this expected behavior?  I expected that I'd only see Items in that menu which belong to a Collection having a "Shared Workspace".
    * I tested this by creating one Collection with a "Shared Workspace" as an Admin
    * Started a new submission to that collection (again as an Admin). I didn't complete it.
    * Logged out and logged in as a different user who only has *Submit* privileges to that Collection (This user is a Collection Admin in other collections but *not* this new Collection)
    * When I visit that user's MyDSpace, I see the following:
        * On the "Your Submissions" menu, I see only **one** item still in the "Workspace"
        * If I change to the "Other Workspace Submissions", then I see 86 items which are still in the "Workspace".  **I'm not sure why I'm seeing so many.**
        * However, I was able to verify that this user can edit/update the shared submission that was started from my Admin account.

Other than the bugs I listed above, other features appear to be working! (I've also done some initial basic testing of the metadata security settings, and they appear to be working well so far)

src/app/submission/objects/submission-objects.actions.ts
FrancescoMolinaro: Hi @tdonohue , thanks for the feedback, the class has already a brief description before the constructor, like all the other actions in the file. I believe is ok as is since these are only describing the type of payload of the action, but I will add a description of the effect that uses this action in submission-objects.effects.ts. Please let me know if that's ok in your opinion; I would apply the same approach to all the other actions.

src/app/submission/form/footer/submission-form-footer.component.html
```diff
@@ -14,6 +14,11 @@
       }
     </div>
     <div class="text-end d-flex justify-content-end align-items-center flex-sm-row flex-column">
+      @if ((hasUnsavedModification | async) !== true && (processingSaveStatus | async) !== true && (processingDepositStatus | async) !== true && (submissionIsInvalid | async)) {
+        <span>
+          <i class="fas fa-warning"></i> {{'submission.general.info.invalid' | translate}}
+        </span>
+      }
```
alexandrevryghem, line 31: There is a spacing issue between the warning message and the other parts for the footer

Image

src/app/collection-page/collection-form/collection-form.component.ts
```diff
@@ -117,35 +136,65 @@ export class CollectionFormComponent extends ComColFormComponent<Collection> imp
...
+    const definitions$: Observable<ConfigObject[]> = this.submissionDefinitionService
+      .findAll({ elementsPerPage: 100, currentPage: 1 }).pipe(
+        getFirstSucceededRemoteListPayload(),
+        catchError(() => of([])),
+      );
 
-      this.formModel = entityTypes.length === 0 ? collectionFormModels : [...collectionFormModels, this.entityTypeSelection];
+    // retrieve all entity types and submission definitions to populate the dropdowns selection
+    combineLatest([entities$, definitions$])
+      .subscribe(([entityTypes, definitions]: [ItemType[], SubmissionDefinitionModel[]]) => {
```
alexandrevryghem: If one of these observables doesn't succeed, this will not emit and this will cause a memory leak

src/app/shared/form/builder/parsers/field-parser.ts
```diff
@@ -43,6 +46,7 @@ export const PARSER_OPTIONS: InjectionToken<ParserOptions> = new InjectionToken<
  * The regex itself is encapsulated inside a `RegExp` object, that will validate the pattern syntax.
  */
 export const REGEX_FIELD_VALIDATOR = new RegExp('(\\/?)(.+)\\1([gimsuy]*)', 'i');
+export const SECURITY_CONFIG: InjectionToken<any> = new InjectionToken<any>('securityConfig');
```
alexandrevryghem: Based on its usages the type of this token is `MetadataSecurityConfiguration`

src/app/shared/form/builder/parsers/field-parser.ts
```diff
@@ -58,6 +62,7 @@ export abstract class FieldParser {
     @Inject(CONFIG_DATA) protected configData: FormFieldModel,
     @Inject(INIT_FORM_VALUES) protected initFormValues: any,
     @Inject(PARSER_OPTIONS) protected parserOptions: ParserOptions,
+    @Inject(SECURITY_CONFIG) protected securityConfig: any = null,
```
alexandrevryghem: The type should be replaced here as well (with `MetadataSecurityConfiguration`). Could you go over all the usages and fix it for all of them?

src/app/shared/form/builder/parsers/field-parser.ts
```diff
@@ -426,5 +452,23 @@ export abstract class FieldParser {
       });
     }
   }
-
+  mapBetweenMetadataRowAndSecurityMetadataLevels( metadata: string): any {
+    // look to find security for metadata
+    if (this.securityConfig && metadata) {
+      if (this.securityConfig.metadataCustomSecurity) {
+        const metadataConfig = (this.securityConfig.metadataCustomSecurity as any)[metadata];
```
alexandrevryghem: This also remove the need for this cast to type any for example

src/app/shared/form/builder/parsers/concat-field-parser.ts
```diff
@@ -35,11 +36,12 @@ export class ConcatFieldParser extends FieldParser {
...
     @Inject(INIT_FORM_VALUES) initFormValues,
     @Inject(PARSER_OPTIONS) parserOptions: ParserOptions,
-      translate: TranslateService,
+    @Inject(SECURITY_CONFIG) securityConfig: any,
+    protected translate: TranslateService,
     protected separator: string,
     protected firstPlaceholder: string = null,
     protected secondPlaceholder: string = null) {
-    super(submissionId, configData, initFormValues, parserOptions, translate);
+    super(submissionId, configData, initFormValues, parserOptions, translate, securityConfig);
```
alexandrevryghem: The order of the `translate` & `securityConfig` should be switched here to match the order of the constructor in the parent class

FrancescoMolinaro:
Hi @tdonohue , thanks for the feedback, I have addressed the change requests for the type docs on the frontend and for the issues you reported.
Of the 6 point you listed there are some that still require some attention, in particular numbers 2,4,6.

2 - We would need to improve the handling of the entity type "none" on the edit modes endpoint, Adamo will get back on this as soon as possible.

4 - All the necessary *-edit forms have been added but we still need to remove the admin-person-edit since that is a configuration customization we have on CRIS but not necessary on DSpace as we don't have a distinguish between Person form and Admin form .

6 - We are currently investigating the issue and we will come back as soon as possible.


For the rest the issues you noted should have been addressed and should be ready to be retested.
Again many thanks for the review.

alexandrevryghem (changes requested):
Thnx @FrancescoMolinaro for this PR I tested all the functionalities and experienced the same issue as Tim for the Shared workspace functionality. For the other functionalities I experienced the following issues

1. The old DSO edit metadata view doesn’t work correctly anymore, currently when you select metadata field `dspace.entity.type` or `dc.subject` a regular text area is shown. Instead it should render the `DsoEditMetadataEntityFieldComponent`/`DsoEditMetadataAuthorityFieldComponent` based on the metadata value
2. Currently when you click on the "Edit" menu, there is no clean way to return back to the item page without saving your changes. I would maybe add a "Cancel"/"Back" button similar to the "Discard" button on the regular submission forms
3. The edit submission forms don't display the breadcrumbs
4. While testing the submission definition form I did have an issue uploading bitstreams after I removed a submission definition that was still assigned to a collection. Restoring the delete submission form did fix it, but I wasn't able to reproduce it anymore when I tried to reproduce this bug again, so this can probably be discarded

I also took a quick look at the code and added some inline comments

FrancescoMolinaro:
Hi @alexandrevryghem , thanks for the review, much appreciated. 
I have fixed the issues you reported but I couldn't reproduce the issue number 4 as well, it could have been a once time thing but in case should happens again I believe we should have a look to the rest errors and note the step to reproduce the issue, as I believe there could have been an error due to the missing submission definition.

This should be anyway ready for a second look, many thanks again.

src/app/submission/objects/submission-objects.actions.ts
tdonohue: Hi @FrancescoMolinaro : I think you may have misunderstood my feedback here.  I was asking for you to add a comment here to describe the new `DisableSectionSuccessAction` action that is created here.  Currently, there's no comment that describe where this action is used or what it does.

I had that same comment for the other new Actions in this same file: `DisableSectionErrorAction` and `UpdateSectionErrorsAction`.  Each of these new actions should have a basic comment that describes their purpose.

tdonohue:
@FrancescoMolinaro : Thanks for the changes you've made.  I've re-reviewed the code changes, and you've addressed everything except my comments in `submission-objects.actions.ts`, so I [added a clarifying comment](https://github.com/DSpace/dspace-angular/pull/5145#discussion_r3012315349).

As for the [previously reported bugs/issues](https://github.com/DSpace/dspace-angular/pull/5145#pullrequestreview-4023451071):

> 1. First, a minor thing. The "Edit all the details" menu option...

This was **fixed**. Thanks!

> 2. The "Edit all the details" menu does _not_ appear for traditional Items (i.e. Items which are _not_ Entities).  Is there a way we could support traditional items as well?  Or is this only possible to do for Entities?

As you said, you are working on this. _Not yet fixed._

> 3. When using the "Edit all the details" menu to edit an existing Item, if I upload a new file and click "Save and return", then I return to the Item page **but the new file does not appear on that page**. If I reload the page, then the new file appears.  It appears there's a caching issue.

This is **not yet fixed**. The issues still occurs. If you edit an archived item & add a new file, you have to reload the page for the new file to appear.

> 4. If I click on "Edit all the details" menu for a Project or a Person, then the page **hangs indefinitely**.  Behind the scenes, I see there's an error in the backend log which says the submission-process named `project-edit` or `person-edit` doesn't exist.  Could we add all the `-edit` forms (on the backend) for all default entities to avoid this issue?

As you said, you are working on this. _Not yet fixed._

> 5. When creating/editing a Collection, I verified the "Submission definition" option works well. However, I wonder if we should find a way to **hide** all the `*-edit` forms because these forms are only used for **editing objects**.

This was **fixed**. Thanks!

> 6. Finally, I tested the "Shared Workspace" options at the Collection level.  They work well overall.  However, I'm confused by the "Other Workspace Submissions" view on my MyDSpace.  I only have a **single** Collection which has "Shared Workspace" enabled, but I see Items under "Other Workspace Submissions" that belong to **other collections**.  Is this expected behavior?  I expected that I'd only see Items in that menu which belong to a Collection having a "Shared Workspace".

As you said, you are working on this. _Not yet fixed._

For any of the ones that are *not yet fixed*, I'd be OK with logging a bug ticket to address in a follow-up PR. Overall, these features are working well.

src/app/shared/form/builder/parsers/concat-field-parser.ts
```diff
@@ -35,11 +36,12 @@ export class ConcatFieldParser extends FieldParser {
     @Inject(CONFIG_DATA) configData: FormFieldModel,
     @Inject(INIT_FORM_VALUES) initFormValues,
     @Inject(PARSER_OPTIONS) parserOptions: ParserOptions,
-      translate: TranslateService,
+    @Inject(SECURITY_CONFIG) securityConfig: any,
```
alexandrevryghem: Could you replace this type with `MetadataSecurityConfiguration` as well

src/app/shared/form/builder/parsers/dropdown-field-parser.ts
```diff
@@ -25,9 +26,10 @@ export class DropdownFieldParser extends FieldParser {
     @Inject(CONFIG_DATA) configData: FormFieldModel,
     @Inject(INIT_FORM_VALUES) initFormValues,
     @Inject(PARSER_OPTIONS) parserOptions: ParserOptions,
+    @Inject(SECURITY_CONFIG)  securityConfig: any = null,
```
alexandrevryghem: Could you replace this type with `MetadataSecurityConfiguration` as well

src/app/shared/form/builder/parsers/name-field-parser.ts
```diff
@@ -18,8 +19,9 @@ export class NameFieldParser extends ConcatFieldParser {
     @Inject(CONFIG_DATA) configData: FormFieldModel,
     @Inject(INIT_FORM_VALUES) initFormValues,
     @Inject(PARSER_OPTIONS) parserOptions: ParserOptions,
+    @Inject(SECURITY_CONFIG) securityConfig: any = null,
```
alexandrevryghem: Could you replace this type with `MetadataSecurityConfiguration` as well

src/app/shared/form/builder/parsers/series-field-parser.ts
```diff
@@ -18,8 +19,9 @@ export class SeriesFieldParser extends ConcatFieldParser {
     @Inject(CONFIG_DATA) configData: FormFieldModel,
     @Inject(INIT_FORM_VALUES) initFormValues,
     @Inject(PARSER_OPTIONS) parserOptions: ParserOptions,
+    @Inject(SECURITY_CONFIG) securityConfig: any = null,
```
alexandrevryghem: Could you replace this type with `MetadataSecurityConfiguration` as well

src/app/collection-page/collection-form/collection-form.component.ts
```diff
@@ -117,19 +136,33 @@ export class CollectionFormComponent extends ComColFormComponent<Collection> imp
...
+      .findAll({ elementsPerPage: 100, currentPage: 1 }).pipe(
+        getFirstSucceededRemoteListPayload(),
+        catchError(() => of([])),
+      );
+
+    // retrieve all entity types and submission definitions to populate the dropdowns selection
+    forkJoin({
+      entityTypes: entities$,
+      definitions: definitions$,
+    }).subscribe(({ entityTypes, definitions }:  {entityTypes: ItemType[]; definitions: ConfigObject[]}) => {
```
alexandrevryghem, line 181: This subscription might still leak if either `entities$` or `definitions$` doesn't emit (for example if the request fails). To fix this we need to save the `Subscription` in a variable and unsubscribe from that subscription in the `ngOnDestroy`

alexandrevryghem (changes requested):
Thnx @FrancescoMolinaro! I've reviewed the code again and added some small inline comments. I also verified that the following issues were fixed: 
1. This is fixed thnx!
2. This is fixed thnx!
3. For the breadcrumbs the (sub-)communities & collections are currently not displayed (similar to the regular submission forms), but this could be fixed in a later PR as well

FrancescoMolinaro:
Hi @alexandrevryghem , thanks again for the review, I removed the any interface in favor of the correct one and added a new breadcrumb resolver for the edit mode so that we get the same behavior as the normal submission.

In addition during the tests we realized we were missing a change done in CRIS for the sections visibility, which resulted in an error loading the forms if the collection step is not present in the *-edit configuration in item-submission.xml.
I have ported also these changes and adapted the tests and mocks accordingly, this change is mainly moving the checks for visibility in utility functions, simplifying the components that used to build internally the logic for the visibility; the core functionality should be unchanged and the missing collection step should be handled properly.

tdonohue:
@FrancescoMolinaro : Could you please rebase this on `main`?  We've just merged #5106 which resulted in merge conflicts on this PR.  Thanks!

src/app/core/config/models/config-submission-section.model.ts
```diff
@@ -31,6 +43,12 @@ export class SubmissionSectionModel extends ConfigObject {
   @autoserialize
   mandatory: boolean;
 
+  /**
+   * A boolean representing if this submission section is opened or collapsed by default
+   */
+  @autoserialize
+  opened: boolean;
```
alexandrevryghem: This property should be removed, because it doesn't exist in the backend on the `SubmissionSectionRest`

FrancescoMolinaro:
Hi @tdonohue , thanks for the notice, it should be now aligned with main.

src/app/core/config/models/config-submission-section.model.ts
```diff
@@ -6,14 +6,26 @@ import {
...
+  ReadOnly = 'read-only',
+  Hidden = 'hidden'
+}
+
+/**
+ * An interface that define section visibility and its properties.
+ */
+export interface SubmissionVisibilityType {
+  [scope: string]: SubmissionVisibilityValue;
+}
```
alexandrevryghem: I assume the reason why you chose to replace the `SectionVisibility` with this new `SubmissionVisibilityType` is that this would match the backend rest objects. I would however move this interface to a separate file to prevent any future circular dependencies. I would also hardcode the scopes instead of allowing all keys like this:

```ts
export interface SubmissionVisibilityType {
  main: SubmissionVisibilityValue;
  other: SubmissionVisibilityValue;
}
```

If you do which to be able to support different scopes at field/section level we could however use something like this, but this way we don't simply allow any strings:

```ts
export type SectionVisibilityType = 'main' | 'other';

export type SubmissionVisibilityType<T extends string> = {
  [scope in T]: SubmissionVisibilityValue;
};

@autoserialize
visibility: SubmissionVisibilityType<SectionVisibilityType>;
```

I would also remove the `section-visibility.model.ts#SectionVisibility` to avoid future confusion

alexandrevryghem (changes requested):
Thnx @FrancescoMolinaro all my prior feedback has been addressed but I did find some new bugs and have a concern about the most recent code changes you pushed in https://github.com/DSpace/dspace-angular/pull/5145/changes/a84e2d313fee95ce97ace48fc95e327d91899221:
- Currently if I try to move the item to a different collection, an error 500 is thrown in the backend. I'm also not sure if we should offer this functionality through the edit item submission forms, or if we should make the edit collection step read only
  ```
  org.dspace.app.rest.exception.DSpaceBadRequestException: The submission step class specified by 'org.dspace.app.rest.submit.step.CollectionStep' does not extend the class org.dspace.submit.AbstractProcessingStep! Therefore it cannot be used by the Configurable Submission as the <processing-class>!
  ```
- The linked backend PR also added a new submission section, but the translation for this step is still missing:
  Image
- When you switch between collections during a regular submission the old collection name is displayed again (I remember this being a bug at some point in time, but it got fixed and I also can't reproduce it anymore on [sandbox.dspace.org](https://sandbox.dspace.org))
  <video src="https://github.com/user-attachments/assets/cfbdb572-49c0-45c3-817a-8e260bb07461"/>
- A lot of places where `scope: SectionScope` was used were replaced by `opened: boolean` this looks like a mistake could you double check? I haven't had time to test scoped fields those changes look like they would break this feature

tdonohue (approved):
👍 Thanks @FrancescoMolinaro !  I've re-reviewed and re-tested today.  I'm a +1 for this feature, but there are a four outstanding bugs that haven't yet been fixed (including a few new ones I've discovered today).

That said, I feel **all of these bugs can be moved to follow-up tickets** and we could merge this PR as-is.  That gives the outstanding bugs more time to be analyzed and addressed during/after Testathon.

Here's the outstanding bugs that I'm still seeing:
1. (Previously reported) When using the new "Edit" feature and modifying a Publication via the submission form, if you upload a *new file* and save the form, then that file will not appear on the Publication until you reload the page.  This appears to be a caching issue.  (This one I'd create a regular bug ticket for)
2. (Previously reported) The "Shared Submissions" feature of the MyDSpace is showing **many more** shared submissions than I expect, including some from collections that do NOT have the "Shared Workspace" checkbox enabled. (This one I'd create a "blocker" ticket for, as it needs solving before 10.0)
3. (NEW) While the new "Edit" feature now *works* for traditional Item objects, it's not correctly handling traditional Authority Control values.  For instance, if I edit a traditional Item that has an Author of "Tim Donohue" which has an ID in the authority column, then the Edit form will display the Authority ID and NOT the value. Here's a screenshot of what I mean. This particular Item has an Author named "Donohue, Tim" which has an Authority ID. You see the Authority ID and not the value in the Author field when editing the Item. (This behavior ONLY occurs for traditional Items and not for Publications) Image
4. (NEW) If you use the new "Edit" feature to edit an Entity that has **no bitstreams**, and then upload a bitstream, it is uploading **twice**! 
    * To reproduce it's easiest to find a Person Entity as many of these won't have a bitstream attached. 
    * Click the "Edit" menu on the Person to bring up the Edit Form
    * Verify it has no Bitstreams uploaded. Then, upload a **single** bitstream
    * After upload, two copies of that same bitstream will be attached to the Entity.

Those are the only outstanding issues I've found.  All my other prior feedback has been addressed/fixed.  As noted, I think all of these outstanding issues can be moved to follow-up tickets to be addressed later.

tdonohue:
@alexandrevryghem : Thanks for all your ongoing reviews here!  

Please let us know whether any of your feedback can be moved to a follow-up ticket.  We can even create `blocker` tickets if there are issues you feel *must* be addressed prior to 10.0.  The reason I'm pointing this out is that we have Testathon **next week** and we really need to get this PR merged by end-of-day on Weds (April 1) if we are going to include it in 10.0 and  Testathon. (As with past major releases, we will not be accepting any new features to 10.0 once Testathon starts)

As you can see from my review above, I've gone ahead and approved this PR despite still finding a few bugs (and I'll create separate bug tickets for those outstanding issues).  But, I feel confident we can get those bugs fixed prior to 10.0 (and likely very early on in Testathon).  If you are in a similar situation, please feel free to create follow-up bug tickets which we'll ensure get addressed.

alexandrevryghem:
@tdonohue: Opening separate tickets for the remaining issues is fine by me. However, if @FrancescoMolinaro still has some time today, I would maybe try to still remove that new `opened` property and to restore the `scope` field on the submission section classes, as well as throughout the codebase, to keep it consistent with the backend code.

It looks like this change may be part of a separate feature that was accidentally partially cherry-picked: https://github.com/4Science/dspace-angular/commit/28ab2b4d6e7990346a33a8e9298dd2ea18261d7b

FrancescoMolinaro:
Hi @tdonohue , @alexandrevryghem , sorry for the issues on the last commit, I did indeed a mistake on porting some of the code from CRIS.
I have removed the 'opened' property and restored the scope everywhere, refactored the visibility models and fixed the issue with the collection dropdown, added the missing label and also the caching issue for the bitstreams should be resolved.

Regarding:

> Currently if I try to move the item to a different collection, an error 500 is thrown in the backend. I'm also not sure if we should offer this functionality through the edit item submission forms, or if we should make the edit collection step read only

the collection step should not be configured in the *-edit forms, similarly to the license step, this should have been addressed on the rest PR, making the collection effectively read only on the submission in edit mode.
 
For this other issue:

> (NEW) While the new "Edit" feature now works for traditional Item objects, it's not correctly handling traditional Authority Control values. For instance, if I edit a traditional Item that has an Author of "Tim Donohue" which has an ID in the authority column, then the Edit form will display the Authority ID and NOT the value. Here's a screenshot of what I mean. This particular Item has an Author named "Donohue, Tim" which has an Authority ID. You see the Authority ID and not the value in the Author field when editing the Item. (This behavior ONLY occurs for traditional Items and not for Publications)

I have tried reproducing like this:

1 - create a standard Item with author (not using the new authority framework)
2 - Add uuid into authority of the author field via administrative table (standard admin edit)
3 - open the item in edit mode -> I can properly see the author name in the field

@tdonohue  following these steps I wasn't able to reproduce the issue but most likely I am missing something, could you please share, also directly in the dedicated bug issue, the steps you did to reproduce the problem?  Thanks in advance.

For the remaning open points I did some debugging and I believe we have some issues on the rest side, in specific:

1) double upload on item with upload section not mandatory -> On CRIS the upload is always mandatory in the form so I believe we might have overlooked this scenario, I see anyway a double bitstream being added with a single binary being sent to the rest.

2) Wrong search results for shared workspace -> I think there is a wrong query being executed with the new discovery config, we should fix the search logic for this use case.

Thanks again for the reviews and sorry for the delay on this.

alexandrevryghem:
Thanks @FrancescoMolinaro! I’ve reviewed the latest code changes and can confirm that the changes in question were properly reverted. I’m now also +1, and I can confirm that with the latest REST changes, it’s no longer possible to modify the collection via the edit forms

src/app/core/submission/models/section-visibility.model.ts
```diff
@@ -1,11 +1,23 @@
...
-  other: any;
+export enum SubmissionVisibilityValue {
+  ReadOnly = 'read-only',
+  Hidden = 'hidden'
 }
 
+/**
+ * Types of possible section's visibility
+ */
+export type SectionVisibilityType = 'main' | 'other' | 'submission' | 'workspace' | 'workflow';
```
alexandrevryghem, line 12: I'm not sure when the backend would use the types `submission`, `workspace` & `workflow`. Based on what I've seen in the backend both the `SubmissionFormFieldRest` & the `SubmissionSectionRest` use the `SubmissionVisibilityRest` object to represent the visibility configuration and this only has 2 properties `main` & `other`. But this is a small detail that could be fixed later.

On DSpace CRIS the `main` & `other` types were indeed replaced with `submission`, `workspace` & `workflow` [at some point](https://github.com/4Science/DSpace/commit/ae7981b61cb3d8ad14d1a8c5c1fb7451418278eb#diff-fbc8cbdc36cb5a7e94b4fdec4c607a1f86d5e3c7444016e6a09a2829b95b0bd8) so this is probably where this originates from

alexandrevryghem (approved):

AdamF42:
> * double upload on item with upload section not mandatory -> On CRIS the upload is always mandatory in the form so I believe we might have overlooked this scenario, I see anyway a double bitstream being added with a single binary being sent to the rest.
> 
> * Wrong search results for shared workspace -> I think there is a wrong query being executed with the new discovery config, we should fix the search logic for this use case.

For the remaning open points : 

- The double upload problem should be fixed.

- Wrong search results for shared workspace -> We can create a follow up task to handle this.

tdonohue:
NOTE: I've done final testing today and verified the "double upload" problem is fixed.  The backend PR was just merged.  Once the fresh Docker images for the backend are created (may take 30-60mins), I'm going to restart the tests in this PR to verify all e2e tests now pass.

Assuming the e2e tests pass, this PR will also be merged as it's already at +2.  Any outstanding issues (including those I reported above) should be moved to follow-up tickets so they can be addressed in 10.0 (or later if necessary)

tdonohue:
@FrancescoMolinaro : After merging the backend PR, all e2e tests except *one* (`item-statistics.cy.ts`) are passing now in this PR.  I'm trying to figure out why that one test is failing...it could just be minor updates are needed.

tdonohue:
@FrancescoMolinaro : After more analysis of the `item-statistics.cy.ts` failures:
* I can reproduce this failure locally if I run `npm run e2e` and run that test via Cypress.  
* It appears the "Statistics" menu (in the header) is not loading on Item pages in this Cypress test environment.  The reason is unclear to me, but I've also locally modified the timeout to be 20 seconds, and the "Statistics" menu **never loads** (so it's not a  timeout issue in the test)
* Strangely, when I test this **outside of Cypress**, the "Statistics" menu does appear on Item/Entity pages.  So, it seems to be only an issue within the Cypress environment.
* I'm not sure why this PR causes this failure, but this failure doesn't exist on current `main`.  My guess is that something in this PR is causing this failure (indirectly), but I'm not sure what it is right now.

In order to still move this PR along quickly, I'd propose we **temporarily disable the failing test** and create a follow-up ticket to fix this issue.  The specific test to disable is the [first one in that file](https://github.com/DSpace/dspace-angular/blob/main/cypress/e2e/item-statistics.cy.ts#L7), and we can disable it by changing the `it` to `xit` (to skip that test).

tdonohue:
@FrancescoMolinaro : A follow-up to my last message.  This PR **definitely** is causing the error in `item-statistics.cy.ts`.  

I've realized that if I load any Publication Entity page while running the UI in *Dev Mode* then the "Statistics" menu never appears in the header.  Here's how to reproduce:
1. Run the UI in Dev Mode
2. Visit any Publication entity page.  If you browse/search to get to the page, the "Statistics" menu appears in the header.
3. Reload the Publication page in your browser.  The "Statistics" menu will disappear.  In other words, a direct link to the publication page will NOT load the "Statistics" menu.

I'm still OK with temporarily disabling this e2e test unless it's obvious how to fix this bug.  We can move that e2e test to a folllow-up ticket, if it's not able to be fixed in this PR.

src/assets/i18n/en.json5
```diff
@@ -3832,6 +3868,8 @@
 
   "mydspace.show.workspace": "Your submissions",
 
+  "mydspace.show.otherworkspace": "Shared Submissions",
```
tdonohue: In re-testing today, it appears this key should be updated to be `mydspace.show.otherWorkspace`, as the key `mydspace.show.otherworkspace` is no longer working.  (The difference is that "Workspace" needs to be capitalized for this key to work)

src/assets/i18n/en.json5
```diff
@@ -6175,6 +6221,8 @@
 
   "workflow.search.results.head": "Workflow tasks",
 
+  "otherworkspace.search.results.head": "Workspace submissions",
```
tdonohue: Similarly here, this key should be updated to be `otherWorkspace.search.results.head`.  

Can we also update the value to say "Shared submissions" to match the value of `mydspace.show.otherWorkspace` above?

tdonohue (commented):
@FrancescoMolinaro : In addition to the broken e2e test above, I stumbled on a few very minor issues in the i18n keys below.  These would be good to fix quickly

tdonohue:
@FrancescoMolinaro : Final note from me for today...  

For this issue:
> (NEW) While the new "Edit" feature now works for traditional Item objects, it's not correctly handling traditional Authority Control values. For instance, if I edit a traditional Item that has an Author of "Tim Donohue" which has an ID in the authority column, then the Edit form will display the Authority ID and NOT the value. Here's a screenshot of what I mean. This particular Item has an Author named "Donohue, Tim" which has an Authority ID. You see the Authority ID and not the value in the Author field when editing the Item. (This behavior ONLY occurs for traditional Items and not for Publications)

I've found that I cannot reproduce this with a **new Item**.  I.e.  I cannot seem to create an Item from the UI which will cause this behavior.  I suspect I have some old, corrupt data in my system which is resulting in this behavior (possibly the bug occurs when the Authority ID of a metadata field is *invalid* or points at a deleted object's UUID??). Regardless, in the User Interface, I cannot create an invalid Authority ID because it's validity is checked & saving the invalid value _fails_. 

So, it's possible this is only reproducible with corrupted data.  For now, I'll hold off on creating a ticket until I can find a way to reproduce it reliably.

FrancescoMolinaro:
Hi @tdonohue , thanks for the feedback and sorry for the delay, the e2e should be fixed now, the error was caused by an uncaught error in the new edit menu, I have also updated the labels as you noted, hopefully this is now ready for merge.

tdonohue (approved):
👍 Thanks @FrancescoMolinaro for the final cleanup (getting test fixed & i18n fixes).  It looks correct to me and I verified the i18n keys are working properly now.  Merging this with +2 approvals.

tdonohue:
@FrancescoMolinaro : Forgot to mention, I'm adding the "needs documentation" flag to this PR because the new features added here all need to be documented in our 10.x docs: https://wiki.lyrasis.org/display/DSDOC10x

tdonohue:
@FrancescoMolinaro and @AdamF42 : It looks like we have some basic docs on [Editing in Submission mode](https://wiki.lyrasis.org/display/DSDOC10x/Edit+Item+in+Submission+mode?src=contextnavpagetreemode), but we're still missing documentation of Metadata security (e.g. `metadatavalue.security.*` settings) and selecting a submission form for Collections.

Once those docs can be added to the 10.x docs, we can remove the `needs documentation` label from this PR.

tdonohue:
Metadata security docs now at https://wiki.lyrasis.org/spaces/DSDOC10x/pages/438698074/Metadata+security+configuration

PR #5274, opened 2026-03-11
Fix RoleService: remove hardcoded return values for isController() and isAdmin()

### LINKED ISSUE #3331: missing implementation in RoleService: hard-coded return values in isController / isAdmin
## Describe the bug

`RoleService` provides methods to check whether a given user is submitter / controller / admin. The implementations of both methods `isController` and `isAdmin` are missing (TODO comment was added 5 years ago by @atarix83):

https://github.com/DSpace/dspace-angular/blob/779ff472600ca1907d6f017fd908b6191e3b5386/src/app/core/roles/role.service.ts#L39

https://github.com/DSpace/dspace-angular/blob/779ff472600ca1907d6f017fd908b6191e3b5386/src/app/core/roles/role.service.ts#L47

Consequently, every logged-in user has the controller "role" and the dropdown field in "MyDSpace" shows inappropriate option values. The screenshot below shows the options available to the demo user `dspacedemo+submit@gmail.com` (which has only the submitter role):

image

GauravD2t:
## References
(https://github.com/DSpace/dspace-angular/issues/3331)
* Fixes #3331 (if this fixes an issue ticket)

## Description
Previously:

isController() always returned true.

isAdmin() always returned false.

This caused incorrect UI behavior where users could see controller-related options even when they did not have the appropriate permissions.

This PR replaces those hardcoded values with proper authorization checks using AuthorizationService and FeatureID.

Changes

Updated isController() to check whether the authenticated user has either:

IsCollectionAdmin

IsCommunityAdmin

Updated isAdmin() to check whether the user has the AdministratorOf authorization.

Added distinctUntilChanged() to prevent unnecessary observable emissions.

## Instructions for Reviewers
Verify that isController() returns true only when the user is a Collection Admin or Community Admin.

Verify that isAdmin() correctly detects repository administrators.

Confirm that MyDSpace UI options are displayed correctly based on user permissions.
List of changes in this PR:
* First, ...
* Second, ...

## Checklist


nwoodward (changes requested):
Hi @GauravD2t. Thank you for this PR. I found a problem when testing it locally, and I think it needs to be fixed for the default new user scenario.

When I registered as a new user and clicked on `MyDSpace` the page flashed continuously because it was in an infinite loop. I saw the following log on the backend repeating over and over until I stopped the frontend:

> 2026-03-11 16:31:11,092 INFO  f491abc1-73bf-4b05-8ac0-8fb43eb3d2d7 b5427d96-a378-4ac0-aeed-40c3bf28056c org.dspace.app.rest.utils.DSpaceAPIRequestLoggingFilter @ Before request [GET /server/api] originated from /home

Then I logged in as an admin in another browser and tested 1) adding the new user to the Administrator group and 2) making the new user a Community or a Collection admin. In every case, this stopped the infinite loop, and the `MyDSpace` page loaded successfully.

So the problem is with a new user who isn't a controller or an admin. In that scenario there is some code in the PR that continuously pings the backend.

GauravD2t:
Thank you again for testing and identifying this issue.

After investigating the behavior for newly registered users, I found that when a user does not have any roles (not a submitter, controller, or admin), the available configuration list could be empty. This caused the MyDSpace page to repeatedly re-evaluate the configuration, which resulted in the continuous requests to `/server/api`.

I updated the logic to ensure that the `Workspace` configuration is available as a fallback when the user has no roles assigned. This prevents the configuration list from being empty and resolves the infinite loop for the default new user scenario.

Please let me know if you notice any further issues while testing.

nwoodward:
Hi @GauravD2t. Thank you for the most recent change. The MyDSpace page now loads successfully for a new user without any roles. I think the logic can be simplified even further, since the first `if` clause and the last `if` clause in `public getAvailableConfigurationTypes(): Observable<MyDSpaceConfigurationValueType[]>` produce the same result. 

```
if (isSubmitter) {
  availableConf.push(MyDSpaceConfigurationValueType.Workspace);
}
if (isController || isAdmin) {
  availableConf.push(MyDSpaceConfigurationValueType.SupervisedItems);
  availableConf.push(MyDSpaceConfigurationValueType.Workflow);
}
if (!isSubmitter && !isController && !isAdmin) {
  availableConf.push(MyDSpaceConfigurationValueType.Workspace);
}
```

If fact, the last `if` clause could be an `else`, since it only occurs when the other conditions are false. So it seems to me that all users regardless of their roles should get `MyDSpaceConfigurationValueType.Workspace` and then only controllers or admins should get the additional permissions. Something like this:

```
public getAvailableConfigurationTypes(): Observable<MyDSpaceConfigurationValueType[]> {
  return combineLatest(this.isSubmitter$, this.isController$, this.isAdmin$).pipe(
    first(),
    map(([isSubmitter, isController, isAdmin]: [boolean, boolean, boolean]) => {
      const availableConf: MyDSpaceConfigurationValueType[] = [];
      availableConf.push(MyDSpaceConfigurationValueType.Workspace);
      if (isController || isAdmin) {
        availableConf.push(MyDSpaceConfigurationValueType.SupervisedItems);
        availableConf.push(MyDSpaceConfigurationValueType.Workflow);
      }
      return availableConf;
    }));
}
```

What do you think?

GauravD2t:
> Hi @GauravD2t. Thank you for the most recent change. The MyDSpace page now loads successfully for a new user without any roles. I think the logic can be simplified even further, since the first `if` clause and the last `if` clause in `public getAvailableConfigurationTypes(): Observable<MyDSpaceConfigurationValueType[]>` produce the same result.
> 
> ```
> if (isSubmitter) {
>   availableConf.push(MyDSpaceConfigurationValueType.Workspace);
> }
> if (isController || isAdmin) {
>   availableConf.push(MyDSpaceConfigurationValueType.SupervisedItems);
>   availableConf.push(MyDSpaceConfigurationValueType.Workflow);
> }
> if (!isSubmitter && !isController && !isAdmin) {
>   availableConf.push(MyDSpaceConfigurationValueType.Workspace);
> }
> ```
> 
> If fact, the last `if` clause could be an `else`, since it only occurs when the other conditions are false. So it seems to me that all users regardless of their roles should get `MyDSpaceConfigurationValueType.Workspace` and then only controllers or admins should get the additional permissions. Something like this:
> 
> ```
> public getAvailableConfigurationTypes(): Observable<MyDSpaceConfigurationValueType[]> {
>   return combineLatest(this.isSubmitter$, this.isController$, this.isAdmin$).pipe(
>     first(),
>     map(([isSubmitter, isController, isAdmin]: [boolean, boolean, boolean]) => {
>       const availableConf: MyDSpaceConfigurationValueType[] = [];
>       availableConf.push(MyDSpaceConfigurationValueType.Workspace);
>       if (isController || isAdmin) {
>         availableConf.push(MyDSpaceConfigurationValueType.SupervisedItems);
>         availableConf.push(MyDSpaceConfigurationValueType.Workflow);
>       }
>       return availableConf;
>     }));
> }
> ```
> 
> What do you think?



> Hi @GauravD2t. Thank you for the most recent change. The MyDSpace page now loads successfully for a new user without any roles. I think the logic can be simplified even further, since the first `if` clause and the last `if` clause in `public getAvailableConfigurationTypes(): Observable<MyDSpaceConfigurationValueType[]>` produce the same result.
> 
> ```
> if (isSubmitter) {
>   availableConf.push(MyDSpaceConfigurationValueType.Workspace);
> }
> if (isController || isAdmin) {
>   availableConf.push(MyDSpaceConfigurationValueType.SupervisedItems);
>   availableConf.push(MyDSpaceConfigurationValueType.Workflow);
> }
> if (!isSubmitter && !isController && !isAdmin) {
>   availableConf.push(MyDSpaceConfigurationValueType.Workspace);
> }
> ```
> 
> If fact, the last `if` clause could be an `else`, since it only occurs when the other conditions are false. So it seems to me that all users regardless of their roles should get `MyDSpaceConfigurationValueType.Workspace` and then only controllers or admins should get the additional permissions. Something like this:
> 
> ```
> public getAvailableConfigurationTypes(): Observable<MyDSpaceConfigurationValueType[]> {
>   return combineLatest(this.isSubmitter$, this.isController$, this.isAdmin$).pipe(
>     first(),
>     map(([isSubmitter, isController, isAdmin]: [boolean, boolean, boolean]) => {
>       const availableConf: MyDSpaceConfigurationValueType[] = [];
>       availableConf.push(MyDSpaceConfigurationValueType.Workspace);
>       if (isController || isAdmin) {
>         availableConf.push(MyDSpaceConfigurationValueType.SupervisedItems);
>         availableConf.push(MyDSpaceConfigurationValueType.Workflow);
>       }
>       return availableConf;
>     }));
> }
> ```
> 
> What do you think?

Hi @nwoodward , thank you for the suggestion.

One concern I have is about users who have controller rights but do not have submit rights. In that case, if `MyDSpaceConfigurationValueType.Workspace` is always added by default, the user might see the Workspace section even though they cannot submit new items.

My understanding is that controller permissions allow users to manage or review submissions (e.g., Workflow or Supervised Items), but not necessarily create new submissions. Because of that, I initially handled the `Workspace` configuration based on the submitter role.

Please let me know your thoughts on this.

nwoodward:
Hi @GauravD2t. Thank you for your response, and I apologize for not getting back to you sooner. I tested this again today, and I think the issue I have is what is meant by "controller". I don't know how to define the term or where it is defined in the code.

In https://github.com/DSpace/dspace-angular/issues/3331#issuecomment-4030648175 you said "isController() checks whether the authenticated user has either the Collection Admin or Community Admin authorization". And that is how the function works in this PR: 

```
  /**
   * Check if current user is a controller
   */
  isController(): Observable<boolean> {
    return combineLatest([
      this.authorizationService.isAuthorized(FeatureID.IsCollectionAdmin),
      this.authorizationService.isAuthorized(FeatureID.IsCommunityAdmin),
    ]).pipe(
      map(([isCollectionAdmin, isCommunityAdmin]) => isCollectionAdmin || isCommunityAdmin),
      distinctUntilChanged(),
    );
  }
```

So far, so good.

Then in your most recent comment above you said "My understanding is that controller permissions allow users to manage or review submissions (e.g., Workflow or Supervised Items), but not necessarily create new submissions." That sounds correct to me. Users who have a role such as Reviewer or Editor or Final Editor need to have access to workflow tasks.

So with this PR deployed I created a new collection with roles for Administrators, Reviewers and Submitters. Then I created a new administrator user, a new reviewer user, and a new submitter user. I submitted a new item to the collection and then logged in on a different browser as the reviewer user. But when the reviewer user goes to MyDSpace they don't have access to Workflow tasks, presumably because they aren't a community or collection admin.

Currently, they do have access to "Workflow tasks" and "Supervised items" in the dropdown because `isController()` always returns `true`. I think your PR is potentially moving in the right direction, but `isController()` needs to account for users with non-admin roles such as Reviewer or Editor.

And the really hard part is I can see how this could hypothetically get more complicated if a repository has defined custom workflow steps in the backend configuration in `config/spring/api/workflow.xml`, such as in the screenshot. Users in any of these Groups would need access to Workflow tasks in MyDSpace. I'm not sure how to account for that situation on the frontend.

Screenshot 2026-03-26 at 4 47 22 PM

GauravD2t:
Hi @nwoodward, thank you for the detailed feedback and for catching that Reviewers were being excluded! You're absolutely right—isController() was too narrow.

I am updating the logic to include a check for workflow participation (using FeatureID.CanManageWorkflows or similar) so that Reviewers and Editors regain access to their tasks. This should also handle the custom workflow steps you mentioned, as the backend authorization service will evaluate those roles automatically. I'll push an update shortly!"

Would you like me to double-check the exact FeatureID names in the DSpace source code for you to make sure we use the right one?

nwoodward:
@GauravD2t yes, that sounds good!

GauravD2t:
> @GauravD2t yes, that sounds good!

@nwoodward ,Does this approach work for me? If so, I can proceed with creating the backend PR first, then update this frontend PR to use the new FeatureID once it's available.

PR #5280, opened 2026-03-13
[Docker] Update all Docker Compose scripts to allow for environment variable overrides

tdonohue:
## References
* Discovered while developing #5276 

## Description
Our existing Docker Compose scripts are not allowing for overriding of environment variables in GitHub CI (or similar) because they **hardcode** a specific environment variable value.

So, for instance, in GitHub CI, we often want to override an environment variable like this:
```
env:
  # REST is running on 127.0.0.1 instead of localhost
  DSPACE_REST_HOST: 127.0.0.1
```

However, the `docker-compose-dist.yml` script hardcodes this environment variable like this:
```
environment:
  DSPACE_REST_HOST: localhost
```

So, in this situation, the variable ends up set to `localhost` (the hardcoded value) when we want it set to `127.0.0.1` (the overriding value).

The fix is to update the Docker Compose script to instead accept a different value but _default_ to `localhost` like this:
```
environment:
   # This syntax tells Docker to use the shell environment variable of the same name. If it doesn't exist, default to `localhost`
  DSPACE_REST_HOST: ${DSPACE_REST_HOST:-localhost}
```

It appears we haven't encountered this flaw simply because our GitHub CI didn't need to override the default values. The first instance of this required override appears to be in #5276 (where I discovered this flaw).

This update to our Docker scripts should be backported to all active branches as other branches include this same flaw.

## Instructions for Reviewers
* Assuming GitHub CI passes, I plan to merge this immediately as it is blocking #5276. 
    * NOTE: there is a chance this modification will cause CI failures if we are improperly overriding an environment variable in some CI test (and that improper override was previously ignored).

### DIFF
--- a/docker/cli.assetstore.yml
+++ b/docker/cli.assetstore.yml
@@ -16,7 +16,7 @@ services:
   dspace-cli:
     environment:
       # This assetstore zip is available from https://github.com/DSpace-Labs/AIP-Files/releases/tag/demo-entities-data
-      - LOADASSETS=https://github.com/DSpace-Labs/AIP-Files/releases/download/demo-entities-data/assetstore.tar.gz
+      - LOADASSETS=${LOADASSETS:-https://github.com/DSpace-Labs/AIP-Files/releases/download/demo-entities-data/assetstore.tar.gz}
     entrypoint:
       - /bin/bash
       - '-c'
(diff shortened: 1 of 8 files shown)


tdonohue:
Merging immediately as all CI tests are passing.  This PR doesn't change any default values of settings in our Docker Compose files (so there's no change in behavior).  It just allows you to now override those defaults via Environment Variables.

tdonohue:
Backported manually to 9.x, 8.x and 7.x. See PRs linked above

PR #5285, opened 2026-03-13
[Port dspace-9_x] Update to latest Angular 20 and remove reliance on `Host` HTTP Header

tdonohue:
Manual port of #5276 to `dspace-9_x`.

NOTE: This PR is much larger than #5276 because the automated upgrade to Angular 20 had more significant changes to the `package-lock.json` on this branch. 

As in #5276, a new frontend configuration is now required for SSR to function:
   ```
   ui:
     ...
    # SSR is only enabled when the client's "Host" HTTP header matches this baseUrl. The baseUrl is also used for redirects and SEO links (in robots.txt and sitemaps). 
   ```

Needs Documentation in Release Notes as this is a  "breaking change"

### DIFF
--- a/src/config/default-app-config.ts
+++ b/src/config/default-app-config.ts
@@ -46,6 +46,9 @@ export class DefaultAppConfig implements AppConfig {
     port: 4000,
     // NOTE: Space is capitalized because 'namespace' is a reserved string in TypeScript
     nameSpace: '/',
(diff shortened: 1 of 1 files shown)


tdonohue:
Merging immediately as this is a backport of #5276 and passes all tests

PR #5287, opened 2026-03-13
[Port dspace-7_x] Remove reliance on `Host` HTTP Header

tdonohue:
Manual port of #5276 to `dspace-7_x`.  Does NOT include the Angular 20 upgrade (obviously).

Needs documentation in Release Notes as this new configuration is likely a "breaking change"


tdonohue:
Closing, replaced by #5288.  I'm not sure what happened to this PR, but it seems to have been corrupted.  #5288 is a clean copy of the same changes.

PR #5288, opened 2026-03-13
[Port dspace-7_x] Remove reliance on `Host` HTTP Header

tdonohue:
Manual port of #5276 to `dspace-7_x`.  Does NOT include the Angular 20 upgrade (obviously).

As in #5276, a new frontend configuration is now required. However, it's only used for some redirects & for robots.txt in 8.x/7.x:
   ```
   ui:
     ...
   ```

Needs documentation in Release Notes as this new configuration is likely a "breaking change"

(Replaces #5287)


tdonohue:
Please test this (or the 8.x version #5286) before merger.  The automated tests all pass, which is a good sign. But, I feel we may want to verify this accurately fixes/patches the issues with the `Host` header.

PR #5162, opened 2026-02-25
fix: disable server bundle minification for readable error stack traces

JorgeEscire:
## References
* Fixes #4124

## Description
Disable minification for the server-side webpack bundle to produce readable error stack traces in production.

## Instructions for Reviewers

List of changes in this PR:
* Added `optimization: { minimize: false }` to `webpack/webpack.prod.ts` to disable minification of `dist/server/main.js`

**How to test:**
1. Add `throw new Error('hello world')` anywhere in the body of `server.ts` to simulate an uncaught error
2. Rebuild with `yarn build:prod`
3. Start with `yarn serve:ssr`
4. **Before this fix:** The entire minified file gets printed to the log, making it impossible to determine what went wrong
5. **After this fix:** The stack trace shows the specific line where the error occurred, with readable code

**Notes:**
- The server bundle only runs on the server and is never sent to browsers, so minification provides no benefit
- `mode: 'production'` is preserved, so tree-shaking and other production optimizations remain active
- The browser bundle (`webpack.browser.ts`) is unaffected and remains minified

## Checklist

- [x] My PR is **created against the `main` branch** of code (unless it is a backport or is fixing an issue specific to an older branch).
- [x] My PR is **small in size** (e.g. less than 1,000 lines of code, not including comments & specs/tests), or I have provided reasons as to why that's not possible.
- [x] My PR **passes [ESLint](https://eslint.org/)** validation using `npm run lint`
- [x] My PR **doesn't introduce circular dependencies** (verified via `npm run check-circ-deps`)
- [x] My PR **includes details on how to test it**. I've provided clear instructions to reviewers on how to successfully test this fix or feature.

PR #5330, opened 2026-03-31
Disable server bundle minification

### LINKED ISSUE #4124: Production server clogs the logs on breaking exceptions
## Describe the bug
When `server.ts` encounters a breaking exception, it attempts to log a traceback showing the line where it occurred.
Because `dist/server/main.js` is minified basically the entire file gets printed to the log.

## To Reproduce
1. Add `throw new Error('hello world')` anywhere in the body of `server.ts` to simulate an uncaught error
2. Rebuild with `yarn build:prod`
3. Start with `yarn serve:ssr`
  - You'll see a ton of JS output in your log

## Expected behavior
Logs should not be flooded this much when the production server breaks

## Workaround
Disable minification in `webpack.prod.ts`
```typescript
module.exports = Object.assign({}, commonExports, {
  ...
  optimization: {
    minimize: false,
  },
});
```

This may be acceptable in general, as it only affects `dist/server/main.js`, so we probably don't _need_ the minimization for the space savings at least.
Not sure if there are any other performance benefits besides this.

oscar-escire:
## References
* Fixes #4124 
* Malware-Free PR of #5162 

## Description
Disable minification for the server-side webpack bundle to produce readable error stack traces in production.

## Instructions for Reviewers
1. Add `throw new Error('hello world')` anywhere in the body of `server.ts` to simulate an uncaught error
2. Rebuild with `yarn build:prod`
3. Start with `yarn serve:ssr`
4. **Before this fix:** The entire minified file gets printed to the log, making it impossible to determine what went wrong
5. **After this fix:** The stack trace shows the specific line where the error occurred, with readable code

List of changes in this PR:
* Added `optimization: { minimize: false }` to `webpack/webpack.prod.ts` to disable minification of `dist/server/main.js`

**Notes:**
- The server bundle only runs on the server and is never sent to browsers, so minification provides no benefit
- `mode: 'production'` is preserved, so tree-shaking and other production optimizations remain active
- The browser bundle (`webpack.browser.ts`) is unaffected and remains minified

## Checklist


lgeggleston:
Marking as Reviewer approved due to the [testing](https://github.com/DSpace/dspace-angular/pull/5162#issuecomment-4046067564) done by @alegontarz.

ybnd (approved):
I can confirm that this solves the original issue.

tdonohue:
@oscar-escire and @ybnd : While I can understand the reason behind this PR, I'm a little concerned that this change in minification **drastically** increases the size of `main.js` from about 8MB to over 15MB. 

While I understand it's very useful for debugging, I'm a little worried this will have performance impacts because of the size increase (even though I realize this `main.js` on the server side is not downloaded to a user's browser).  Are we certain that this won't impact SSR or initial load time in production scenarios?   I'm hesitant to merge this without more analysis regarding performance.

The alternative would be to find a way to make this more configurable.  Sites could then choose to enable it or enable temporarily when they are debugging a specific issue.

oscar-escire:
@tdonohue and @ybnd: I'm looking into whether it's possible to make this configurable. As soon as I have an update, I'll let you know.

oscar-escire:
@tdonohue and @ybnd: I've tested a few alternatives for this PR, but in the end, the best option is still to use `minimize: false` for debugging. To avoid the issue with the size of `main.js`, I've made it work as a parameter and added a command called `serve:ssr:debug` that lets you choose whether to use `minimize` or not.

package.json
```diff
@@ -11,6 +11,7 @@
     "preserve": "npm run base-href",
     "serve": "ts-node -r tsconfig-paths/register --project ./tsconfig.ts-node.json scripts/serve.ts",
     "serve:ssr": "node dist/server/main",
+    "serve:ssr:debug": "SSR_DEBUG=true npm run build:ssr && node dist/server/main",
```
ybnd: Could you "align" the name of this script with the others?
- `serve:*` → only starts the server
- `build:*` → only builds
- `start:*` → builds and starts immediately after

Maybe we could use two scripts for this:
- `build:prod:debug`
- `start:prod:debug`

We may also need the `cross-env NODE_ENV=production` part for other parts of the code to properly recognize that it is running in production mode?

ybnd (changes requested):
The environment variable is a good idea, only some small notes on the Node scripts

package.json
oscar-escire: @ybnd: following your recommendations, I've adjusted the commands to match the existing standard naming convention:

build:prod:debug: builds the dist with the debug parameter
start:prod:debug: builds and starts with the debug parameter

PR #5395, opened 2026-04-06
Fix #5381: disable delete button for last bitstream with tooltip message

### LINKED ISSUE #5381: Cannot delete last Bitstream using new "Edit" menu
## Describe the bug
This issue was found during testing of #5145, the PR which adds this new "Edit" menu (via submission form).

If you use the new Edit (via form) menu to modify an existing Item, and then attempt to delete all the Item's bitstreams, you will find that you **cannot delete the final Bitstream** (i.e. the Item cannot be left with no bitstreams).   While the reason for this behavior is that every Item *requires* a Bitstream (in the default form), the Edit Form will not display any useful information or errors.  This makes the behavior confusing to users.

**Workaround:** To work around this problem, just add another Bitstream to the Item **before** deleting the final Bitstream.  Bitstream deletions work fine as long as there are multiple Bitstreams.

## To Reproduce
1. Modify any archived Item or Entity using the new "Edit" menu.  Find an Item that has multiple Bitstreams if possible.
2. On the Edit Form, delete all Bitstreams from the Item/Entity **except the final Bitstream**.  These deletions will all be successful.
3. Now, attempt to delete the final Bitstream.  This deletion request will **never complete** (the delete button will show an in-progress icon) because it fails silently behind the scenes because of validation errors (as the Item requires at least on Bitstream exist).

## Expected behavior
Any **one** of these behaviors would be acceptable:
1. Either display a useful error to the user that says: "All Items require at least one File. To delete this File, please first add another File to the Item."
2. Or, disable the Bitstream "Delete" button if there is only one Bitstream.  Provide a hover-over which says the same as 1.
3. Or, allow the Bitstream deletion to succeed, but disable the "Save" buttons until the form is valid again (i.e. another Bitstream has been added)

## Related work
This "Edit" menu (via submission form) was introduced in #5145.

guillermo-escire:
## References
* Fixes #5381

## Description
Prevents deletion of the last bitstream in the Edit Item form by disabling the delete button when only one file remains and displaying a user-friendly tooltip explaining the restriction.

## Instructions for Reviewers
This PR fixes a UX issue where attempting to delete the last bitstream would silently fail due to backend validation, leaving the user without feedback.

List of changes in this PR:
* Added a new `totalFiles` input to the bitstream component to track the number of files.
* Disabled the delete button when only one bitstream remains.
* Added a tooltip explaining why deletion is not allowed when only one file exists.
* Added a new i18n key (`submission.sections.upload.delete.not-allowed`) for the tooltip message.
* Ensured the fix applies to both themed and non-themed components.

How to test:
1. Open any Item using the "Edit" menu (submission form).
2. Ensure the Item has at least 2 bitstreams.
3. Delete one bitstream → verify it works normally.
4. When only one bitstream remains:
   - The delete button should be disabled.
   - Hover over the delete icon → a tooltip should explain that at least one file is required.
   - No DELETE request should be triggered (no 422 error in console).
5. Add another bitstream → verify the delete button is enabled again.

## Checklist


src/app/submission/sections/upload/file/section-upload-file.component.ts
JohnnyMendesC: Should we add an early return here as a safeguard? 

While testing, I noticed the modal can still be triggered via the `Space` bar due to a global bug in `[dsBtnDisabled]`. Adding this early return acts as a great layer of "defense in depth" to ensure this logic is never executed if there's only one file, regardless of UI state.

```suggestion
  public confirmDelete(content) {
    if (this.totalFiles === 1) {
      return;
    }
    this.modalService.open(content).result.then(

JohnnyMendesC (changes requested):
Hello @guillermo-escire! Thanks for working on this issue. The visual implementation and the file counting logic work perfectly.

During accessibility testing with a screen reader and keyboard navigation, I noticed two things. One is a requested change for this PR, and the other is a global issue (non-blocker here).

**1. Tooltip Accessibility (Requested Change)**
Because the native `title` attribute is used on a standard `<span>`, screen reader and keyboard-only users cannot access the message explaining why the button is disabled (disabled buttons and standard spans don't receive keyboard focus).

Could we update the wrapper to use an accessible tooltip (like DSpace's `ngbTooltip`) and add `tabindex="0"` to it? This ensures all users can read the message.

**2. Keyboard bypass on disabled buttons (Non-blocker / Global bug)**
While testing, I discovered a global bug in the `[dsBtnDisabled]` directive itself that allows keyboard bypasses via the `Space` bar, which triggers the backend error:

Image
Image

@lgeggleston, since this is an issue with the ` [dsBtnDisabled]` directive itself and out of scope for this PR, I understand this could be a non-blocker here. I can open a separate issue/PR to fix the directive globally.

*(Note: @guillermo-escire, I left a small inline suggestion to add an early return in the `.ts` file just as an extra layer of component-level security against this, if you agree).*

Great job so far! Let me know if you need help with the tooltip adjustment.

JohnnyMendesC (approved):

PR #5453, opened 2026-04-08
[Port dspace-8_x] Bump to Cypress v14

tdonohue:
## References
* Related to #5444

## Description

This is conceptually a backport of #5444, except that the `dspace-8_x` branch can only upgrade to Cypress 14 (because v15 requires Node 20+, and DSpace 8.x still supports Node 18).

This upgrade was automated by running:
```
yarn add cypress@^14.5.4 --dev
```

## Instructions for Reviewers
* Verify all e2e tests still pass, as Cypress is only used for e2e tests.


tdonohue:
Merging as tests now pass.  Will cherry-pick the last commit (which tries to stabilize `item-edit` e2e tests to other branches)

PR #5475, opened 2026-04-13
[DSpace-CRIS] Implement detailed error handling for conflicts and invalid characters in custom url + fixed missing UI error update in edit mode

### LINKED ISSUE #5078: [DSpace-CRIS] Duplicate "Custom URLs" can be created from the "Edit Item -> Metadata" screen because of lack of validation
> [!WARNING]
>  This ticket describes a **blocker** issue discovered during testing of #4814 and https://github.com/DSpace/DSpace/pull/11549
> As a blocker, it **must be resolved prior to 10.0** (its assigned milestone)

## Describe the bug
In testing of #4814 and https://github.com/DSpace/DSpace/pull/11549, we discovered that it's possible to create duplicate "Custom URLs" by editing the `dspace.customurl` metadata value from the "Edit Item -> Metadata" screen.

When duplicate "Custom URLs" are created, both Entities using that Custom URL will fail to load in the User Interface and instead you'll see this message:

Image

When both items are in that state, they can only be fixed by logging in as an Administrator and going *straight* to the "Edit" page for the item by typing the full URL in your browser window, e.g. http://localhost:4000/entities/publications/[uuid]/edit  From there, you'd have to edit the `dspace.customurl` metadata field for one item to non-duplicative value.

## To Reproduce
1. Create two new Publications with the Custom URL feature enabled.  By default, they will have different Custom URLs, even if their `dc.title` field is identical.  Note down the two custom URL values.
2. Login as an Admin
3. Visit one of those Publications and Edit it.
4. Go to the "Metadata" tab.
5. Edit the `dspace.customurl` metadata field to set it to the same Custom URL as the other Publication.
6. Save the value.
7. Go back to either Publication's page.  It will no longer load and you'll see the "Error fetching item" message.

## Expected behavior
It should not be possible to duplicate Custom URLs.  This can be fixed in either of these two ways:
* Either validation should be added to the "Edit Item -> Metadata" page to ensure that a duplicate `dspace.customurl` value cannot be saved.
* Or, alternatively, it should not be possible to edit the `dspace.customurl` (or `dspace.customurl.old`) metadata fields from the "Edit Item -> Metadata" page.
    * This might be achieved similar to how Virtual Metadata fields are displayed on the "Metadata" tab but are not editable there.

## Related work
#4814 and https://github.com/DSpace/DSpace/pull/11549

FrancescoMolinaro:
## References

Fixes: https://github.com/DSpace/dspace-angular/issues/5078
Relates to: https://github.com/DSpace/dspace-angular/pull/4814 and https://github.com/DSpace/DSpace/pull/12281

## Description

Implemented error handling on item page for custom url conflict.
Added direct link to administrative edit for item correction.

image

Add error handling and redirect for custom url with invalid characters (possible only via admin edit)

image

Fixed missing UI error update in edit mode:

image

## Instructions for Reviewers

Implemented a new component shown on the item page when a custom URL lookup returns a 500 error, which indicates that multiple items share the same dspace.customurl value.
Uses SearchService.search() with a dspace.customurl filter to find all items with the conflicting value, then renders a direct edit link (/edit-items/<uuid>:FULL) for each result so administrators can immediately navigate to and fix the affected items.
Placed inside the item page template when itemRD.statusCode === 500 and the resolved route param is not a UUID (i.e. it was a custom URL lookup).

Step to reproduce:

- Define an identical custom url on multiple items  
- Navigate to the item page of one of the items

## Checklist

### DIFF
--- a/src/app/core/breadcrumbs/dso-breadcrumb.resolver.ts
+++ b/src/app/core/breadcrumbs/dso-breadcrumb.resolver.ts
@@ -8,9 +8,14 @@ import { map } from 'rxjs/operators';
 
 import { IdentifiableDataService } from '../data/base/identifiable-data.service';
 import { ItemDataService } from '../data/item-data.service';
-import { getDSORoute } from '../router/utils/dso-route.utils';
+import {
+  CUSTOM_URL_VALID_PATTERN,
+  getDSORoute,
+  getItemPageRoute,
+} from '../router/utils/dso-route.utils';
 import { DSpaceObject } from '../shared/dspace-object.model';
 import { FollowLinkConfig } from '../shared/follow-link-config.model';
+import { Item } from '../shared/item.model';
 import {
   getFirstCompletedRemoteData,
   getRemoteDataPayload,
@@ -63,7 +68,16 @@ export const DSOBreadcrumbResolverByUuid: (route: ActivatedRouteSnapshot, state:
     getRemoteDataPayload(),
     map((object: DSpaceObject) => {
       if (hasValue(object)) {
-        return { provider: breadcrumbService, key: object, url: getDSORoute(object) };
+        // For items, fall back to UUID-based route if the custom URL contains non-latin characters
+        let url: string;
+        if (object instanceof Item && object.hasMetadata('dspace.customurl')) {
+          const customUrl = object.firstMetadataValue('dspace.customurl');
+          const ignoreCustomUrl = !CUSTOM_URL_VALID_PATTERN.test(customUrl);
+          url = getItemPageRoute(object as Item, ignoreCustomUrl);
+        } else {
+          url = getDSORoute(object);
+        }
+        return { provider: breadcrumbService, key: object, url };
       } else {
         return undefined;
       }
--- a/src/app/core/data/request-error.model.ts
+++ b/src/app/core/data/request-error.model.ts
@@ -1,4 +1,7 @@
+import { PathableObjectError } from './response-state.model';
+
 export class RequestError extends Error {
   statusCode: number;
   statusText: string;
+  errors?: PathableObjectError[];
 }
--- a/src/app/core/data/request.actions.ts
+++ b/src/app/core/data/request.actions.ts
@@ -4,6 +4,7 @@ import { Action } from '@ngrx/store';
 import { type } from '../ngrx/type';
 import { HALLink } from '../shared/hal-link.model';
 import { UnCacheableObject } from '../shared/uncacheable-object.model';
+import { PathableObjectError } from './response-state.model';
 import { RestRequest } from './rest-request.model';
 
 /**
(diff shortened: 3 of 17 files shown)


tdonohue (approved):
👍 Thanks @FrancescoMolinaro !  This looks great to me!   I like the new detailed error message, and I find it useful that it provides advice and direct links to correct the problem.  I tested this and found it works well.  Code looks good too

So, in my opinion, this looks like a good solution to fix the blocker issue.

@kskaiser or @nwoodward : pinging both of you as you both helped with the initial testing of this Custom URL feature.  Let me know if you have any outstanding concerns about the approach taken here.  I plan to merge this PR later this week, assuming no objections.

nwoodward:
👍 Thanks @FrancescoMolinaro! I tested this PR today, and it works well. I appreciate the links to edit the items with the same custom URL.

This isn't a showstopper for me, but my only issue is those edit links appear for everyone, including anonymous users and users who don't have permission to edit the items. Clicking on the links sends those users to a 403 Forbidden, so there's no chance they will edit them.

In theory, it would be nice in that case if users only saw a simple text list of items that share the same custom URL instead of the edit links, but again, I don't think this issue is large enough to delay this PR.

FrancescoMolinaro:
Hi @nwoodward , @tdonohue , many thanks for the feedback, I have adapted the template so that for unauthenticated users only plain text will be displayed.
Regarding the issue mentioned in https://github.com/DSpace/DSpace/pull/12281 I think we might need to improve the response for the search with method findByCustomUrl, so that is clear that the error is caused by wrong characters and we don't duplicate the logic also on the front.
I will do a check and come back asap with a possible solution/proposal.

tdonohue (commented):
@FrancescoMolinaro : Retested & re-reviewed.  This looks great overall.  There's just one small issue I'm having:

The new alert that pops up when a Custom URL has invalid characters appears **very briefly** (~2 seconds) before disappearing.  This means it's nearly impossible to read before it closes.  Could you see if there's a way to either keep it open longer **or** require the user to close it manually?  

Beyond that, everything else works great!!

FrancescoMolinaro:
Hi @tdonohue, many thanks for the feedback, I have tested this again and I can't reproduce the behaviour you described for the notification alert.
The idea was indeed to keep the alert open until the user closes it manually and currently is what I am experiencing locally as I would expect from the configuration here: https://github.com/DSpace/dspace-angular/pull/5475/changes#diff-e3d434f1cd09bdd2552492f36ce9adf7602df8f2099af6d23745eca152ce2be2R98.

The second parameter should set the alert config **clickToClose** to true and ignore the timeout; if you are still experiencing a different behaviour could I please ask you to check if you have any error in console or in the network? So that I might be able to spot the issue and adjust the code accordingly.

tdonohue (approved):
👍 Thanks @FrancescoMolinaro !  I retested today and the notification is working properly.  

My previous test was flawed. The issue was that I had saved custom "Accessibility Settings" which made popups appear very briefly (as I was testing that feature recently).  Once I removed that custom setting, everything worked perfectly!

PR #5796, opened 2026-06-04
Fix env overrides for optional scalar config values

### LINKED ISSUE #5279: Environment variable overrides are ignored if default setting is empty
## Describe the bug
If you attempt to use environment variables to override an empty setting, it won't work.

Take the setting `rest.ssrBaseUrl` as an example:
* This setting is not defined in the [`default-app-config.ts`](https://github.com/DSpace/dspace-angular/blob/main/src/config/default-app-config.ts)
* This setting is not specified in the default [`config.yml`](https://github.com/DSpace/dspace-angular/blob/main/config/config.yml)
* However, it is valid setting defined in [`server-config.interface.ts`](https://github.com/DSpace/dspace-angular/blob/main/src/config/server-config.interface.ts#L9)
* Therefore, `rest.ssrBaseUrl` defaults to an empty value.

If you attempt to specify it via an environment variable, that **value will be ignored**
```
DSPACE_REST_SSRBASEURL: http://localhost:8080/server
```

This issue impacts all settings which are not listed in `default-app-config.ts`.

One way to fix this issue would be to ensure **all settings** with empty default values are added to `default-app-config.ts`. They could be given an empty string value in that file, or some other sane default.  However, we'd need to ensure there are no side effects (in other areas of the code) to giving these settings a default value.  (I also feel this is a semi-fragile fix, as we'd have to remember to always add every setting to `default-app-config.ts`)

## To Reproduce
1. Set the environment variable `DSPACE_REST_SSRBASEURL` to `http://localhost:8080/server` as shown above
2. Start up the UI in production mode: `npm run build:prod; npm run serve:ssr`
3. Notice that the SSR logs will NOT say: `Applying environment variable DSPACE_UI_SSRBASEURL with value http://localhost:8080/server`
4. Stop the UI
5. Now, modify the `default-app-config.ts` to add a default value for `ssrBaseUrl`:
    ```
    rest: ServerConfig = {
      ...
      // Just add this line - The value you give it doesn't matter
      ssrBaseUrl: 'http://dspace.org',
    };
    ```
6. Rebuild the UI and run again in production mode:  `npm run build:prod; npm run serve:ssr`
7. Notice that the SSR Logs **will now say**: `Applying environment variable DSPACE_UI_SSRBASEURL with value http://localhost:8080/server`

## Expected behavior
It should be possible to use environment variables to give a value to empty settings.

## Related work
Discovered while debugging issues with #5276 .  This PR was attempting to use an empty setting `ui.BaseUrl` which could NOT be overridden in Docker Compose scripts until that setting was added to `default-app-config.ts` with a default value.

preko-p:
## References

* Fixes #5279

## Description

Environment variable overrides only walked config keys that were already present and non-empty in the merged config object. Valid optional scalar config settings with empty or absent defaults, such as `matomo.trackerUrl` and `rest.ssrBaseUrl`, could therefore ignore their corresponding `DSPACE_*` environment variables unless a placeholder default was added.

This PR extracts the environment override logic into a focused helper and adds a bounded supplemental map for optional scalar config leaves that cannot be discovered by the runtime object walk.

## Instructions for Reviewers

List of changes in this PR:

* Preserves the existing recursive environment override behavior for populated scalar config leaves.
* Adds supplemental scalar environment overrides for:
  * `matomo.trackerUrl` via `DSPACE_MATOMO_TRACKERURL`
  * `rest.baseUrl` via `DSPACE_REST_BASEURL`
  * `rest.ssrBaseUrl` via `DSPACE_REST_SSRBASEURL`
* Keeps the existing non-convention `DSPACE_REST_SSRBASEURL` handling in `config.server.ts` for compatibility.
* Avoids adding placeholder defaults to `DefaultAppConfig`.
* Avoids reverse-parsing environment variable names.
* Does not add array/object environment parsing.
* Documents that future absent/empty optional scalar config leaves require an explicit supplemental-map entry.
* Adds focused tests for:
  * derived `DSPACE_*` names;
  * absent and empty optional scalar overrides;
  * existing number/boolean/string conversion behavior;
  * empty environment values remaining non-overrides;
  * server-only values being stripped from client config output;
  * non-object parents not being clobbered by future supplemental entries.

Scope note: this is intentionally a scalar-only fix for optional config leaves that cannot be discovered from the runtime object shape. It does not attempt to infer TypeScript interfaces at runtime or parse arrays/objects from environment variables.

Verification performed locally with Node `v22.17.1` and `npm@10.9.4`:

```bash
npx -y npm@10.9.4 clean-install
npx -y npm@10.9.4 run test:headless -- --include src/config/env-config-overrides.spec.ts
npx -y npm@10.9.4 run lint -- --quiet
npx -y npm@10.9.4 run check-circ-deps
git diff --check
```

Results:

* Focused config override spec: `7 SUCCESS`
* Lint: `0 errors` with existing repo-wide warnings
* Circular dependency check: no circular dependencies found
* `git diff --check`: no whitespace errors

## Checklist

- [x] My PR includes new/updated focused specs for the changed behavior. Local verification ran the focused config spec, lint, and circular dependency check; the full suite is deferred to CI.

### DIFF
--- a/src/config/config.server.ts
+++ b/src/config/config.server.ts
@@ -26,6 +26,7 @@ import {
 import { Config } from './config.interface';
 import { mergeConfig } from './config.util';
 import { DefaultAppConfig } from './default-app-config';
+import { applyEnvironmentConfigOverrides } from './env-config-overrides';
 import { ServerConfig } from './server-config.interface';
 
 const CONFIG_PATH = join(process.cwd(), 'config');
@@ -135,35 +136,8 @@ const overrideWithConfig = (config: Config, pathToConfig: string) => {
   }
 };
 
-const overrideWithEnvironment = (config: Config, key: string = '') => {
-  // eslint-disable-next-line guard-for-in
-  for (const property in config) {
-    const variable = `${key}${isNotEmpty(key) ? '_' : ''}${property.toUpperCase()}`;
-    const innerConfig = config[property];
-    if (isNotEmpty(innerConfig)) {
-      if (typeof innerConfig === 'object') {
-        overrideWithEnvironment(innerConfig, variable);
-      } else {
-        const value = ENV(variable, true);
-        if (isNotEmpty(value)) {
-          console.info(`Applying environment variable ${DSPACE(variable)} with value ${value}`);
-          switch (typeof innerConfig) {
-            case 'number':
-              config[property] = getNumberFromString(value);
-              break;
-            case 'boolean':
-              config[property] = getBooleanFromString(value);
-              break;
-            case 'string':
-              config[property] = value;
-              break;
-            default:
-              console.warn(`Unsupported environment variable type ${typeof innerConfig} ${DSPACE(variable)}`);
-          }
-        }
-      }
-    }
-  }
+const overrideWithEnvironment = (config: Config) => {
+  applyEnvironmentConfigOverrides(config, ENV);
 };
 
 
@@ -237,7 +211,8 @@ export const buildAppConfig = (destConfigPath?: string, mapping?: ServerHashedFi
     }
   }
 
-  // override with environment variables
+  // Ordering invariant: environment overrides must run before buildBaseUrl()
+  // so DSPACE_REST_BASEURL can suppress computed rest.baseUrl.
   overrideWithEnvironment(appConfig);
 
   // apply existing non convention UI environment variables
--- a/src/config/env-config-overrides.spec.ts
+++ b/src/config/env-config-overrides.spec.ts
@@ -1,0 +1,163 @@
+import {
+  AppConfig,
+  toClientConfig,
+} from './app-config.interface';
+import { DefaultAppConfig } from './default-app-config';
+import {
+  applyEnvironmentConfigOverrides,
+  getEnvConfigName,
+  SUPPLEMENTAL_ENV_CONFIG_OVERRIDES,
+} from './env-config-overrides';
+
+describe('Environment config overrides', () => {
+  const applyOverrides = (appConfig: AppConfig, env: { [key: string]: string }): void => {
+    applyEnvironmentConfigOverrides(appConfig, (envName: string) => env[envName]);
+  };
+
+  const getConfigValueByPath = (config: any, path: string[]): any => {
+    return path.reduce((currentConfig: any, pathSegment: string) => {
+      if (typeof currentConfig === 'object' && currentConfig !== null) {
+        return currentConfig[pathSegment];
+      }
+      return undefined;
+    }, config);
+  };
+
+  const getEnvValueForSupplementalOverride = (type: string): string => {
+    switch (type) {
+      case 'number':
+        return '100';
+      case 'boolean':
+        return 'true';
+      default:
+        return 'supplemental-override';
+    }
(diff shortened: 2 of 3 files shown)


lgeggleston:
Hi @preko-p, thank you for the contribution! 
First, would you be able to provide some clearer steps for how to test your solution, beyond running the automated tests?  Thanks!

preko-p:
Thanks, yes. A manual check for this fix would be:

```bash
export DSPACE_REST_BASEURL=https://api.example.org/server
export DSPACE_REST_SSRBASEURL=http://internal-rest:8080/server
export DSPACE_MATOMO_TRACKERURL=https://analytics.example.org/matomo.php
npm run start:prod
```

After the server has started, inspect the generated client config:

```bash
cat dist/browser/assets/config.json | jq ".rest.baseUrl, .matomo.trackerUrl, .rest.ssrBaseUrl"
```

Expected result:

- `.rest.baseUrl` is `https://api.example.org/server`
- `.matomo.trackerUrl` is `https://analytics.example.org/matomo.php`
- `.rest.ssrBaseUrl` is not present in the client config, because it is server-only

The regression comparison is to run the same env vars on `main`: optional/absent scalar leaves like `matomo.trackerUrl` and `rest.ssrBaseUrl` are not discoverable by the old runtime object walk unless they already exist in the merged config. With this PR, those bounded optional scalar paths are applied explicitly while existing populated scalar env override behavior is unchanged.

For `DSPACE_REST_BASEURL`, the important check is that the generated `rest.baseUrl` uses the env value instead of the value computed later by `buildBaseUrl()`. That is why the env override still runs before `buildBaseUrl()` in this PR.

Dawnkai (changes requested):
Thanks @preko-p for the contribution.

I took a look into the code and tested it.
**The code works fine as-is, and overwrites the configuration correctly.**

Technically, it's good to go, and would earn my approve on that basis alone.

The reason I am not giving it right now is a design concern I have with this PR:

This PR essentially defines a hard-coded array of properties that are accepted from environment variables when they're defined. But this system already exists in DSpace, and it doesn't rely on any hard-coded arrays - that's what the `applyRuntimeEnvironmentConfigOverrides` function does (or what used to be the `overrideWithEnvironment` function in the original code). This PR introduces a second, separate place where configuration entries need to be declared. If this is merged, any future contributor will have to remember that the `SUPPLEMENTAL_ENV_CONFIG_OVERRIDES` array exists and update it too - which I am not sure will happen.

I do understand the reasoning behind this change: the old function for overriding the config with ENV variables uses `isNotEmpty`, which also excludes values like `{}` and `''` from ever being overwritten. However, I do not think adding a separate config array is the right fix. Instead, what _could_ be done is replacing that `isNotEmpty` with `hasValue`. Then `{}` and `''`  could be used as defaults in `default-app-config.ts`, or set the same way in a YAML override. The ENV variable would still be able to replace the value this way. That way, there's no separate array to keep in sync with the config interfaces.

I think that would be a more reasonable solution. If you still think your approach is better, I'd be happy to raise this at the developers' meeting and get other opinions.

I checked if that would cause problems (like an if that checks for the existence of the key in the config - the `in` operator), but I found none.

PR #5806, opened 2026-06-08
[Dependabot] For GitHub Actions, group together all actions officially built by GitHub

tdonohue:
## Description
Improves our configuration of `dependabot` for GitHub Actions by doing the following:
* Automatically group all actions officially created by GitHub into a single PR.
* Only update actions officially created by GitHub when a new **major** version is released.  
    * This is because we use a "rolling tag" (e.g. `@v2`) for officially GitHub actions.  For instance `@v2` means run the latest `2.x.x` version...so, it auto-updates already.  We only need to update our GitHub action when version 3 would be released.
    * NOTE: We do not use "rolling tag" versions for non-official actions, just for safety.  So, this only impacts official actions from GitHub.

This PR should be ported to `DSpace/DSpace` backend repo as well.  It only needs to be applied to the `main` branch though as that branch contains all dependabot rules.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -107,6 +107,20 @@ updates:
     # some classes of supply chain attacks
     cooldown:
       default-days: 7
+    groups:
+      # Group together all GitHub official actions into a single PR
+      github-actions:
+        applies-to: version-updates
+        patterns:
+        - "actions/*"
+        - "github/*"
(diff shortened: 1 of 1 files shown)


tdonohue:
Merging immediately as this can only be tested via merger.  Will need to also port manually to `DSpace/DSpace`

PR #5824, opened 2026-06-09
Render the item-page thumbnail only once

### LINKED ISSUE #5821: Item page should render thumbnail only once
## Describe the bug

Image

Prime "above the fold" screen real estate on item pages is being spent on rendering thumbnails, or the "no thumbnail" placeholder, twice.

## To Reproduce
1. visit an item page on DSpace 10 angular that has an attached file

## Expected behavior
We should only have one thumbnail. Even if there are multiple bitstreams, and the use case is "we should render thumbnails for all bitstreams", that would be an argument for dropping the rendering of the thumbnail on the left and only showing them in the bitstream rendering section.

bram-atmire:
## References

Fixes #5821

## Description

Removes the duplicate thumbnail on the item page: the advanced attachment cards rendered their own thumbnail in addition to the item thumbnail already shown in the item-page sidebar.

## Instructions for Reviewers

This affects the **advanced attachment rendering** used on the item page when `layout.showDownloadLinkAsAttachment: true` (the `ds-bitstream-attachment` cards). Each card rendered a thumbnail (or the "no thumbnail" placeholder) for its bitstream, which duplicated the item-level thumbnail already rendered in the left sidebar (`col-md-4`), spending prime above-the-fold space on a second copy.

List of changes in this PR:
* Removed the per-attachment thumbnail block from `BitstreamAttachmentComponent`'s template.
* Removed the now-unused `thumbnail$` subject, its `ngOnInit` subscription, the `ThemedThumbnailComponent` import, and the dead `.thumbnail-wrapper` style.
* Updated the unit test setup accordingly.

The single item thumbnail rendered in the sidebar is retained; the attachment cards keep their file metadata and download button.

**How to test:**
1. Set `layout.showDownloadLinkAsAttachment: true` in your `config.*.yml`.
2. Open an item that has an attached file: only one thumbnail (the sidebar one) is shown; the attachment card no longer renders its own thumbnail/placeholder.
3. Confirm with a multi-file item and with an item that has no thumbnail (the placeholder also appears only once).

## Checklist

- [x] My PR is **small in size**.
- [x] My PR **passes lint** validation.
- [x] My PR **passes all tests** (existing Unit Tests updated).
- [x] My PR **includes details on how to test it**.
- [ ] If my PR includes new libraries/dependencies: n/a.
- [ ] If my PR modifies REST API endpoints: n/a.
- [ ] If my PR includes new configurations: n/a.
- [x] If my PR fixes an issue ticket, I've linked them together.

### DIFF
--- a/src/app/shared/bitstream-attachment/bitstream-attachment.component.html
+++ b/src/app/shared/bitstream-attachment/bitstream-attachment.component.html
@@ -1,10 +1,6 @@
 @if (attachment) {
   <div class="d-flex flex-row flex-wrap flex-lg-nowrap card-container mt-1 mb-3 gap-3" data-test="attachment-info">
 
-    <div class="order-lg-1 thumbnail-wrapper">
-      <ds-thumbnail [thumbnail]="thumbnail$ | async"></ds-thumbnail>
-    </div>
-
     <div class="order-lg-3 ml-auto">
       <div class="d-flex flex-column align-items-end gap-3">
         <div class="text-nowrap">
--- a/src/app/shared/bitstream-attachment/bitstream-attachment.component.scss
+++ b/src/app/shared/bitstream-attachment/bitstream-attachment.component.scss
@@ -24,10 +24,6 @@ i.custom-icon {
   vertical-align: middle;
 }
 
-.thumbnail-wrapper {
-  min-width: 120px;
-}
-
 .word-break {
   word-break: break-word;
 }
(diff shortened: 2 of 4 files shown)


tdonohue:
@bram-atmire : I'm not sure I agree with this PR as it removes the thumbnail entirely from the `ds-bitstream-attachment`.

My understanding is that this came over from DSpace-CRIS because it's a *part* of the Configurable Layout of Entity/Item pages coming in DSpace 11, see https://github.com/DSpace/DSpace/issues/11779

I *believe* the thumbnail appears in this component because it does NOT always appear in the other location, based on how the layout is configured.

So, I think it'd be a *better* solution to make this configurable... *or* wait until https://github.com/DSpace/DSpace/issues/11779 is fixed before we move this forward.

PR #5856, opened 2026-06-18
Make the “New collection” menu option visible for Community Admins

toniprieto:
## References
* Reported in https://groups.google.com/g/dspace-tech/c/wZ3AaF4lJAQ
* Related to https://github.com/DSpace/dspace-angular/pull/3994

## Description
Currently, visibility of the “New Collection” menu option is restricted to Collection Administrators, but it should instead be restricted to Community Administrators.

## Instructions for Reviewers
Verify that the issue reported in the linked message is no longer reproducible. In particular, confirm that Community Administrators who do not have permissions to administrate a collection can now see the “New Collection” option in the menu.

## Checklist

### DIFF
--- a/src/app/shared/menu/providers/new.menu.spec.ts
+++ b/src/app/shared/menu/providers/new.menu.spec.ts
@@ -38,7 +38,7 @@ describe('NewMenuProvider', () => {
       },
     },
     {
-      visible: false,
+      visible: true,
       model: {
         type: MenuItemType.ONCLICK,
         text: 'menu.section.new_collection',
--- a/src/app/shared/menu/providers/new.menu.ts
+++ b/src/app/shared/menu/providers/new.menu.ts
@@ -54,12 +54,11 @@ export class NewMenuProvider extends AbstractExpandableMenuProvider {
 
   public getSubSections(): Observable<PartialMenuSection[]> {
     return combineLatest([
-      this.authorizationService.isAuthorized(FeatureID.IsCollectionAdmin),
       this.authorizationService.isAuthorized(FeatureID.IsCommunityAdmin),
       this.authorizationService.isAuthorized(FeatureID.AdministratorOf),
       this.authorizationService.isAuthorized(FeatureID.CanSubmit),
       this.authorizationService.isAuthorized(FeatureID.CoarNotifyEnabled),
-    ]).pipe(map(([isCollectionAdmin, isCommunityAdmin, isSiteAdmin, canSubmit, isCoarNotifyEnabled]: [boolean, boolean, boolean, boolean, boolean]) => {
+    ]).pipe(map(([isCommunityAdmin, isSiteAdmin, canSubmit, isCoarNotifyEnabled]: [boolean, boolean, boolean, boolean]) => {
 
       return [
         {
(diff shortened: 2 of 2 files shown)


Kasinhou:
Confirmed working locally at dataquest, works as described.

Screenshots
Before:
image

After: 
image

milanmajchrak (approved):

tdonohue (approved):
👍 Thanks @toniprieto !  The code looks good...and looks like others have tested it. Merging immediately and backporting.

eulereadgbe:
Thank you for this fix! I confirmed that PR [#5856](https://github.com/DSpace/dspace-angular/pull/5856) successfully resolves the issue where Community Administrators could not create a collection in an empty community.

However, while testing this PR on my local install (DSpace 9.3 with the commits applied) and on sandbox.dspace.org (running DSpace 11.0-SNAPSHOT), I encountered a critical follow-up permission issue immediately after creation.

**Steps to Reproduce:**
1. As Super Admin, create a new EPerson and a new Community, assigning the EPerson as the Community Administrator.
2. Impersonate/log in as the new Community Administrator.
3. Create a new collection under that community (which now works thanks to PR #5856).
4. Attempt to edit or manage the newly created collection.

Expected Result: The Community Administrator should be able to edit and manage the collection they just created under their community.

Actual Result: The user is met with a 403 Forbidden error page (/403).

**Environment Notes:**
- Replicated on: Local DSpace 9.3 (with PR #5856 patch applied), production server, and sandbox.dspace.org (DSpace 11.0-SNAPSHOT).
- Not replicated on: demo.dspace.org (DSpace 10.1-SNAPSHOT).

It seems that while creation rights are now unlocked, the creation process is either not adding the Community Admin to the newly spawned Collection's Admin group (COLLECTION_<ID>_ADMIN) or not granting the appropriate resource policy tokens upon creation.

**Update / Additional Finding:**

Upon further testing, I identified the root cause of the 403 Forbidden error immediately following collection creation: it appears to be a timing/race condition issue between collection creation and authorization indexing.

**Behavior observed:**

- If a Community Administrator creates a collection and is immediately redirected to (or attempts to access) the Edit Collection page right after saving, DSpace returns a 403 Forbidden page (/403).
- However, if the user waits a few seconds, refreshes, or navigates away and clicks "Edit Collection" from the sidebar menu later, the page loads successfully and permissions work as expected.

**Conclusion:**
While the backend policies are being applied correctly, there is an asynchronous delay/indexing lag before the new collection's permissions become active. The frontend appears to check permissions before the backend finishes propagating them, resulting in an unhandled 403 error.

Reporting this so the frontend/router can potentially handle this race condition more gracefully (e.g., waiting for authorization indexing or handling transient 403s on newly created objects) rather than displaying a hard forbidden screen to end users.

toniprieto:
Hi @eulereadgbe , 
I think the issue you're reporting makes sense, although I suspect it may not actually be related to indexing.

My guess is that, before creating the collection, the frontend makes an authorization request to check whether the current user is a Collection Administrator. That request is then cached, and the result (false) may still be used by the UI after the collection has been created, causing the "Edit Collection" option/route to behave as if the user still doesn't have the required permissions.

It might be possible to fix this by invalidating the authorization request cache immediately after creating the collection, for example with:

`this.authorizationService.invalidateAuthorizationsRequestCache();`

Unfortunately, I don't have time to test this hypothesis at the moment, but I'm leaving the comment in case it helps narrow down the issue.

I think it would also be a good idea to open a separate issue describing this behavior. That way, other developers can investigate and hopefully confirm whether this is indeed a frontend authorization-cache issue and work on a proper fix.

PR #5859, opened 2026-06-20
Bump @angular/* framework packages to 20.3.25 (combines split security PRs #5850/#5851/#5852)

bram-atmire:
## References

Supersedes and resolves the three split Dependabot security PRs:
- Closes #5850 (`@angular/core`)
- Closes #5851 (`@angular/compiler`)
- Closes #5852 (`@angular/common`)

## Description

Dependabot raised the Angular `20.3.24 -> 20.3.25` security update as three separate per-package PRs (security updates do not honor the `@angular*` grouping configured in `.github/dependabot.yml`, which only applies to version updates). Each one fails CI at `npm clean-install` with `ERESOLVE`, because Angular peer dependencies require every `@angular/*` framework package to be the exact same version, and bumping one in isolation leaves the rest at `20.3.24`.

This PR bumps the whole peer-locked framework family together so the update can actually install and pass CI:

`@angular/animations`, `@angular/common`, `@angular/compiler`, `@angular/core`, `@angular/forms`, `@angular/localize`, `@angular/platform-browser`, `@angular/platform-browser-dynamic`, `@angular/platform-server`, `@angular/router`, and `@angular/compiler-cli` -> `^20.3.25`.

`@angular/compiler-cli` is included because it has an exact peer dependency on `@angular/compiler`, so it must move in lockstep (leaving it behind reproduces the same `ERESOLVE`). Independently versioned packages (`@angular/cdk`, `@angular/cli`, `@angular/ssr`) are left untouched.

### Security advisories resolved (all fixed in 20.3.25)

| Advisory | Package | Severity |
|----------|---------|----------|
| GHSA-rgjc-h3x7-9mwg | `@angular/core` (hydration DOM clobbering & response-cache poisoning) | High |
| GHSA-39pv-4j6c-2g6v | `@angular/common` (weak 32-bit cache key in `HttpTransferCache`, cross-request data leakage) | High |
| GHSA-48r7-hpm6-gfxm | `@angular/common` (DoS via OOM in `formatDate`) | High |
| GHSA-58w9-8g37-x9v5 | `@angular/compiler` (two-way binding sanitization bypass / XSS) | Medium |

## Instructions for Reviewers

`package.json` changes are limited to the eleven Angular packages above. The `package-lock.json` diff is those version bumps plus a few in-range transitive patch refreshes in the mirador/react subtree (`react-rnd`, `notistack`, `goober`, `clsx`) that npm reconciles automatically; there are no major or out-of-range changes. Verified locally with `npm ci` (exit 0), which is the exact command CI runs.

## Checklist

- [x] My PR is small in size, or I have provided reasons as to why that's not possible (lock file changes are auto-generated).
- [x] My PR passes ESLint validation.
- [x] My PR doesn't introduce new dependencies (version bump of existing ones only).
- [x] My PR is created against the `main` branch.


tdonohue (approved):
👍 Thanks @bram-atmire !  Deployed this locally and verified it works well. 

Will attempt to auto-port this to 10.x and 9.x branches, as both of them still have Angular v20.3.24

tdonohue:
@bram-atmire : This was able to be auto-backported to 10.x, but failed to auto-backport to 9.x.  If you would have time to backport this manually to `dspace-9_x`, I think it'd be best to ensure it gets to that branch as well.

tdonohue:
Backported to 9.x in #5958

PR #5882, opened 2026-06-30
Fix/5524 collection selector hide communities

### LINKED ISSUE #5524: Administrator reports + Metadata Queries: Communities appear in the Collection Selector but can not be selected
## Describe the bug
When using the Metadata Query feature of the Administrator report, a list of both collections and communities is shown in the "Collection selector" drop-down menu. The communities can not be selected, and it appears as if you cannot select anything when a lot of communities are shown at the beginning of the list.
Observed on Sandbox during Testathon (10.0) as well as on Demo (9.x)

Image

## To Reproduce
Login on the sandbox site as an admin
1. Open the Reports -> Metadata query
2. Try to select something under "Collection Selector"

## Expected behavior
I would assume no communities should be shown in the collection selector list. The feature documentation can be found here: https://wiki.lyrasis.org/pages/viewpage.action?pageId=408945575

## Related work
Kent Gerber and Marcella Huggard initially reported this issue in the testathon test document.

### LINKED ISSUE #5883: Administrator Reports - Metadata Query: "Whole Repository" and specific collections can be selected simultaneously in Collection Selector
## Describe the bug

In the Administrator Reports > Metadata Query page, the "Whole Repository" option in the Collection Selector can be selected simultaneously with one or more specific collections. While the query logic appears to apply only to the specific collections selected, this behavior is visually confusing and semantically contradictory — selecting "Whole Repository" implies no collection filter should apply, yet specific collections can remain selected at the same time.

Observed on demo.dspace.org and sandbox.dspace.org using **Google Chrome**.

Image

## To Reproduce

1. Log in as an admin
2. Navigate to Reports → Metadata Query
3. Open the Collection Selector panel
4. Click "Whole Repository" to select it
5. Without deselecting it, click one or more specific collections
6. Both "Whole Repository" and the specific collection(s) appear selected simultaneously

## Expected behavior

Selecting "Whole Repository" should automatically deselect any previously selected specific collections, and selecting any specific collection should automatically deselect "Whole Repository". The two options represent mutually exclusive filtering scopes and should not be selectable at the same time.

## Related work

Identified while working on #5524 (Administrator reports + Metadata Queries: Communities appear in the Collection Selector but cannot be selected).

shernandez-dev:
## References
* Fixes #5524
* Fixes #5883

## Description
Replaces the flat `<select>` list in the Metadata Query Report Collection Selector with a grouped tree-style selector that shows communities as non-selectable headers and collections as clickable options, eliminating the confusion caused by unselectable community entries mixed with collections.

### Root Cause
The `loadCollections()` method in `FilteredItemsComponent` built a flat `OptionVO[]` array by mixing communities (marked `disabled=true`) and collections together in a single `<select multiple>` element. Communities appeared as entries that users could not select, and when many communities appeared at the top of the list, it was unclear whether anything could be selected at all.

### Changes

* **`filtered-items.component.ts`**:
  - Added `CollectionGroup` interface to group collections under their parent community
  - Added `collectionGroups: CollectionGroup[]` property
  - Refactored `loadCollections()` to no longer push communities as disabled `<option>` entries — instead, each community becomes a group header with its collections as selectable items
  - Added `toggleCollection(collectionId: string)` method to handle multi-selection via click
  - Added `isCollectionSelected(collectionId: string)` method to track selected state
  - `collections: OptionVO[]` is still maintained for form compatibility with `toQueryString()`
   - Updated `toggleCollection()` to enforce mutual exclusion between "Whole Repository" and specific collections — selecting one automatically deselects the other (#5883)

* **`filtered-items.component.html`**:
  - Replaced `<select multiple>` with a custom list using Bootstrap `list-group-item` buttons
  - Communities rendered as `text-muted small` non-interactive headers
  - Collections rendered as `fw-bold` clickable buttons with `bg-primary` highlight on selection
  - Follows the same visual pattern used in the existing Edit Item / Move Item collection selector (`dso-selector`), as suggested in the issue discussion. A full reuse of the `dso-selector` component was considered but not pursued, as that component is designed for single selection and would require significant refactoring to support multi-selection. Instead, the same visual hierarchy (community as breadcrumb context, collection as bold selectable item) was replicated using standard Bootstrap `list-group-item` buttons, keeping the change isolated to this component.

* **`filtered-items.component.spec.ts`**:
  - Added 14 unit tests covering `loadCollections()`, `toggleCollection()`, and `isCollectionSelected()`

### Before
Communities and collections mixed in a flat `<select>` list. Communities appeared as unselectable entries, making it unclear what could be selected:

Before

### After
Communities shown as non-interactive group headers (grey, small text). Collections shown as bold clickable buttons with highlight on selection:

image

> Note: screenshot taken from a local instance in Spanish. Collection names and community names will appear in the repository's configured language.

### How to Test
1. Log in as an admin
2. Navigate to Reports → Metadata Query
3. Open the Collection Selector panel
4. Verify communities appear as non-clickable grey headers
5. Verify collections appear as bold clickable buttons below each community
6. Click one or more collections and verify they highlight correctly
7. Run a query with selected collections and verify results are filtered correctly

Reproduced and confirmed on **DSpace 9.x** and **DSpace 10.0** (as reported in the original issue).

## Checklist
- [x] My PR is **created against the `main` branch**
- [x] My PR is **small in size** (3 files modified)
- [ ] My PR **passes ESLint** — `npm run lint` fails with a pre-existing error unrelated to this fix (`src/themes/custom/eager-theme-components.ts`). ESLint was run directly on the 3 modified files and passes without errors.
- [x] My PR **doesn’t introduce circular dependencies** (verified via `npx madge --circular` on the affected directory)
- [x] My PR **includes TypeDoc comments** for new methods
- [x] My PR **passes all specs/tests and includes new/updated specs**
- [x] My PR **includes details on how to test it**
- [x] My PR **uses i18n keys** (no hardcoded English text added)


lgeggleston:
@shernandez-dev Thanks for your contribution, and identifying the [secondary issue](https://github.com/DSpace/dspace-angular/issues/5883) on this component!

@tinsch Since you reported the [original issue](https://github.com/DSpace/dspace-angular/issues/5524) about the collection selector, would you by chance be interested in testing this to confirm if this addresses the usability concerns appropriately?

tinsch:
@lgeggleston yes I am happy to test this 👍

tinsch:
@shernandez-dev thanks for the fix! I tested it and the UX is much better now. The tree structure between communities and collections is visible and it is clear only collections are selectable. The toggle between "Whole repository" und collections also works well 👍

MMilosz:
While the grouped presentation is much clearer than the previous flat list. However, I noticed one usability regression in the new design

The previous `<select multiple>` allowed users to select large numbers of collections using:
- Shift-click
- Ctrl/Cmd-click
- Mouse drag selection
- Keyboard selection (accessibility!)

With the new button-based selector, we lose all of these and can only click every collection **individually**, which feels like a significant usability regression

## Before

Tested on demo.dspace.org

[before.webm](https://github.com/user-attachments/assets/ffb892c9-5883-4315-a539-3a9dd6e6a8b2)

## After

[after.webm](https://github.com/user-attachments/assets/551c6d80-79ef-4042-b4e8-8b9def5df418)

shernandez-dev:
@MMilosz Thank you for the feedback — you're absolutely right, and this is a valid concern.

The reason these behaviors were lost in the current implementation is that the button-based list replaces the native `<select multiple>` element entirely, so the browser's built-in multi-selection mechanics (Shift-click range, Ctrl/Cmd-click toggle, drag selection, and keyboard navigation) no longer apply — those are handled natively by the browser only when a `<select multiple>` is present.

I've been looking into this and the solution is to revert to a native `<select multiple>` element, but using `<optgroup>` elements to render the community/collection hierarchy. This preserves all native browser behaviors while fixing the original issue of communities appearing as selectable options.

The visual result would be:
- Communities rendered as `<optgroup label>` — bold, non-selectable, native browser separator
- Collections rendered as `<option>` inside each group — fully selectable with all native multi-select behaviors

I already have this approach working locally. If this sounds like the right direction, I can update the PR with the tested changes. Any observations or recommendations on the approach are also welcome.

PR #5934, opened 2026-07-06
Add ESLint rule to prevent Observable-returning method calls in the async pipe (#5933)

### LINKED ISSUE #5933: Introduce ESLint rule to prevent Observable-return methods from being used in Angular templates
We commonly use this pattern in DSpace Angular:
- Component method `returnSomething$()` returns an Observable
- Called from the template, with `| async`

Effectively this causes the RxJS pipe to be executed again for each change detection cycle, often leading to performance issues.

We should look into introducing an ESLint rule to guard against this.

Existing cases should be rewritten into the following pattern:
- Subscribe within `ngOnInit()`
- Pass emitted values to a `BehaviorSubject`; this will be the new target for `| async`
- Unsubscribe within `ngOnDestroy()`

Aryakoste:
## References
* Fixes #5933 

We commonly call a component method directly as the input of the `async` pipe in templates, e.g. `{{ getFoo$() | async }}`. Because the method runs as part of template evaluation, Angular re-executes the whole RxJS pipeline (subscribe + unsubscribe) on every change-detection cycle instead of once — a real, recurring perf issue in this codebase (confirmed instances in `audit-table`, `ePerson-data`, and `section-coar-notify` components).

This PR adds a new `dspace-angular-html` ESLint rule, `no-method-call-with-async-pipe`, that flags this pattern anywhere it appears in a template — interpolations, bound attributes/events, `@if`/`@for`/`@switch` blocks, `@defer` triggers, `@let`, and object/array literals or optional-chaining wrapped around the pipe.

- New rule: `lint/src/rules/html/no-method-call-with-async-pipe.ts`
- Registered in `lint/src/rules/html/index.ts` and enabled in `.eslintrc.json`
- Docs auto-generated via `npm run docs:lint` (`docs/lint/html/index.md`, `docs/lint/html/rules/no-method-call-with-async-pipe.md`)

### Implementation note

The straightforward approach — an ESLint selector like `` BindingPipe[name="async"] > Call `` — turned out to miss common real-world cases, because `@angular-eslint/template-parser`'s own visitor-keys map doesn't cover every expression AST node type (e.g. `SafePropertyRead`, `LiteralMap`), so ESLint's traversal silently stops before reaching calls wrapped in things like `(getFoo$() | async)?.length` or `[ngClass]="{x: getFoo$() | async}"`. To get full, correct coverage, the rule instead walks each template expression with `@angular/compiler`'s own `RecursiveAstVisitor`, which handles every AST node type regardless of that gap.

This rule does not attempt to fix existing offending call sites — per the issue, migrating those (subscribe in `ngOnInit()`, push into a `BehaviorSubject`, `| async` on that, unsubscribe in `ngOnDestroy()`) is a separate follow-up. The rule is not auto-fixable, since that migration can't be done mechanically.

## Checklist

### DIFF
--- a/lint/src/rules/html/no-method-call-with-async-pipe.ts
+++ b/lint/src/rules/html/no-method-call-with-async-pipe.ts
@@ -1,0 +1,233 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+import {
+  AST,
+  BindingPipe,
+  Call,
+  PropertyRead,
+  RecursiveAstVisitor,
+  SafeCall,
+  TmplAstBoundAttribute,
+  TmplAstBoundDeferredTrigger,
+  TmplAstBoundEvent,
+  TmplAstBoundText,
+  TmplAstForLoopBlock,
+  TmplAstIfBlockBranch,
+  TmplAstLetDeclaration,
+  TmplAstSwitchBlock,
+  TmplAstSwitchBlockCase,
+} from '@angular-eslint/bundled-angular-compiler';
+import { ESLintUtils } from '@typescript-eslint/utils';
+import {
+  RuleContext,
+  RuleListener,
+} from '@typescript-eslint/utils/ts-eslint';
+
+import {
+  DSpaceESLintRuleInfo,
+  NamedTests,
+} from '../../util/structure';
+import { getSourceCode } from '../../util/typescript';
+
+export enum Message {
+  NO_METHOD_CALL_WITH_ASYNC_PIPE = 'noMethodCallWithAsyncPipe',
+}
+
+export const info = {
+  name: 'no-method-call-with-async-pipe',
+  meta: {
+    docs: {
+      description: `Don't call a method directly as the input of the \`async\` pipe (e.g. \`getFoo$() | async\`).
+
+This re-executes the method - and the RxJS pipeline it returns - on every change detection cycle, which often leads to performance issues.
+
+Instead:
+- Subscribe to the Observable in \`ngOnInit()\`
+- Push emitted values into a \`BehaviorSubject\`
+- Bind \`| async\` to that \`BehaviorSubject\` in the template
+- Unsubscribe in \`ngOnDestroy()\`
+      `,
+    },
+    type: 'problem',
+    schema: [],
+    messages: {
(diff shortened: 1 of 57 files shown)


lgeggleston:
Thank you for the PR @Aryakoste! You may have already seen this, but looks like this is failing a few build tests that will need to be addressed.

lint/src/rules/html/no-method-call-with-async-pipe.ts
```diff
@@ -0,0 +1,233 @@
...
+/**
+ * @angular-eslint/template-parser's own visitor keys (used to drive ESLint's node traversal)
+ * don't cover every expression AST node type (e.g. SafePropertyRead, LiteralMap, KeyedRead, ...),
+ * so selectors like `BindingPipe[name="async"] > Call` silently miss anything wrapped in one of
+ * those unsupported node types (e.g. `(getFoo$() | async)?.length`, or `[ngClass]="{x: getFoo$() | async}"`).
+ * We work around this by walking each expression ourselves with the Angular compiler's own
+ * RecursiveAstVisitor, which correctly covers the full expression AST regardless of what
+ * @angular-eslint/template-parser exposes to ESLint's selector engine.
+ */
+class AsyncPipedCallVisitor extends RecursiveAstVisitor {
```
ybnd, line 75: Good catch on those selectors, I wasn't aware of that!

ybnd (changes requested):
↑ those CI failures are due to the new lint rule

> This rule does not attempt to fix existing offending call sites — per the issue, migrating those (...) is a separate follow-up. 
> The rule is not auto-fixable, since that migration can't be done mechanically.

It won't be possible to auto-fix, but **we'll still need to sort out that "follow up step" within this PR** as we did in previous PRs that introduced new configuration or rules.

Aryakoste:
Ok will do that

Aryakoste:
@ybnd @lgeggleston I have done the migration of all the files.

tdonohue:
@Aryakoste : You may have seen this already, but it looks like this PR is failing ~12 unit tests.  We'd obviously need those to pass before we can look for additional reviewers.

Aryakoste:
@tdonohue Yeah i have solved those pushing them

ybnd:
I haven't had the time to check the latest changes, so I'm adding @alexandrevryghem as a reviewer while I'm on holiday 🫥

FWIW the ESLint rule changes look good to me, the only thing that remains is checking that the issues it caught were resolved correctly.

PR #5936, opened 2026-07-08
Fix infinite route-loader spinner from per-navigation root cache inva…

### LINKED ISSUE #5855: Infinite route-loader spinner: per-navigation invalidateRootCache() deadlocks getEndpointMapAt when the root endpoint goes stale
## Describe the bug

After the UI has been open a while, navigating to / revisiting a page can leave the app stuck on the **route-loader spinner** (`<ds-loading>` in `ds-base-root`, driven by `AppComponent.isRouteLoading$`) forever. The shell stays responsive (navbar, sidebar, overlays work); a full reload or clicking any other in-app link recovers it, but the routed content never renders. While stuck: **nothing is pending on the network, the NgRx store is frozen, and CPU/memory are idle** — a logical deadlock, not a loop or leak. It is much harder to reproduce with DevTools open and far worse in production than on a local dev server (cf. #3584, #3697).

## To Reproduce

Authenticated, ideally behind a reverse proxy / with non-trivial REST latency:
1. Load a content page (e.g. a Community/Collection).
2. Navigate around / wait a bit.
3. Navigate back (or keep navigating). The spinner appears and never resolves; reload or any other link recovers it. Higher REST latency → more reliable (essentially every revisit through a proxy).

## Expected behavior

Navigation completes and the page renders; a transiently-stale root endpoint must not wedge navigation indefinitely.

## Root cause

1. `BrowserInitService.listenForRouteChanges()` calls `rootDataService.invalidateRootCache()` on **every** `NavigationStart` → `RequestService.setStaleByHref('/server/api')` marks the root request **stale** (regardless of its TTL).
2. `HALEndpointService.getEndpointMapAt()` — used by `getEndpoint()`, which every data request needs — does `filter(rd => !rd.isStale)`, **discarding the stale root**; its `tap(... getEndpointMapAt())` re-request does not reliably produce a fresh one, so it **never emits**.
3. So `getEndpoint()` never resolves → the route resolver never completes → the router never emits `NavigationEnd` → `isRouteLoading$` (cleared only on `NavigationEnd`/`Cancel`) stays `true` → the route loader spins forever, store frozen.

**Evidence (frozen store at the hang):** root `/server/api` is `SuccessStale` with TTL **not** expired (explicitly invalidated, not aged out); **0 requests pending** (nothing re-fetching); the only `RequestStaleAction` caller is `setStaleByHref` ← `invalidateRootCache` ← the `NavigationStart` subscription. Consistent with #2669 (`invalidateRootCache` ↔ `HALEndpointService` stale interaction) after #2510 (7.6.1) made HAL endpoint requests stale-able.

## Proposed fix

The root endpoint map is static between navigations, so the per-`NavigationStart` invalidation is unnecessary. Removing it eliminates the deadlock (backend availability is still established at app init and surfaces through normal request failures):

```diff
 import {
-  NavigationStart,
   Router,
 } from '@angular/router';
@@ listenForRouteChanges()
   protected listenForRouteChanges(): void {
-    // we'll always be too late for the first NavigationStart event with the router subscribe below,
-    // so this statement is for the very first route operation.
+    // Invalidate the root endpoint cache once, for the very first route operation.
     this.rootDataService.invalidateRootCache();
-
-    this.router.events.pipe(
-      filter(event => event instanceof NavigationStart),
-    ).subscribe(() => {
-      this.rootDataService.invalidateRootCache();
-    });
   }
```

(Alternative, if the per-navigation backend check is to be kept: make `getEndpointMapAt` tolerate a stale-but-completed root, e.g. `filter(rd => !rd.isStale || rd.hasCompleted)`, so a stale endpoint map is used while the `tap()` refreshes it — fixing the deadlock without changing the invalidation cadence.)

Verified: the change removes the hang and the E2E suite is unaffected.

## Environment

- dspace-angular 9.3 (Angular 20.3); `listenForRouteChanges()`/`getEndpointMapAt()` are unchanged on `main` as of this report.
- DSpace 9.3 REST backend behind an nginx reverse proxy; CSR (SSR disabled); reproduced in Safari and Firefox.

## Related

#2669, #2510, #3584, #3697, #3888

naomigassler:
# PR: Fix infinite route-loader spinner on navigation (stale root-endpoint cache deadlock)

**Target branch:** `DSpace/dspace-angular:main`
**Relates to:** #3584, #3697
Fixes https://github.com/DSpace/dspace-angular/issues/5855

## Description

`BrowserInitService` invalidates the root API endpoint cache on **every**
`NavigationStart`. On a subsequent request, `HALEndpointService.getEndpointMapAt`
(hit on essentially every request via `getEndpoint()`) discards the now-stale
root `/server/api` entry and triggers a re-fetch that can **deadlock** — the
route resolver never completes, `NavigationEnd` never fires, and the
`ds-base-root` route-loader spinner hangs indefinitely on a frozen store.

The root endpoint map is effectively static between navigations, so
re-invalidating it on each `NavigationStart` is unnecessary. Backend-down
detection still happens at init and through normal request-failure handling.

## Root cause

- `BrowserInitService` → `invalidateRootCache()` on every `NavigationStart`.
- Marks the root `/server/api` endpoint cache stale.
- Next `getEndpoint()` → `getEndpointMapAt` discards the stale root → re-fetch
  deadlocks → resolver never resolves → `NavigationEnd` never fires → spinner
  hangs forever.
- Higher REST latency makes it fire on nearly every revisit (why it looks
  intermittent).

## Steps to reproduce

1. Open any listing (collection items, MyDSpace, Browse, Search — even
   Communities & Collections).
2. Navigate listing → listing → listing.
3. The UI hangs on the route-loader spinner; a full page reload clears it.

Reproducible on a small instance (~7k items) — **not** load- or scale-dependent.
Higher REST/proxy latency increases the frequency.

## Fix

Remove the per-`NavigationStart` `invalidateRootCache()` call in
`browser-init.service.ts`; keep the one-time invalidation at init. No behavior
change to backend-down detection.

## How to test

1. Before the fix: reproduce the hang via the steps above.
2. After the fix: the same navigation sequence no longer hangs; the spinner
   always resolves.
3. Backend-down handling still works: stop the REST backend and confirm the
   app still detects/handles it (init-time + request-failure paths).

## Tests

Added a spec asserting the root endpoint cache is invalidated **once at init**
and **not** on subsequent `NavigationStart` events.

## Checklist notes

- One-file logic fix; no new dependencies; no user-facing strings (i18n N/A).
- Passes `yarn lint` and `yarn check-circ-deps`.
- TypeDoc added/updated on any modified public method.

### DIFF
--- a/src/modules/app/browser-init.service.spec.ts
+++ b/src/modules/app/browser-init.service.spec.ts
@@ -1,0 +1,44 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+import { NavigationStart } from '@angular/router';
+import { Subject } from 'rxjs';
+
+import { BrowserInitService } from './browser-init.service';
+
+describe('BrowserInitService', () => {
+  describe('listenForRouteChanges', () => {
+    let service: BrowserInitService;
+    let rootDataServiceSpy;
+    let routerEvents$: Subject<any>;
+
+    beforeEach(() => {
+      rootDataServiceSpy = jasmine.createSpyObj('rootDataService', ['invalidateRootCache']);
+      routerEvents$ = new Subject();
+
+      service = new BrowserInitService(
+        null, null, null, null, null, null, null, null, null, null, null, null, null, null,
+        rootDataServiceSpy,
+        { events: routerEvents$.asObservable() } as any,
+        null, null, null, null,
+      );
+
+      (service as any).listenForRouteChanges();
+    });
+
+    it('should invalidate the root endpoint cache once, at init', () => {
+      expect(rootDataServiceSpy.invalidateRootCache).toHaveBeenCalledTimes(1);
+    });
+
+    it('should not invalidate the root endpoint cache again on subsequent NavigationStart events', () => {
+      routerEvents$.next(new NavigationStart(1, '/some-route'));
+      routerEvents$.next(new NavigationStart(2, '/another-route'));
+
+      expect(rootDataServiceSpy.invalidateRootCache).toHaveBeenCalledTimes(1);
+    });
+  });
+});
--- a/src/modules/app/browser-init.service.ts
+++ b/src/modules/app/browser-init.service.ts
@@ -10,10 +10,7 @@ import {
   Injectable,
   TransferState,
 } from '@angular/core';
-import {
-  NavigationStart,
-  Router,
-} from '@angular/router';
+import { Router } from '@angular/router';
 import {
   APP_CONFIG,
   APP_CONFIG_STATE,
@@ -226,21 +223,18 @@ export class BrowserInitService extends InitService {
   }
 
   /**
-   * Listen to all router events. Every time a new navigation starts, invalidate the cache
-   * for the root endpoint. That way we retrieve it once per routing operation to ensure the
-   * backend is not down. But if the guard is called multiple times during the same routing
-   * operation, the cached version is used.
+   * Invalidate the cache for the root endpoint once, at startup, so the first request for it
+   * is guaranteed to hit the backend rather than serve a value cached before the app booted.
+   *
+   * This used to also run on every `NavigationStart`, but the root endpoint map does not change
+   * between navigations, so that repeated invalidation was unnecessary. Worse, it could mark the
+   * root request stale while {@link HALEndpointService} was still resolving it for an in-flight
+   * data request; the stale value is filtered out of the endpoint map and the retry for it can
+   * lose the race with the next invalidation, so `getEndpoint()` never resolves and the route
+   * that depends on it never finishes loading.
    */
   protected listenForRouteChanges(): void {
-    // we'll always be too late for the first NavigationStart event with the router subscribe below,
-    // so this statement is for the very first route operation.
     this.rootDataService.invalidateRootCache();
-
-    this.router.events.pipe(
-      filter(event => event instanceof NavigationStart),
-    ).subscribe(() => {
-      this.rootDataService.invalidateRootCache();
-    });
   }
 
 }


tinsch:
I _was_ able to reproduce the issue with the DSpace sample data. The steps to reproduce it locally were:
1. login as admin
2. start a simple search on the whole repo (hit search on front page)
3. click on DCAT Journal Publications
4. go to "MyDSpace" listing
5. observe infinite loading on the list

I tested this PR, and the issue did not occur anymore 🎉. But then again, I tried to reproduce the original issue on main, just to double check - and could not successfully reproduce it anymore. So I would suggest:
- more people should try to reproduce the original issue and confirm that this PR fixes it
- a frontend dev could review the code and based on that we can decide to merge this PR, even if the original bug is not reproducible every time

Thanks for the PR! This is a great contribution in my opinion.

jlipka:
Hey @tinsch and others in here.
Thanks for listing your click sequence—unfortunately, I wasn't able to reproduce the effect using that, but I've included a similar attempt below.

Starting point:
The code in the  `main` branch without any changes from this MR!

I see the infinite route loader appear when I double-click very quickly on a navigation item.
You can also reproduce this in the DSpace Sandbox.

1. Open https://sandbox.dspace.org
2. Hover over “All of DSpace” and double-click on “By title”
3. As a result, you’ll see the loading spinner centered in the middle (it no longer disappears).
4. If that didn’t work right away, double-click again on, for example, “By author”
(You should see the effect after a few attempts at most.)

[Screencast on the DSpace sandbox page](https://github.com/user-attachments/assets/c63b6b75-d2ba-4904-88e1-7c683df934d7)
(on the second attempt the spinner appears)


The issue can apparently be reproduced this way—unless I’m missing something.
I used Firefox running on an Ubuntu/Linux.

**If I now apply the fix from the merge request, the error no longer occurs. Therefore, in my opinion, the fix makes perfect sense—especially since the logic in the underlying code apparently isn’t needed at all.**

--- 
So far, so good.

But now I’m wondering if this really solves the problem. Even if we probably don’t actually need to manually invalidate the `/server/api` API endpoint on every `NavigationStart` event, invalidating requests is still a common pattern in DSpace (at this point, I’d just like to loosely refer to the `msToLive` property in the configuration). Other requests will also become invalid or stale over the course of the application’s lifecycle, and the app shouldn’t freeze up in such cases either.

So, another approach or at least a try could be to swap the order of the following two lines:
Original code in `findbyHref` in `base-data.service.ts` 

https://github.com/DSpace/dspace-angular/blob/a52ba244fa454cf20b65ab23a274d98080d9f1ee/src/app/core/data/base/base-data.service.ts#L312-L314 

We could move the `skipWhile` operator below the `reRequestStaleRemoteData` sideffect, as its a regular tap operator wrapped inside another function.
- Could this be a valid (additional) solution?
- Does `skipWhile` still serve its purpose, or has the logic been modified in an unintended way?

**No matter what sequence of clicks I used after making this change, I could no longer trigger the state where the loading spinner is displayed continuously.**


**Long story short:**
- I think the change in this MR makes sense, and I can reproduce it myself using the sequence of clicks I described above
- At the same time, in my opinion, the fix does not resolve a potential issue in BaseDataService that might be blocking streams unintentionally—because if the `skipWhile` mechanism is moved, the freezes won’t occur, regardless of whether we use the fix in this merge request or not. However, it hasn't yet been determined whether this would cause problems elsewhere in the app.
- I’m aware that the calculated Boolean for the LoadingIndicator is retrieved in `app.component.ts`. We could certainly make adjustments here as well, but this again raises the question for me of whether this might be masking a potential (!—I’m not sure) problem elsewhere.

I’d be very interested in hearing your thoughts on these points, and of course whether you can reproduce the behavior.

jlipka:
After further investigation, I might have found a better, more logic solution, which leaves my previous idea seeming obsolete.

--- 

Original code:
https://github.com/DSpace/dspace-angular/blob/a52ba244fa454cf20b65ab23a274d98080d9f1ee/src/app/core/data/request.service.ts#L155-L176

My (partial) code changes:
```
const isValid = (entry: RequestEntry): boolean => {
  if (hasNoValue(entry)) {
    // undefined entries are invalid
    return false;
  } else {
    if (isStale(entry.state)) {
      return false;
    } else if (isLoading(entry.state)) {
      return true;
    } else {
      // check whether it should be stale
      const timeOutdated = entry.response.timeCompleted + entry.request.responseMsToLive;
      const now = new Date().getTime();
      const isOutDated = now > timeOutdated;
      return !isOutDated;
    }
  }
};
```

**This change targets to check `isStale` first (no matter if the loading state is currently true or false), before to check the loading state, and last to check the TTL.**

---

Another, additional check I added in the `shouldDispatchRequest` method in the same file (`request.service.ts`), starting around Line 493:

Original code:
https://github.com/DSpace/dspace-angular/blob/a52ba244fa454cf20b65ab23a274d98080d9f1ee/src/app/core/data/request.service.ts#L431-L436

My (partial) code changes:
```
    } else {
      // if we are, check the request cache
      const urlWithoutEmbedParams = getUrlWithoutEmbedParams(request.href);
      if (this.hasByHref(urlWithoutEmbedParams) === true) {
        return false;
      } else if (this.hasByHref(urlWithoutEmbedParams, false) === true) {
        // an entry exists but is stale/outdated -> always fetch fresh data,
        // regardless of whether the object is still present in the object cache
        return true;
      } else {
```

**This change ensures that a request entry that exists in the request cache but is stale or expired is always re-fetched from the server—regardless of whether the associated object is still present in the object cache. Previously, in this case, the system would incorrectly fall back on the object cache and suppress the request, even though the data was marked as stale.**

---

As before, the fix in this merge request makes perfect sense. My additional changes could make the app even more robust.

I have to do some more testing about these changes (especially in our customized instance)... but wanted to let you know about my current investigation.

tinsch:
Thanks @jlipka for your investigation. I can reproduce the spinner issue on sandbox with your click sequence, but not locally. 

I tried my click sequence again to reproduce the issue, and it happened reliably on main and was fixed on the PR branch. However, I looked into the PR code and tested it further: The problem is, we now don't invalidate the backend route anymore on NavigationStart. This means that if the backend is down (simulated locally by turning off the backend container) the frontend won't notice. I can still click around in the UI and get some errors on certain pages, but this will be confusing for end users who don't know what's happening when the UI still looks somewhat fine. It can lead to data loss on the users side, and a lot of questions to the repository owners should the backend be down. Unfortunately I think we cannot use the fix for this reason.

@naomigassler could you look into this please? Maybe there is a way to prevent the problem I described and still fix the spinner issues by tweaking the code a bit.
@jlipka would you be willing to open an additional PR that includes your fix, so we could discuss it separately?

To help with further investigations: 
When reproducing the spinner issue on the workspace page, I saw a 400 response from the backend every time the spinner issue occurred. 
Screenshot 2026-08-20 at 14 40 24
Screenshot 2026-08-20 at 14 40 50

The route was `http://localhost:8080/server/api/discover/search/objects?sort=lastModified,DESC&page=0&size=10&configuration=workspace&embed=thumbnail&embed=item%2Fthumbnail` and the errors in the backend indicate that the frontend does not properly send the login cookie on that request:

```
2026-08-20 12:45:03,898 ERROR unknown 66fe317a-003e-44f0-a1f5-1841e2c99fe7 org.dspace.discovery.utils.DiscoverQueryBuilder @ anonymous::Error in Discovery while setting up date facet range:date facet\colon; org.dspace.discovery.configuration.DiscoverySearchFilterFacet@38816a6c
org.dspace.discovery.SearchServiceException: An anonymous user cannot perform a workspace or workflow search
```

naomigassler:
Hello @tinsch and @jlipka,

<head></head><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;">We measured every candidate on a test box against two scenarios:<span class="Apple-converted-space"> </span><strong>A</strong><span class="Apple-converted-space"> </span>the<br>deadlock,<span class="Apple-converted-space"> </span><strong>B</strong><span class="Apple-converted-space"> </span>the backend stopped. Full write-up and raw data available if<br>useful.</p>

| variant | Scenario A (deadlock) | Scenario B (backend down) | cost |
|---|---|---|---|
| baseline | hangs 3/3 | `/500` every navigation, 338 ms | — |
| this PR | fixed 0/3 | no `/500`; inline "Error fetching …" at 353 ms; `/mydspace` blank, no error | none |
| our published alternative | **hangs 3/3** | = baseline | — |
| @jlipka's `isValid` reorder | **wedges the browser tab** | rejected | — |
| decoupled liveness probe | fixed 0/3 | = baseline (`/500` every navigation) | 1 root request per navigation |

<p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;">Measured on our DSpace 9.3 deployment with added REST latency, not on<span class="Apple-converted-space"> </span><code>main</code>. The<br><code>/mydspace</code><span class="Apple-converted-space"> </span>blank page in row 2 is reachable only with this PR applied — see below.</p><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;"><strong>A correction to our own issue report.</strong><span class="Apple-converted-space"> </span>The alternative we suggested there —<br><code>filter(rd =&gt; !rd.isStale || rd.hasCompleted)</code><span class="Apple-converted-space"> </span>in<span class="Apple-converted-space"> </span><code>getEndpointMapAt</code><span class="Apple-converted-space"> </span>—<span class="Apple-converted-space"> </span><strong>does not<br>work</strong>. It hangs 3/3 with a frozen-store signature byte-identical to the unfixed<br>baseline: the stale-but-completed<span class="Apple-converted-space"> </span><code>RemoteData</code><span class="Apple-converted-space"> </span>never reaches that filter, so the<br>deadlock is upstream of it. Please disregard it, and apologies to anyone who spent<br>time on it.</p><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;"><strong>On the objection.</strong><span class="Apple-converted-space"> </span>"Won't notice the backend is down" isn't quite what we<br>measured — the outage surfaces on the first navigation at 353 ms, as inline errors<br>on most routes. But there is no<span class="Apple-converted-space"> </span><code>/500</code>, the root stays<span class="Apple-converted-space"> </span><code>Success</code>/200 for its full<br>6 h TTL, and<span class="Apple-converted-space"> </span><code>/mydspace</code><span class="Apple-converted-space"> </span>renders an empty main area with no error at all. There a<br>user cannot distinguish an outage from an empty repository, and that is exactly<br>where the data-loss concern lives, because submissions are on that page.</p><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;">So the concern was sound even where the mechanism wasn't, and chasing it turned up<br>a pre-existing bug:<span class="Apple-converted-space"> </span><code>&lt;ds-search&gt;</code><span class="Apple-converted-space"> </span>renders nothing — no results, no empty state, no<br>error — when its search-configuration request fails from a cold cache. It backs 6<br>routes, is reproducible on<span class="Apple-converted-space"> </span><code>main</code><span class="Apple-converted-space"> </span>today, and is filed as #6111.</p><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;"><strong>But on unmodified<span class="Apple-converted-space"> </span><code>main</code><span class="Apple-converted-space"> </span>you cannot reach it by stopping the backend</strong>, because<br><code>ServerCheckGuard</code><span class="Apple-converted-space"> </span>redirects to<span class="Apple-converted-space"> </span><code>/500</code><span class="Apple-converted-space"> </span>first. This PR's deletion removes that<br>blanket, so it converts an unreachable blank page into a reachable one on<br><code>/mydspace</code>. That is a fair reading of the objection and we would rather state it<br>than have it found in review. The consequence is a merge order rather than a<br>rejection: with #6111 in, those 6 routes report the outage honestly and this PR has<br>no silent hole on search-backed pages. #6111 stands on its own either way.</p><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;"><strong>What the actual decision is.</strong><span class="Apple-converted-space"> </span>Today's<span class="Apple-converted-space"> </span><code>/500</code><span class="Apple-converted-space"> </span>is a<span class="Apple-converted-space"> </span><em>categorical</em><span class="Apple-converted-space"> </span>guarantee: every<br>route reports an outage. This PR replaces it with per-route handling — better where<br>implemented, worse where it isn't. Two coherent positions, and we don't think our<br>measurements settle which is right:</p><ol style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;"><li><strong>Per-route errors.</strong><span class="Apple-converted-space"> </span>This PR + #6111 + finish the remaining routes. Smallest<br>changes, better messages, but inherently whack-a-mole:<span class="Apple-converted-space"> </span><code>/community-list</code><span class="Apple-converted-space"> </span>still<br>sits on "Loading…" indefinitely with the backend down, on its own data path.</li><li><strong>Keep a categorical signal.</strong><span class="Apple-converted-space"> </span>Make liveness an explicit<span class="Apple-converted-space"> </span><code>HttpClient</code><span class="Apple-converted-space"> </span>probe that<br>never enters the ngrx store, keeping this PR's deletion. Scenario B came out<br>bit-for-bit identical to baseline, which is what proves the decoupling. Cost:<br>one root request per navigation (settle 293–847 ms vs this PR's 162–656 ms;<br>baseline pays the same), a new service plus a guard rewrite instead of deleting<br>five lines, and a de-duplication TTL we picked rather than derived.</li></ol><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;">They are compatible. Happy to open the probe separately if there is appetite.</p><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;"><strong>On the<span class="Apple-converted-space"> </span><code>isValid</code><span class="Apple-converted-space"> </span>reorder</strong>, tested as its own variant: the page stops responding<br>to any script about 9 s in and never recovers — worse than the original deadlock,<br>which at least leaves the shell responsive. Requests freeze rather than climb, so<br>it isn't a request storm, and controls through the same harness stay clean, so it<br>isn't the harness. The accompanying<span class="Apple-converted-space"> </span><code>shouldDispatchRequest</code><span class="Apple-converted-space"> </span>change also looks<br>redundant: with<span class="Apple-converted-space"> </span><code>isValid</code><span class="Apple-converted-space"> </span>corrected, the object-cache fallback calls<br><code>hasByUUID(..., true)</code><span class="Apple-converted-space"> </span>and runs the same corrected check; the two were identical in<br>our runs.</p><p style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;"><strong>Caveat.</strong><span class="Apple-converted-space"> </span>Our reproduction amplifies the production race with added REST latency,<br>so we have shown these mechanisms are<span class="Apple-converted-space"> </span><em>sufficient</em><span class="Apple-converted-space"> </span>to produce the hang, not that<br>they are the only cause.</p>

tinsch:
@naomigassler thanks for taking a look into it again. However I don't feel this conversation leads to a real fix, and it _feels_ like a conversation with an LLM rather than a conversation between humans. I am sorry if I am wrong here. Could you please disclose it here and on your other PR? (https://github.com/DSpace/dspace-angular/pull/6111). If the comments above were not made by an LLM, please ignore the text below, and let me know I was wrong.

If the comments and the code are entirely submitted by an LLM, I unfortunately have to step back from reviewing it further. 

The reasons are:
- The fix provided here is not a real fix for the issue. With the PR applied, our application does not indicate anymore that it is in a broken state if the backend is down on a lot of routes and pages. The other PR fixes that only for one particular page. This is not acceptable, and I think that implementing a liveness probe through this LLM conversation could lead to further bugs.
- I also feel that I am wasting my time, arguing with an LLM when I did not explicitly choose to do so.
- I feel disrespected when a comment is not written by a person, but every one of my replies is. It just feels disrespectful that the other person did not take the time to think about it and write a short paragraph of text to acknowledge my thoughts.

jlipka:
Hey,
I apologize for the confusion here. We've been hitting a few of these small bugs lately, and in my eagerness to find a quick fix, I might have overlooked some edge cases. I’ll open a separate bug ticket for it when I get a chance. Thanks, @tinsch, for catching this!

---

Regarding the actual issue and @tinsch’s valid comments:
The question remains for me whether a check to see if the backend is still accessible should be implemented in this way. Perhaps a simple “health” request running in the background at regular intervals could just as effectively highlight the problem of an unreachable backend.

It seems a bit odd to use the user’s navigation/click behavior as the trigger for determining the status of the backend server. After all, in the best-case scenario, the user wouldn’t even be able to perform UI interactions if the server is actually unreachable—in the sense of a proactive check rather than a reactive response to the interaction itself. But that’s just a quick thought.

tinsch:
@jlipka you are right, there might be a better way to to the liveness/health check than on every user navigation. It is just the way how it is implemented now. We could open a separate ticket for this.

naomigassler:
Hello @tinsch,

i apologise for not having disclosed that my comments were written by an LLM. I am not a java developer but I have been working with programmers for more than 20 years on customising our dspace installation. Now i can do this myself with an LLM. Having spent many days on finding the solution for this bug, I was delighted to give something in return to the dspace project. but without prior experience in filing PRs in GitHub I relied on an LLM for help. I understand if that disqualifies this patch. 

I agree with @jlipka that we should have a different way to check if the backend is still alive. But I don't think I can contribute to this issue, I would be a bit out of my depth.

tinsch:
@naomigassler thanks for disclosing it now, and thanks for taking the time to finding a fix with the help of an LLM. Using an LLM does not disqualify the patch, the problem is that the patch introduces other issues. I understand that you were happy to contribute something, and I myself also would have loved to find a fix for this annoying endless spinner issue! So thanks for bringing this up and attempting a fix.

Do you want to have this PR and the other one (https://github.com/DSpace/dspace-angular/pull/6111) stay open, or can we close them and continue the work on a fix somewhere else?

naomigassler:
@tinsch I think this PR is the correct fix, we just need to find another way to check that the backend is still responding. #6111 fixes a real bug which is not apparent now but might be in the future. the fix is tiny and obvious so no harm to keep it. But at the moment it doesn't do anything so you might want to just close it.

tinsch:
@naomigassler but how do you think we could proceed here? Since the fix needs adjustments, but you cannot provide them?

naomigassler:
This comment is a bit long, but it wraps up the entire story. So please bear with me.

We have a reliable way to reproduce the hang on our system. So I looked precisely again (with the help of a coding agent) what is causing the hang. (I should have done this earlier but got too excited with finding a fix for a bug that has plagued my team for so long.)

The timeline below involves two root request cache entries (…b577eb5662f8 and …259dce4bdb8f), reproduced with 2,500 ms injected REST latency and two navigations 800 ms apart.

+340 ms
nav 1: **invalidateRootCache** marks cached root …b577eb5662f8 stale

+361 ms
guard's **findByHref: hasByHref** false → object-cache fallback → true → dispatches …259dce4bdb8f. Working correctly

+1147 ms
nav 2: **invalidateRootCache** marks …259dce4bdb8f, still in flight, stale → ResponsePendingStale

+1209 ms
guard again: **hasByHref** returns true for that entry → early return, nothing dispatched

+1210 / +2918 ms
**skipWhile(rd => rd.isStale)** skips **ResponsePendingStale,** then skips the **SuccessStale** completion

then nothing, for ~27 s

So there are two defects playing together:

(a) **isValid()** calls a stale-but-loading entry valid — it tests **isLoading** before **isStale**, and **isLoading** includes **ResponsePendingStale**, so the stale check is never reached. This is precisely the predicate jlipka's reorder fixes.

(b) **skipWhile** sits in front of the re-request. **findByHref** places **skipWhile(rd => rd.isStale)** immediately before **reRequestStaleRemoteData** — the operator whose entire job is "if this is stale, re-request it". In 3/3 runs its tap saw only RequestPending and ResponsePending, willReRequest: false; it never saw the stale completion. **The culprit is this skipWhile, not anything in HALEndpointService.** (The original description of this PR points at HALEndpointService as the culprit. This is now determined to be wrong.)

**#5936 removes one trigger of these 2 defects playing out, but it doesn't remove the defects.**

**@jlipka was closer to the root cause than we were.** Their first idea — moving `skipWhile` below `reRequestStaleRemoteData` — is defect (b). Their second — the `isValid` reorder — is defect (a). (But it stops the UI responding entirely on our system in 3/3 runs, independently of whether their first idea was applied or not.)

I tried moving `skipWhile` below `reRequestStaleRemoteData` on a clean system based on `main` @ `8a63671` with our system backend. This removes the hang without fail (0 hangs in 24 rounds across 3 runs, against 3/3 hangs on the unmodified clone). **So all credit to @jlipka for the swap they proposed in their 13 August comment.**

There is some extra cost: with the move, root requests per navigation go from 9–10 to 26–27 . That is the re-request actually happening — requests that should have been sent and were not — but it is roughly a threefold increase, and it may be worth looking at whether the re-request should be de-duplicated.

For the path forward: this PR should now be closed, up to @jlipka to put in a PR with their reordering idea.

Note: It is a question whether invalidating the root cache on every navigation is the best way to test whether the backend is responsive. I have written a probe for our system which proactively tests the backend at a given interval and gives the user a message when the backend becomes unresponsive even if the user is not navigating the UI. It is 283 lines across 13 files. Removing the invalidation of the root cache then becomes an efficiency gain. @tinsch you said earlier that you worry, understandably, about new bugs being introduced with new code. But, if there is an interest to try out this probe I'll be happy to put it into a PR.

Below is a spec to insert into `src/app/core/data/base/base-data.service.spec.ts` to test the fix. 

```ts
// Regression: the operator that re-requests stale data sits behind a skip that
// removes stale data. These two specs drive the real pipeline — note that they
// deliberately do NOT stub reRequestStaleRemoteData, unlike every other spec in
// this file, because the interaction between it and the skipWhile immediately in
// front of it is exactly what is broken.
describe(`findByHref with an entry that goes stale while in flight`, () => {
  beforeEach(() => {
    spyOn(service, 'buildHrefFromFindOptions').and.returnValue(selfLink);
  });

  it(`should re-request the stale RemoteData`, () => {
    testScheduler.run(({ cold }) => {
      // What a caller subscribing after the invalidation sees: the request it
      // is waiting on was marked stale while still in flight, and then
      // completed stale. There is no non-stale value in the stream, so the
      // skipWhile never stops skipping.
      spyOn(rdbService, 'buildSingle').and.returnValue(cold('a-b', {
        a: remoteDataMocks.ResponsePendingStale,
        b: remoteDataMocks.SuccessStale,
      }));

      service.findByHref(selfLink, true, true, ...linksToFollow).subscribe();
    });

    // createAndSendGetRequest is called once by the original findByHref, and
    // once more by each re-request. A count of 1 therefore means nothing
    // re-requested the stale data, and the caller waits for ever.
    expect((service as any).createAndSendGetRequest.calls.count()).toBeGreaterThan(1);
  });

  it(`should not emit the stale RemoteData to the caller`, () => {
    testScheduler.run(({ cold, expectObservable }) => {
      spyOn(rdbService, 'buildSingle').and.returnValue(cold('a-b', {
        a: remoteDataMocks.ResponsePendingStale,
        b: remoteDataMocks.SuccessStale,
      }));

      // The skip exists to stop stale cached data reaching consumers, and it
      // must keep doing that: re-requesting is not a licence to leak.
      expectObservable(service.findByHref(selfLink, true, true, ...linksToFollow)).toBe('---');
    });
  });
});
```

A few notes from my coding agent about this:

- On unmodified `main` @ `8a63671` the first spec fails with `Expected 1 to be greater than 1`; with the `skipWhile` moved it passes, and the other 5,926 specs are unaffected. It needs no backend and no configuration — it is a marble test over a stubbed request service so anyone can check it with a checkout and `npm test`.
- **Placement.** It sits inside the existing `describe(findByHref, …)` block, immediately after the `when useCachedVersionIfAvailable is false` block. That's not cosmetic — it inherits that block's `spyOn(service as any, 'createAndSendGetRequest')`, which is what the first assertion counts. My first attempt put it one level out, and the spec then failed with a `TypeError` instead of the assertion.
- **No stub on `reRequestStaleRemoteData`.** Every other spec in the file does `spyOn(service as any, 'reRequestStaleRemoteData').and.callFake(() => (source) => source)`, which neutralises exactly the operator that's broken. Dropping that stub is the whole point.
- **The marble `'a-b'`** with two stale values and no non-stale prefix models the second navigation's subscriber from the browser trace: `skipWhile` is a prefix operator, so once it passes a non-stale value it never skips again. The frozen consumer is the one whose stream starts stale and stays stale.
- **`toBeGreaterThan(1)` rather than `toBe(2)`.** With two stale emissions the fixed code re-requests once per emission, so the exact count is 3, not 2. `toBeGreaterThan(1)` states the property that matters — "something re-requested it" — without pinning a number that a future dedup change would break. It still fails cleanly on unmodified main: `Expected 1 to be greater than 1`.
- **The second `it` passes with and without the fix, deliberately.** It documents that the skip still does its legitimate job, so the change can't be accused of leaking stale data to consumers. It's a non-regression assertion, not a discriminating one.

PR #5984, opened 2026-07-22
W2P-141449 Add heartbeat support to process overview and detail UI

### LINKED ISSUE #12967: Heartbeat support for scripts & processes
## Is your feature request related to a problem? Please describe.
When scripts are started through the UI there is currently no way to indicate that the script is running & actually making progress.

## Describe the solution you'd like
Extend the "Processes Overview" page with a heartbeat column that shows when the script last reported some activity.

ana-postolache-atmire:
## Description
Fixes https://github.com/DSpace/DSpace/issues/12967
Requires backend PR: https://github.com/DSpace/DSpace/pull/12840

Displays the process heartbeat timestamp in the UI: as a column in the running processes table on /processes, and as a field on the process detail page. Shows "N/A" when no heartbeat has been registered yet.

## Instructions for Reviewers
* Added a `heartbeat` field to the `Process` model
* Process detail page (`/processes/{id}`) shows the heartbeat below Start time (N/A if absent)
* Process overview table shows a heartbeat column, only for processes with status RUNNING
* Requires the corresponding backend PR exposing `heartbeat` via `ProcessRest`

To test: start a long-running process (e.g. reindex with enough items) and observe the heartbeat updating in the running processes table and in the process detail page.

### DIFF
--- a/src/app/core/processes/process.model.ts
+++ b/src/app/core/processes/process.model.ts
@@ -71,6 +71,11 @@ export class Process implements CacheableObject {
   endTime: string;
 
   /**
+   * The heartbeat for this process
+   */
+  @autoserialize
+  heartbeat: string;
+  /**
    * The name of the script run by this process
    */
   @autoserialize
--- a/src/app/process-page/detail/process-detail.component.html
+++ b/src/app/process-page/detail/process-detail.component.html
@@ -44,6 +44,10 @@
         <div>{{ process.startTime | date:dateFormat:'UTC' }}</div>
       </ds-process-detail-field>
     }
+      <ds-process-detail-field id="process-heartbeat"
+                              [title]="'process.detail.heartbeat' | translate">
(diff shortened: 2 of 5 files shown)


KevinVdV:
Adding a screenshot of what it looks like in the UI for a visual reference of what information it provides
Scherm­afbeelding 2026-08-14 om 15 32 38

tdonohue:
@KevinVdV and @ana-postolache-atmire :  I haven't reviewed/tested this, but I wanted to add a quick note. 

I'm not sure if the term "heartbeat" would have a shared meaning to all users of DSpace.  While I'm not against using that term, I think we need to consider providing some sort of hint/tip/popup in the UI which allows users to understand that the "heartbeat" time represents the date/time that the process last checked in...and maybe information about how frequently a process should be updating the heartbeat (so that you can more easily tell if the process is stalled).

This hint/tip/popup could be done via our usual inline help options or some other manner (it doesn't matter to me).  I just feel it needs to exist for people to understand this new column.

PR #6057, opened 2026-08-03
Show the uploaded file name in submission upload notifications

milanmajchrak:
## Description
Drop three files into a submission and you get three identical "Upload successful" toasts. If one of them fails, there's no way to tell which. This makes each notification name its file. When we don't have a name to show, you get the old generic message, same as before.

## Instructions for Reviewers

What changed:

* `ds-uploader` gets a second output, `onCompleteItemWithFile`, carrying `{ response, fileName? }`. The old `onCompleteItem` still fires first and still emits the bare response, so the three other places using the uploader don't notice anything.
* `onUploadError` is typed as `UploaderError` now instead of `any`, and the upload notification keys all moved into one small `getNotificationContent()` helper.
* Two new keys in `en.json5`. Locales that don't have them yet fall back to the generic message instead of printing a raw `submission.sections.upload.*` key — that's what the `default` interpolate param is doing.

To try it: drop two or three files into the Upload section and check each toast names the right one. For the size-limit path you need `dspace.upload.max` configured, otherwise the feature is simply off and the test tells you nothing.

Before, both files give you the same message:

Before

After:

After

## Checklist
- [x] Created against `main` — no, this one is a backport onto `dtq-dev`.
- [x] Small — 51 lines of actual code across 4 files, the rest is specs and i18n.
- [x] Follows our coding conventions.
- [x] Passes ESLint — green in CI.
- [x] No circular dependencies — green in CI.
- [x] TypeDoc on everything new.
- [x] Specs — 24 new ones. CI went from 5467 to 5491 passing, nothing else moved.
- [x] Accessibility — the notification component isn't touched, so it's the same `role="alert"` region and the same escaped rendering path, only the text differs. I haven't put a screen reader on it.
- [x] i18n keys, no hardcoded English.
- [x] How to test it — above.
- [x] No new dependencies.
- [x] Nothing new to document.

<sub>Written with some help from [Claude Code](https://claude.com/claude-code).</sub>

### DIFF
--- a/src/app/shared/upload/uploader/uploader-complete-event.model.ts
+++ b/src/app/shared/upload/uploader/uploader-complete-event.model.ts
@@ -1,0 +1,17 @@
+/**
+ * An interface that represents a completed single-file upload, carrying both the
+ * parsed response body and the client-side file name of the file that completed.
+ */
+export interface UploaderCompleteEvent {
+  /**
+   * The parsed response body (e.g. the submission object returned by REST)
+   */
+  response: any;
+
+  /**
+   * The client-side name of the file that completed uploading. Present only when a
+   * non-empty file name is known — an empty file name is never emitted, so the presence
+   * of this key means a usable name is available. Whitespace-only names are not trimmed.
+   */
+  fileName?: string;
+}
--- a/src/app/shared/upload/uploader/uploader.component.spec.ts
+++ b/src/app/shared/upload/uploader/uploader.component.spec.ts
@@ -29,6 +29,21 @@ describe('UploaderComponent', () => {
   let testFixture: ComponentFixture<TestComponent>;
   let html;
 
+  /**
+   * Bring an injected UploaderComponent instance to the state in which its ng2-file-upload
+   * callbacks are installed, so that `app.uploader.onCompleteItem(...)` runs the component's code.
+   */
+  const driveUploader = (app: UploaderComponent): void => {
+    app.uploadFilesOptions = Object.assign(new UploaderOptions(), {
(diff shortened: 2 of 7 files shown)


lgeggleston:
@milanmajchrak thanks for submitting this improvement to usability!

MMilosz (commented):
I ran the changes locally without diving into the code. Overall, the functionality works very well. What I verified:

- Properly uses filename received from backend: when uploading `file.txt` and re-uploading it, backend names the duplicate `file (1).txt` and the exact same name is shown in notification
- Batch uploads work as expected
- Error handling works properly
- Non-Latin alphabet filenames display correctly
- Changing to other language (without the new i18n key) shows generic notification
- Poisonous filenames (XSS attempts, unusual encodings, control characters) don't corrupt the UI or cause unexpected behavior
- Missing filenames are handled gracefully

One concern from me: each file upload now triggers **4 additional API requests** (see screenshot), which can impact performance especially for batch uploads in peak hours. 

### Before
Upload completes with a single 201 response.

Image

### This PR
After a 201 response, upload fires another 4 requests (I see that the first 201 response already returns filename, in this case it's `core-js-banners`, without extension).

Image

@milanmajchrak is it possible there's some component re-rendering happening? Or is it intentional?

#### Other screenshots

All looks good, I'm sharing some screenshots of the different cases tested

Image

Image

Image

Image

Image

Image

nathanmlf (approved):
Hi @milanmajchrak! I've tested it on my end and everything seems to be working as expected. Nice job with the new implementation!

Image

PR #6075, opened 2026-08-06
Add detailed warnings to SSR-related configs about impact on SEO.

tdonohue:
## Description
This PR is coming out of a discussion I had with the Google Scholar team today.  Google Scholar is still noticing a large number of DSpace sites that are either disabling SSR entirely, or accidentally disabling it on pages that require indexing.  

This PR therefore adds inline comments next to the key SSR-related configurations which warn sites about the risk of SEO impacts if they modify the current settings in Production scenarios.  This is just an attempt to ensure these warnings are documented at the place(s) where sites might be modifying their configurations...to ensure they don't overlook them.

## Instructions for Reviewers
* PR only adds inline comments.  No actual code added.
* Therefore, just review the comments and determine whether they are clearly written or need any updates.

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -8,8 +8,10 @@ ui:
   ssl: false
   host: localhost
   port: 4000
-  # Specify the public URL that this user interface responds to. This corresponds to the "dspace.ui.url" property in your backend's local.cfg.
-  # SSR is only enabled when the client's "Host" HTTP header matches this baseUrl. The baseUrl is also used for redirects and SEO links (in robots.txt).
+  # Specify the public baseURL that this user interface responds to. This corresponds to the "dspace.ui.url" property in your backend's local.cfg.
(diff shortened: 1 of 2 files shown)


tdonohue:
Closing & reopening as tests never ran for this PR.

config/config.example.yml
EikLoe: In the last time, I saw several repositories upgrading to 9.3 or 10.0 and forgetting to set the baseURL despite the fact, it's mentioned as a breaking change in the release notes. They didn't noticed the problem until someone tried accessing the page with javascript disabled. Perhaps, we might want to add more explicit warning as well as a mention of the [CVE-2026-27739](https://github.com/advisories/GHSA-x288-3778-4hhx).

tdonohue: @EikLoe : Good point!  I've just updated this PR to also add a warning above the "baseUrl" configuration.

EikLoe (approved):
This looks good to me! Thanks @tdonohue

tdonohue:
Merging with one approval as this is just code comments.  We can always improve further as necessary.

PR #6095, opened 2026-08-13
[DURACOM-320] fix Drag-and-Drop Malfunction in submission form

### LINKED ISSUE #3624: Drag-and-Drop Malfunction in submission form
## Describe the bug
When starting a new submission of an Item that has a field with the chips buttons (e.g 'dc.subject' for Publication) if the metadata has multiple values distributed into rows, the drag&drop functionality will not work between rows.

[Screencast from 08-11-24 17:25:32.webm](https://github.com/user-attachments/assets/7ca632a2-bb90-484f-8739-1d61141e7006)

In DSpace, when using Angular Material’s drag-and-drop (cdkDropList) with cdkDrag in a multi-row layout, drag-and-drop functionality fails if items span more than one row. Specifically, when attempting to drag an item between rows, the intended swap does not occur, and items in other rows may shift unexpectedly. This issue limits the functionality to single-row layouts.

This issue arises because the current version (17.3.10) of Angular Material’s cdkDropList does not natively support multi-row. Support of multiple row has been added to Angular 18.

So for DSpace 9 the fix can be done when upgrading to Angular 18. While in DSpace 7 and DSpace 8, which use earlier Angular versions, we need a custom fix.

## To Reproduce
1. Go to [sandbox.dspace.org](https://sandbox.dspace.org/home) and start a new Item submission in the "Articles" collection.
2. In the Describe section, under "Subject Keyword," add enough items to create multiple rows.
5. Attempt to drag an item from one row to another, and observe the bug as items shift incorrectly.

## Expected behavior
Drag-and-drop functionality should work seamlessly across multiple rows, allowing users to swap items without causing other rows to shift.

FrankGameDev:
## References
* fixes #3624 , as requested here https://github.com/DSpace/dspace-angular/issues/3624#issuecomment-3164913496
* https://4science.atlassian.net/browse/DURACOM-320

## Description
Updated the CDK drag-and-drop behavior from *horizontal* to *mixed*to enable dropping chips on wrapped rows

## Instructions for Reviewers

List of changes in this PR:
- Changed the **cdkDropListOrientation** from **horizontal** to **mixed**

### How to test the fix
Open a submission form, try adding some authors to let the chips wrap (or resize the browser/use DevTools to achieve it), and then try dragging and dropping chips onto different rows

### DIFF
--- a/src/app/shared/form/chips/chips.component.html
+++ b/src/app/shared/form/chips/chips.component.html
@@ -8,7 +8,7 @@
       }
     </ng-template>
     @for (c of chips.getChips(); track c; let i = $index) {
-      <div role="listitem" class="nav-item me-2 mb-1 d-flex flew-row"
+      <div role="listitem" class="nav-item me-2 mb-1 d-flex flex-row"
         #t="ngbTooltip"
         triggers="manual"
         [ngbTooltip]="tipContent"
--- a/src/app/shared/form/chips/chips.component.html
+++ b/src/app/shared/form/chips/chips.component.html
@@ -1,5 +1,5 @@
 <div [className]="'float-start w-100 ' + wrapperClass">
-  <div role="list" class="nav nav-pills d-flex flex-column flex-sm-row" cdkDropList cdkDropListOrientation="horizontal" (cdkDropListDropped)="onDrop($event)">
+  <div role="list" class="nav nav-pills d-flex flex-column flex-sm-row" cdkDropList cdkDropListOrientation="mixed" (cdkDropListDropped)="onDrop($event)">
     <ng-template #tipContent>
       @for (tip of tipText; track tip) {
         <p class="text-start p-0 m-0">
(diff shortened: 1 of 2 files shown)


lgeggleston:
Thanks for the fix @FrankGameDev! Since this was fixed on 8.1, we can try to port this back to 9.x and 10.x as well.

FrankGameDev:
> Thanks for the fix @FrankGameDev! Since this was fixed on 8.1, we can try to port this back to 9.x and 10.x as well.

Here they are: #6096 #6097!

FrankGameDev:
I'm closing this PR, as i created the branch with the wrong name. Here it is the new one https://github.com/DSpace/dspace-angular/pull/6098

lgeggleston:
Hi! Thanks @FrankGameDev, sounds good.  
Also just a note for future PRs, usually it's not necessary to create backports beforehand - I apologize for the confusing wording of this [comment](https://github.com/DSpace/dspace-angular/pull/6095#issuecomment-5283820302)!  The "port to" labels added here mean that a Github action to auto-port will be attempted when a PR is merged, which is usually the easiest method to try before manually backporting (and will save you from having to transfer any review changes to the other PRs).

PR #6111, opened 2026-08-21
Fix <ds-search> rendering an empty page when the search configuration request fails

naomigassler:
This PR was developed with the help of an LLM. I also relied on an LLM to write the description below.

---------------

## References

_No issue ticket — this was found while investigating #5936 and is filed on its own merits. 

* Related to #5936 (closed) — the investigation that surfaced this, but this PR does not depend on it and reproduces on unmodified `main`.

---

## Description

`<ds-search>` renders nothing at all — no search form, no results, no empty state and no error — when its **search-configuration** request fails. The page looks like an empty repository. This PR lets the component initialise so its existing error handling can run.

---

## Instructions for Reviewers

`getConfigurationSearchConfig()` pipes through `getAllSucceededRemoteDataPayload()`, so a failed configuration request never emits. `combineLatest` therefore never fires, `initialized$` stays `false`, and the whole `<ds-search>` template is skipped — including the `ds-error` block that `search-results.component` already has. The error handling is present; it is simply never constructed.

**List of changes in this PR:**
* `search.component.ts` — initialise the component when the search-configuration   request fails, falling back to empty sort options, so the existing error and   empty states can render.
* `search.component.spec.ts` — two specs covering error-versus-empty.

### How to test

**1. Unit specs — no backend, no configuration required.**

```
npm test
```

The two added specs in `search.component.spec.ts` cover the case directly.

**2. In a browser.** This reproduces on unmodified `main`; the patch is not needed to see the bug.

1. Log in to any DSpace instance.
2. Go to **`/mydspace`**.
   ⚠ **Not `/search`** — `/search` does not issue this request at all, because the    homepage has already cached the `default` configuration. `/mydspace` uses    `workspace`, which has not been cached. `/access-control/bulk-access` also    reproduces it.
3. DevTools → Network → tick **Enable request blocking**, add the pattern:
   `*discover/search?configuration=*`
   ⚠ Use exactly that pattern. `*/discover/search*` also blocks the HAL
   endpoint-discovery request, which produces a different failure that
   demonstrates nothing.
4. Reload.

**Before this PR:** the content area is empty — no search form, no results, no
empty state, no error. Only the submission drag-and-drop zone remains, because that sits outside `<ds-search>`. The Network panel shows **no
`discover/search/objects` request at all**: the component never gets far enough to ask for results.

**After this PR:** the page renders and is usable — though with an empty sort
dropdown, and results in the fallback `dc.title,ASC` order rather than the
configured `lastModified,DESC`, because only the configuration request was blocked and the results request still succeeds.

**To see the error state**, add a second blocking pattern
`*discover/search/objects*`. You then get **"Error fetching search results"**
instead of a blank page.

### Known limitation, and a question for reviewers

Some installations report a `400` on this endpoint when the search fires before
authentication is attached (*"An anonymous user cannot perform a workspace or workflow search"*). Upstream already treats that as "no results" via `showError()`'s `statusCode !== 400` exemption, and this PR leaves that exemption intact. Our stack does not reproduce that 400, so we have shown that this change does not *introduce* a spurious error state — not that it handles that case correctly.

Separately: when only the configuration request fails, this PR substitutes defaults **silently** — the sort order changes and the sort dropdown is empty, with nothing saying so. It may be better to render the error *alongside* the results rather than
degrading quietly. I do not have a settled view and would welcome yours.

---

## Checklist

- [x] My PR is **small in size** — 44 insertions, 3 deletions, across 2 files.
- [ ] My PR **follows all coding best practices** based on the Code Conventions Guide. 
- [x] My PR **passes ESLint** validation using `npm run lint` — full-repo clean.
- [x] My PR **doesn't introduce circular dependencies** — verified via `npm run check-circ-deps`.
- [ ] My PR **includes TypeDoc comments** for all new or modified public methods and classes. _
- [x] My PR **passes all specs/tests and includes new/updated specs** — 40/40 plus 2 new specs.
- [ ] My PR **aligns with Accessibility guidelines**. _
- [x] My PR **uses i18n keys** instead of hardcoded English text — no new key; reuses the existing `ds-error` component and message.
- [x] My PR **includes details on how to test it** — see above.
- [x] My PR does not include **new libraries/dependencies**.
- [x] My PR does not include **new features or configurations**.
- [ ] My PR **links an issue ticket** — none exists; see References.

### DIFF
--- a/src/app/shared/search/search.component.spec.ts
+++ b/src/app/shared/search/search.component.spec.ts
@@ -41,6 +41,7 @@ import {
 } from '@dspace/core/shared/search/search-filters/search-config.model';
 import { SidebarServiceStub } from '@dspace/core/testing/sidebar-service.stub';
 import {
+  createFailedRemoteDataObject$,
   createSuccessfulRemoteDataObject,
   createSuccessfulRemoteDataObject$,
 } from '@dspace/core/utilities/remote-data.utils';
@@ -200,6 +201,7 @@ export function configureSearchComponentTestingModule(compType, additionalDeclar
     getConfigurationSortOptions: sortOptionsList,
     getConfig: filtersConfigRD$,
     getConfigurationSearchConfig: of(searchConfig),
+    getSearchConfigurationFor: createSuccessfulRemoteDataObject$(searchConfig),
     getCurrentConfiguration: of('default'),
     getCurrentScope: of('test-id'),
     getCurrentSort: of(sortOptionsList[0]),
@@ -450,6 +452,30 @@ describe('SearchComponent', () => {
         //Check that the last method from which the search depend upon is being called
         expect(searchManagerStub.search).toHaveBeenCalled();
       }));
+
+      describe('when the search configuration request fails', () => {
+        beforeEach(() => {
+          searchConfigurationServiceStub.getSearchConfigurationFor.and.returnValue(
+            createFailedRemoteDataObject$('Error fetching search configuration', 500),
+          );
+        });
+
+        it('should still initialize, so the search results error can be rendered', fakeAsync(() => {
+          comp.ngOnInit();
+          tick(100);
+
+          let initialized: boolean;
+          comp.initialized$.subscribe((res) => initialized = res);
+          expect(initialized).toBeTrue();
+        }));
+
+        it('should still request results, whose failure is what the user is shown', fakeAsync(() => {
+          comp.ngOnInit();
+          tick(100);
+
+          expect(searchManagerStub.search).toHaveBeenCalled();
+        }));
+      });
     });
   });
 });
--- a/src/app/shared/search/search.component.ts
+++ b/src/app/shared/search/search.component.ts
@@ -38,7 +38,10 @@ import { DSpaceObject } from '@dspace/core/shared/dspace-object.model';
 import { followLink } from '@dspace/core/shared/follow-link-config.model';
 import { Item } from '@dspace/core/shared/item.model';
 import { ListableObject } from '@dspace/core/shared/object-collection/listable-object.model';
-import { getFirstCompletedRemoteData } from '@dspace/core/shared/operators';
+import {
+  getAllCompletedRemoteData,
+  getFirstCompletedRemoteData,
+} from '@dspace/core/shared/operators';
 import { PaginatedSearchOptions } from '@dspace/core/shared/search/models/paginated-search-options.model';
 import { SearchFilterConfig } from '@dspace/core/shared/search/models/search-filter-config.model';
 import { SearchObjects } from '@dspace/core/shared/search/models/search-objects.model';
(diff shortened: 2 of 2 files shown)


tinsch:
@naomigassler could you please provide a self written short text that outlines
- what this PR is attempting to fix
- how we can reproduce and test it

lgeggleston:
Hi @naomigassler, since this PR seems to have spun off from https://github.com/DSpace/dspace-angular/pull/5936 which was closed and based on [your comment there](https://github.com/DSpace/dspace-angular/pull/5936#issuecomment-5394858333), I lean towards closing this as well.  

However, if it stands separately from 5936 and you feel strongly about moving it forward, it would need a clear PR description with testing steps as @tinsch describes, and should also use the [PR template and checklist](https://github.com/DSpace/dspace-angular/blob/main/.github/pull_request_template.md#checklist) before this could be reviewed.

naomigassler:
Hi @lgeggleston and @tinsch,

After the resolution of #5936 I have looked into this patch more deeply. Here is what this patch is trying to accomplish. If for some reason (network glitch, etc) the **search configuration** request initiated by the user fails to reach the backend, there is currently no notification for the user. The search simply shows no results. This is particularly problematic when looking at /mydspace. The user might be led to believe that there are no items at all in the queue, and nothing is telling them that reloading the search would reveal what is there.

This addresses specifically the configuration request. If the *results* request fails, existing code already renders "Error fetching search results".

This situation can easily be reproduced as follows (it doesn't need #6111 to be implemented):

1. Log in to any DSpace instance.
2. Go to /mydspace. (NOT /search — /search does not issue this request at all, because the homepage has already cached the `default` configuration. /mydspace uses `workspace`, which has not been cached.)
3. Open DevTools → Network → tick "Enable request blocking", and add the pattern: `*discover/search?configuration=*`
4. Reload the page.

Before the fix: the content area is empty — no search form, no results, no empty state, no error message. Only the submission drag-and-drop zone remains, because that sits outside `<ds-search>`.

After the fix: the page renders and is usable, but with an empty sort dropdown, and results come back in the fallback `dc.title,ASC` order rather than the configured default — because only the configuration request was blocked and the results request still succeeds. (The patch substitutes the result of the missing configuration with default values.)

To see the error state instead, add a second pattern `*discover/search/objects*` — you then get "Error fetching search results" rather than a blank page.

The PR also carries two unit specs, so this can be checked with a checkout and `npm test`.

One thing to decide on and possible change: when the configuration request fails, the patch substitutes defaults silently. The sort order changes and the sort dropdown is empty, but nothing says so. It may be better to render the error *alongside* the results rather than quietly degrading the /mydspace page.

I updated the PR description with the testing steps using the PR template and checklist.

lgeggleston:
@naomigassler thank you for adding the testing steps and comment.  For additional context, did you run into a search configuration request failure while testing, or is this a hypothetical for now?

Adding this to a project board for reviewers to take a look, but note that this will need to find a developer to thoroughly review the code in order to move forward.  

Finally just to let you know, we are working on a policy for DSpace regarding AI-assisted contributions.  It's not finalized yet, but if you want to get a sense of those discussions you can take a look at recent [developer meeting notes](https://wiki.lyrasis.org/spaces/DSPACE/pages/453738793/2026-08-27+DSpace+Developers+Meeting).

naomigassler:
@lgeggleston Thanks for adding this to the project board. Your answer your question: no, we have not seen this while using our installation.  We came at it from the other end while working on resolving a system hang: while testing a solution for that we noted that  /mydspace could render nothing at all — and then reproduced it deterministically with request blocking.

What makes this disconcerting is how little it takes to trigger the fault: just a single failed request to discover/search?configuration (which is not that unlikely on our installations running behind an auth layer in front of the API). And, that the consequence is out of proportion: the user sees what looks like an empty repository without any indicator that a reload would make their items reappear. 

Thanks for looking into how to incorporate AI into this project. Just to say again: i deeply value this project, have been using it for decades so I would be happy to contribute even though I'm not a trained java developer.

PR #6186, opened 2026-09-04
Explicitly add `chokidar`, `readdirp` and `yaml` as dev dependencies to solve `dependabot` issues

tdonohue:
## Description
Recently, in every PR it creates, `dependabot` will attempt to remove the `chokidar`, `readdirp` and `yaml` dependencies from our `package-lock.json` file because they are _not explicitly used_ in dspace-angular (as they are transitive dependencies).  

However, removing these dependencies causes immediate errors in `npm install` because all three dependences are **required** by Angular (and a few other direct dependencies).

Examples of `dependabot` PRs which attempt to remove these dependencies include the following:
* #6185
* #6150 
* #6152 
(This currently occurs on almost EVERY `dependabot` PR against the `main`, `dspace-10_x` and `dspace-9_x` branches)

This PR attempts to resolve the issue by adding all three as *development* dependencies, which should make it clear to `dependabot` they are needed.  

```
# The versions here correspond to the versions that are currently listed in our `package-lock.json` file.
npm install --save-dev chokidar@^4.0.3
npm install --save-dev readdirp@^4.1.2
npm install --save-dev yaml@^2.9.0
```

**NOTE:** Because this impacts several branches, this PR will need to be ported to `dspace-10_x` and `dspace-9_x`.  It's likely it will need to be recreated on each branch _using the above commands_ as these branches do not have identical dependencies.  It does NOT need to be ported to 8.x because that uses `yarn`, and this `dependabot` issue seems to be specific to `npm`.

**NOTE 2:** After this PR is merged, every `dependabot` PR against `main` will need to be recreated via `@dependabot recreate`.  That should tell dependabot to use the updated packaging settings & hopefully resolve the issues.

## Instructions for Reviewers
* Verify no change in behavior to the `npm install` process as well as the build (e.g. `npm run build:prod`) or running of the UI (e.g. `npm run serve:ssr`).
    * I've tested all these commands locally and found no differences in behavior.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -114,6 +114,7 @@
         "@typescript-eslint/rule-tester": "^8.62.1",
         "@typescript-eslint/utils": "^8.47.0",
         "axe-core": "^4.12.1",
+        "chokidar": "^4.0.3",
         "commander": "^14.0.3",
         "compression-webpack-plugin": "^9.2.0",
         "copy-webpack-plugin": "^14.0.0",
@@ -148,6 +149,7 @@
         "postcss-import": "^14.0.0",
         "postcss-loader": "^4.0.3",
         "postcss-preset-env": "^7.4.2",
+        "readdirp": "^4.1.2",
         "rimraf": "^6.1.3",
         "sass": "^1.101.0",
         "sass-loader": "^12.6.0",
@@ -156,7 +158,8 @@
         "tsconfig-paths": "^4.2.0",
         "typescript": "~5.9.3",
         "webpack": "^5.107.2",
-        "webpack-cli": "^6.0.1"
+        "webpack-cli": "^6.0.1",
+        "yaml": "^2.9.0"
       },
       "engines": {
         "node": ">=20.0.0 <25.0.0"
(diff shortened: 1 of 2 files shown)


tdonohue:
@alanorth : Pinging you on this because I *think* this should finally fix the `dependabot` issues we've been seeing.  This PR is not ideal because it adds these transitive dependencies as dev dependencies, but the output of the PR is very promising.  If you look closely it makes similar changes to `package-lock.json` which `dependabot` keeps trying to do...but then it also moves these dependencies to explicit dev dependencies.

Thoughts welcome.  As I noted above, if we find this works for `main`, I think we'd need to build similar PRs for 10.x and 9.x. I'm not confident that this PR would port cleanly in an automated fashion...we may have to run the commands (see description) on each branch individually and create a new PR for each.

kshepherd (approved):
+1 approved by inspection, looks good. thanks @tdonohue

alanorth:
Thanks @tdonohue. I think this is the best we can do for now, but it's an unfortunate solution. For reference, I saw an upstream dependabot issue that is similar: https://github.com/dependabot/dependabot-core/issues/14110.

tdonohue:
@alanorth : Thanks for the reference. It looks like others have recently reported this issue again with `dependabot` in https://github.com/dependabot/dependabot-core/issues/15039  In any case, once this is fixed at the dependabot-level, I think it would be good to try undoing this PR.

But, for the time being, this PR has fixed the issues with all our `dependabot` PRs. They are all working again (once we ask dependabot to recreate it).

PR #6208, opened 2026-09-08
fix: avoid undefined community breadcrumb lookup

### LINKED ISSUE #6168: `/communities/create` sends `undefined` request to the API
## Describe the bug
When creating a new top level Community, devtools show that the app sends the following `undefined` request to the server:

[server URL]`/api/core/communities/undefined?embed=logo&embed=subcommunities&embed=collections&embed=parentCommunity`

Issue discovered on both sandbox.dspace.org and demo.dspace.org

## To Reproduce
1. Log in as admin
2. Open devtools, go to the network tab
3. From the sidebar, select New > Community (or go directly to the `/communities/create`). Press the 'Create a new top-level community'
4. Observe the `undefined` request sent to the backend

## Expected behavior
No undefined request being send

## Related work
Nothing found

dvd233:
## References

Fixes #6168

## Description

When opening the top-level community creation route, the community breadcrumb resolver is configured with `breadcrumbQueryParam: 'parent'`, but no `parent` query parameter is present. The resolver currently falls back to the route `id`, which is also absent on `/communities/create`, and requests the REST endpoint with `undefined`.

Return an empty breadcrumb result when the optional parent parameter is absent. Existing community routes and sub-community creation routes continue to resolve their breadcrumbs normally.

## Instructions for Reviewers

List of changes in this PR:
* Skip the community data lookup for top-level community creation.
* Add resolver coverage for normal routes, parent-community creation, and top-level creation.

To verify:
* Open **New → Community** as an administrator.
* Confirm that creating a top-level community does not issue a request to `/api/core/communities/undefined`.
* Confirm that creating a sub-community still loads the parent breadcrumb.

## Testing

* TypeScript semantic check for the resolver and regression spec.
* Resolver branch harness: baseline fails on the top-level route; the fix passes all three route cases.
* Prettier check for the changed files.
* `git diff --check`.

The full repository npm test suite was not run locally because the dependency tree is not installed in this environment; CI should run the repository's standard checks.

### DIFF
--- a/src/app/core/breadcrumbs/community-breadcrumb.resolver.spec.ts
+++ b/src/app/core/breadcrumbs/community-breadcrumb.resolver.spec.ts
@@ -1,0 +1,91 @@
+import { Community } from '../shared/community.model';
+import { createSuccessfulRemoteDataObject$ } from '../utilities/remote-data.utils';
+import { communityBreadcrumbResolver } from './community-breadcrumb.resolver';
+
+describe('communityBreadcrumbResolver', () => {
+  let breadcrumbService: any;
+  let communityDataService: any;
+  let community: Community;
+  let parentCommunity: Community;
+  let state: any;
+
+  beforeEach(() => {
+    community = Object.assign(new Community(), {
+      uuid: '1234-65487-12354-1235',
+      type: 'community',
+    });
+    parentCommunity = Object.assign(new Community(), {
+      uuid: '4321-76548-45321-5321',
+      type: 'community',
+    });
+    breadcrumbService = {};
+    state = {};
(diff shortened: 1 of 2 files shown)


lgeggleston:
Hi @dvd233, thanks for your first contribution to DSpace.  Since the issue was submitted recently, @MMilosz would you be willing to take a quick look at this fix maybe?

One note on the DSpace GitHub process for future PRs: we encourage you to comment on the issue (in this case, https://github.com/DSpace/dspace-angular/issues/6168) when you start working on it, ideally before submitting a PR. This allows us to assign the issue to you, check if there are any blockers, and avoid multiple people accidentally working on the same issue. Thanks!

PR #6233, opened 2026-09-10
Update DOMPurify by overriding version used by Mirador

tdonohue:
## Description
Update the version of [`dompurify`](https://www.npmjs.com/package/dompurify) used by Mirador by overriding the value (in the `overrides` section of `package.json`). Bumps  `dompurify` from v2.5.8 to v3.4.15  

This major upgrade to v3 moves us to a version of `dompurify` that has much fewer vulnerability warnings (as v2 has a larger number of potential vulnerabilities).

However, from my testing, this major upgrade to `dompurify` is still compatible with Mirador v3 because `dompurify` is only used in a few places to sanitize the HTML in Mirador.

## Instructions for Reviewers
* Verify tests pass
* Optionally, verify no changes in behavior to Mirador viewer for [IIIF Integration ](https://wiki.lyrasis.org/spaces/DSDOC10x/pages/408944973/IIIF+Configuration)
    * NOTE: I've verified locally that the Mirador viewer still works the same as before. I've not found any differences in behavior.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9891,6 +9891,13 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/@types/trusted-types": {
+      "version": "2.0.7",
+      "resolved": "https://registry.npmjs.org/@types/trusted-types/-/trusted-types-2.0.7.tgz",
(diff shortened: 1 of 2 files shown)


kshepherd (approved):
+1 thanks @tdonohue

Issue #6256, opened 2026-09-20
Skip authenticated-only authorization checks for anonymous visitors


bram-atmire:
## Summary

An anonymous request for an item page performs ~15 separate `authz/authorizations/search/object` feature checks, one HTTP request per feature. Roughly ten of them ask whether the visitor can perform an action that only an authenticated user could ever perform (edit, administer, create a version, manage groups, submit, ...). For a visitor who is not logged in, the answer to those is always "no," yet each still costs a round-trip and makes the backend re-resolve the visitor's groups and scan resource policies.

Because crawler and browser-impersonating scraper traffic is unauthenticated and dominates access load, this waste falls on exactly the requests that matter most for scaling.

This complements #3161 (obtain permissions in fewer REST requests). That issue proposes *batching* the checks; this one proposes *not making* the ones that can never succeed for an anonymous visitor. The two compose: fewer checks asked, and the remainder batched.

## Evidence (measured)

Measured on a local DSpace 11 stack (Angular SSR + REST backend) with OpenTelemetry tracing, one anonymous item render:

| Metric | Before | After skip |
|---|--:|--:|
| Backend REST calls (Contexts) | 42 | 31 |
| Authorization feature checks | 15 | 5 |
| Backend DB queries | ~300 | substantially fewer |
| Single-core SSR render throughput | ~3.3 renders/s | ~4.8 renders/s |
| Warm SSR render latency | ~0.30s | ~0.21s |

Notes:
- Only ~5 of the 42 calls serve the item's actual content; authorization is the single largest share of the backend's database load for the page.
- The fan-out is the item page's data pattern, not a rendering-mode quirk: with SSR the Node server makes the calls; without SSR (or after hydration, navigating item to item) the browser makes the same calls. Both hit the backend.
- The ~15 checks fire against the item and the Site object.

## Proposed change (Angular, existing REST contract)

In `AuthorizationDataService.isAuthorized()`, when checking the **current** visitor for a feature that only makes sense once authenticated, answer `false` locally without a REST call whenever no one is logged in. Features an anonymous visitor can legitimately be granted are always checked normally.

Kept (still checked for anonymous): `canDownload`, `canRequestACopy`, `canSendFeedback`, `epersonRegistration`, `epersonForgotPassword`, `canViewUsageStatistics`.

[...]

PR #6257, opened 2026-09-20
Skip authenticated-only authorization checks for anonymous visitors

bram-atmire:
## References
Implements the anonymous case proposed in #6256. Complements #3161 (obtain permissions in fewer REST requests): #3161 batches the checks, this skips the ones that can never succeed for an anonymous visitor. The two approaches compose.

## Description
An anonymous item-page request performs ~15 `authz/authorizations/search/object` feature checks, one HTTP request per feature. About ten ask whether the visitor can edit, administer, create a version, manage groups, submit, and so on. For a visitor who is not logged in the answer is always "no," yet each still costs a round-trip and makes the backend re-resolve the visitor's groups and scan resource policies. Crawler and browser-impersonating traffic is unauthenticated and dominates access load, so this waste falls on the requests that matter most.

This changes `AuthorizationDataService.isAuthorized()` so that, when checking the **current** visitor for a feature that only makes sense once authenticated, it answers `false` locally without a REST call whenever nobody is logged in. Features an anonymous visitor can legitimately be granted (`canDownload`, `canRequestACopy`, `canSendFeedback`, `epersonRegistration`, `epersonForgotPassword`, `canViewUsageStatistics`) are always checked normally.

Correctness:
- The frontend is not the security boundary — skipping a check only avoids showing a control; the backend still enforces on any real action. The skipped features map to WRITE/ADMIN/login-only capabilities never granted to the ANONYMOUS group, so the skip returns the same "no" the server would.
- The check waits for `isAuthenticationLoaded()` before reading auth state, so a logged-in admin is never transiently treated as anonymous while authentication resolves.

## Measured impact
Local DSpace 11 stack (Angular SSR + REST) with OpenTelemetry tracing, one anonymous item render:

| | Before | After |
|---|--:|--:|
| Backend REST calls | 42 | 31 |
| Authorization checks | 15 | 5 |
| SSR render throughput (1 core) | ~3.3/s | ~4.8/s |
| Warm render latency | ~0.30s | ~0.21s |

The fan-out is the item page's data pattern, so it applies to both SSR and client-side rendering.

## For reviewers — open design question
The authenticated-only feature set is a static list in `AuthorizationDataService`. That holds for every deployment I know of, but the edit/write features check a policy the backend genuinely evaluates, so a static list steps slightly outside the policy-driven model. An alternative is to have each `AuthorizationFeature` declare whether it requires authentication (a property the UI reads) — more principled, but a larger change. Happy to go either way.

## Instructions for reviewers
1. As an anonymous visitor, open an item page and watch the requests to `authz/authorizations/search/object`: the admin/edit checks no longer fire.
2. Log in as an admin and confirm all edit/admin controls still appear on the item page (they wait for authentication to load before being decided).

## Checklist
- [x] Uses only the existing REST contract (no backend change).
- [x] Includes unit tests (skip returns false with no request for anonymous; the check still runs when authenticated; kept features are never skipped; nothing is decided before auth loads).
- [x] Targets `main`.
- [ ] CI green (pending).

---
*Investigated and developed with Claude Fable 5.1 (Anthropic) via Claude Code. All figures measured on a local DSpace 11 stack.*


lgeggleston:
Hi @bram-atmire, thanks for the PR! Would you be able to add the complete PR checklist to the description? Also you may have already seen, but this is currently running into a few errors in the CI (looks like just linting).

